PR #24, opened 2018-01-22
Added instructions about specifying memory allocation in .sbtopts

adarshp:


bsharpataz:
good change! thanks!
if it matters -- we can specify a diff port fairly easily (and one case in which it could matter is that the influence search also uses 9000, and that is coded in pretty deep, so when I need to run both simultaneously I run something like `sbt "webapp/run 9091"` and then go to localhost:9091

PR #25, opened 2018-01-22
Reorganize rules

bsharpataz:
made some changes including:

- moved reverse-causal rules to a separate file, now loaded by master.yml
- removed some no-longer-relevant rules
- made due-to rule more robust (and added a 2nd)
- made "with" rule more robust
- added some examples to rules and some comments above some rule about perhaps removing, etc

### DIFF
--- a/src/main/resources/org/clulab/wm/grammars/causal.yml
+++ b/src/main/resources/org/clulab/wm/grammars/causal.yml
@@ -12,19 +12,6 @@ vars:
 
 rules:
 
-# ------------------- Verbal PROMOTION/INHIBITION Rules --------------------------
-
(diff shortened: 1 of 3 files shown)


MihaiSurdeanu:
Why did you remove the rules in causal.yml?

bsharpataz:
I didn't really remove any

   - The promotion_trigger one  I removed bc I am now passing those
   triggers with causal triggers
   - the reverse_nou_trigger one got put in its own file
   - the with one wasnt removed, it was revamped
   - the others were commented out in master anyway bc they were old rules
   from wmseed that were largely made obsolete by the REACH causal rules

MihaiSurdeanu:
Got it.

PR #30, opened 2018-01-24
added POS trimming to local entity finder, with a test

bsharpataz:


bsharpataz:
OK -- found the failing test - it was bc the test in master was written to account for *lack* of trimming.  I fixed the test and pushed, so tests should pass and then I'll merge

MihaiSurdeanu:
Ok!

PR #37, opened 2018-01-28
Affect rules

bsharpataz:
- moved generic rules to linkersTemplate.yml
- left explicitly causal rules only in causal.yml
- removed "CauseEffect" from rule names
- now passing addLabel into templates
- added Affect to taxonomy
- added Affect triggers
- using linkersTemplate to extract Affect mentions
- added a couple comments for organization in master.yml


kwalcock:
It looks reasonable to me, but that's not saying much.

MihaiSurdeanu:
I'd like a review of cause and affect triggers by a linguist (Becky?)...

PR #38, opened 2018-01-28
Fix polarity of dueToSyntax2, fixes p4s3

bgyori:
This rule seemed to be working but assigned the cause/effect arguments incorretly, as far as I can tell. Did I get this right?


MihaiSurdeanu:
I think the original rule was (theoretically) correct. Note that "<" means the traversal goes *against* the flow of the dependency.

bsharpataz:
Yep, I found that last night indirectly, went to fix it, but then realized
it had been assigned to you, thanks for fixing it!

kwalcock:
It sounds like this it good to go then?  Right now I don't see much other than tests passing or failing.

PR #41, opened 2018-01-29
Making the webapp show increased NPs as green, decreased as red, and otherwise quantified as blue...

bsharpataz:
To be honest -- I'm not sure this is the best way to accomplish this, so @kwalcock if you have another, cleaner suggestion or if @MihaiSurdeanu  you think this is all dumb, we can close the PR w/o merge

These changes are done for the benefit of the webapp display.  They include:
- adding NounPhrase-Inc, -Dec, and -Quant to the taxonomy under NounPhrase
- adding a helper method in the keepMostComplete action that collapses all of these to "NounPhrase" so that only one (most complete) nounPhrase is kept for each span
- adding extra entity types in the brat display stuff (main.js)
- and the whole reason for doing this -- I updated the colors of the display items

### DIFF
--- a/src/main/resources/org/clulab/wm/grammars/taxonomy.yml
+++ b/src/main/resources/org/clulab/wm/grammars/taxonomy.yml
@@ -2,9 +2,6 @@
   - Param
   - Agrovoc
   - NounPhrase
-    - NounPhrase-Inc
-    - NounPhrase-Dec
-    - NounPhrase-Quant
 - Transparent
 - Quantifier
 - Event:
--- a/src/main/resources/org/clulab/wm/grammars/taxonomy.yml
+++ b/src/main/resources/org/clulab/wm/grammars/taxonomy.yml
@@ -2,6 +2,9 @@
   - Param
   - Agrovoc
   - NounPhrase
+    - NounPhrase-Inc
+    - NounPhrase-Dec
+    - NounPhrase-Quant
 - Transparent
 - Quantifier
 - Event:
--- a/src/main/scala/org/clulab/wm/AgroActions.scala
+++ b/src/main/scala/org/clulab/wm/AgroActions.scala
@@ -41,7 +41,7 @@ class AgroActions extends Actions with LazyLogging {
     // remove incomplete entities (i.e. under specified when more fully specified exists)
 
     val tbMentionGroupings =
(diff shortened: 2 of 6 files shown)


kwalcock:
Well, at some point I'd like to find an alternative for these string comparisons, but now is probably not the time.  It might be good to keep repetitions of values like "#2471A3" to a minimum.  However, if it works, great!  I'll merge (or watch you do it) after you've had some time to read this.

bsharpataz:
OK -- thanks! good tips!  I'd prefer to get @Mihai's OK before it's merged
bc  it drastically changes the look of the visualizer. I can change the
string html color refs to a variable in a subsequent PR?

MihaiSurdeanu:
I would prefer if this wasn't merged in master. In addition of Keith's comments, I would prefer if the taxonomy were agnostic to the state.

kwalcock:
Is it easy to include a screen shot?  I admit to not pulling the branch, rebuilding, etc. to see how it looks.  Maybe it's not necessary now that Mihai has commented.

bsharpataz:
per @MihaiSurdeanu I will refactor this PR and re-submit -- @kwalcock do you prefer for me to close and open a new one? or refactor and push again?

kwalcock:
Whatever is easiest.  I'll find out what "taxonomy were agnostic to the state" means.

bsharpataz:
oh, sorry -- I didn't understand that that was what  you wanted the
screenshot of.  He's referring to my including the -Inc, -Dec, and -Quant
sub-types under NounPhrase.  That is, rather than have just NounPhrase in
the taxonomy all possible states of NP were listed (which is messy and hard
to maintain).  What I will do instead (I think), and your input on this is
welcome, is keep only NP and later when sending the mentions to the
brat-visualizing webapp code I can edit the label of the NP.  I *think*
this will work.... but admittedly am not 100% sure.

kwalcock:
I just meant a picture of how the webapp looks in the browser with the improvements.

bsharpataz:
I made the changes: 
- removed states from taxonomy (etc, e.g. Actions)
- added a "post-processing" step in the webapp to recover the states for brat-json
- moved colors to variables at the top of main.js
- renamed RAPShell to better title

here's a picture, btw:
screen shot 2018-01-29 at 5 38 14 pm

PR #42, opened 2018-01-30
Handle "poor" to fix p6s2/8

bgyori:
In an attempt to fix the test for p6s2/8, I added "poor" to the decrease triggers and get
```
List(Causal, EntityLinker, Event) => displacement causing poor access to services
    Rule => ported_syntax_1_verb-Causal
    Type => EventMention
    trigger => causing
    effect (NounPhrase, Entity) => access to services
     * Attachments: Quantification(poor), Decrease(poor,None)
    cause (NounPhrase, Entity) => displacement
```
with the expected Decrease on the "access to services" entity, which I thought would result in the test passing, but it is still failing; could someone help me figure out why?

### DIFF
--- a/src/main/resources/org/clulab/wm/grammars/avoidLocal.yml
+++ b/src/main/resources/org/clulab/wm/grammars/avoidLocal.yml
@@ -49,7 +49,7 @@ rules:
       #removed: support, produc
       attenu|abolish|abrog|antagon|arrest|attenu|block|deactiv|decreas|degrad|deplet|deregul|diminish|disrupt|
       down-reg|downreg|dysregul|elimin|impair|imped|inactiv|inhibit|knockdown|limit|loss|lower|negat|nullifi|
-      perturb|prevent|reduc|reliev|repress|resist|restrict|revers|sequester|shutdown|slow|starv|suppress|supress|worse/] # neg reg guys
+      perturb|poor|prevent|reduc|reliev|repress|resist|restrict|revers|sequester|shutdown|slow|starv|suppress|supress|worse/] # neg reg guys
 
 
   - name: "gradable"
--- a/src/main/resources/org/clulab/wm/grammars/master.yml
+++ b/src/main/resources/org/clulab/wm/grammars/master.yml
@@ -3,7 +3,7 @@ taxonomy: org/clulab/wm/grammars/taxonomy.yml
 vars:
   increase_triggers: "acceler|aid|augment|better|boost|elev|enhanc|greater|improv|increas|prolong|promot|rais|reactivat|restor|retent|stimul|support|synerg|up-regul|upregul"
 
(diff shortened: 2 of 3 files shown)


bgyori:
Shouldn't everything listed in Quantifiers.tsv be ignored as a whole? I added a specific one, poor, to the ignore list but most others in the list are not ignored (e.g. large) and thus still show up as part of the extracted entity.

bsharpataz:
yes, @bgyori  -- they should be.  it's a matter of I/someone needs to either figure out how to import the contents of that tsv into the odin grammar yml OR we can generate the avoidLocal on loading...? (or something else obvious and simple that I'm not thinking of...)

bsharpataz:
and @bgyori  re:the failing test... it may be bc of the "extra" Quantification attachment (@kwalcock am I right?)..... I feel like rather than removing the Quantification in the extraction, we should add it to the test? thoughts @MihaiSurdeanu ?

kwalcock:
I'll check...

kwalcock:
If access is changed to the following, then the test passes:
val access = newNodeSpec("access to services", newDecrease("poor"), newQuantification("poor"))
So yes, the extra quantification is a problem.  The complete set of attachments is required to match.  Should the test be changed?  Should it be testing for a subset instead?

bsharpataz:
no -- def not a subset (IMO).  But I think we should consider changing the test -- the attachment is ambiguous... and this will happen a lot I imagine.  _poor_ is the trigger on the Decrease, it's not quantifying a decrease (i.e. it's not a _poor decrease_, which would be something diff).  It's quantifying the noun, so I think it should also have the Quantification(poor).  OK, I convinced myself at least, @kwalcock , wanna add the Quantification to the test and merge (assuming passes other tests)? (unless vetoed by @MihaiSurdeanu )

MihaiSurdeanu:
I agree. 

But this leads to the issue that some quant triggers overlap with INC and DEC triggers. How do we formally handle this situations?

bsharpataz:
we allow them in both places... the quant triggers and inc/dec triggers
don't get consumed with the finding of Quant/Inc/Dec events, so they are
"available" for later rules... is this what you are asking?

MihaiSurdeanu:
Yes. Good.

bsharpataz:
great -- ok, so @kwalcock should update the test and merge...?

MihaiSurdeanu:
Yep

kwalcock:
So, I guess the test/picture will change to reflect

val access = newNodeSpec("access to services", newDecrease("poor"), newQuantification("poor"))

bsharpataz:
yes, exactly

kwalcock:
My sentence,

"In Eastern Equatoria - one of the drought-hit areas - the latest food security and nutrition assessments in November 2016 revealed that households experiencing poor food consumption nearly doubled from 19 to 37 percent when compared with the same time in 2015.",

has a case

foodConsumption = newNodeSpec("food consumption", newDecrease("poor")

which would now pass if it was similarly changed to

foodConsumption = newNodeSpec("food consumption", newDecrease("poor"), newQuantification("poor"))

Is this valid?  By ambiguous above, do you mean (poor(food consumption)) vs. ((poor food) consumption)?

bsharpataz:
yes IMO!  it's the same as was found by Ben -- I think we should use
foodConsumption
= newNodeSpec("food consumption", newDecrease("poor"),
newQuantification("poor"))
@mihai do you agree?
and no, I don't think it's particularly ambiguous about what poor attaches
to in this case, I may have mis-spoke earlier?

MihaiSurdeanu:
I agree.
Mihai

PR #46, opened 2018-02-01
Fixed p1s4 unit test

adarshp:


MihaiSurdeanu:
@bsharpataz: we really need to find a solution to avoid duplicating triggers in avoidLocal...

bsharpataz:
Agreed, I’ll try to work with Marco tomorrow

PR #49, opened 2018-02-01
Egoitz test1

EgoLaparra:
- Changed keepMostCompleteEvents to remove incomplete events only when they overlap.
- Added dueToSyntax3 in causal.yml to capture entities with clausal modifiers.
- Added ported_syntax_3_noun in modifiersTemplate.yml to link quantifiers to entities with clausal modifiers.
- Added newQuantification("low") and newQuantification("poor") for the p1s1 tests in TestCagP1.scala.
- Changed "coping capacities" to "capacities" for the p1s1 tests in TestCagP1.scala.
- Changed one passingTest of p3s2 in TestCagP3.scala to failingTest. After adding collapse as a decrease trigger the test fails due to a parse fail.


bsharpataz:
@EgoLaparra thanks for fixing issue in keepMostComplete!
@kwalcock good to me... (famous last words, right?) :)

kwalcock:
Checking... Not that I'm qualified.

PR #54, opened 2018-02-02
Zheng test

ZhengTang1120:


MihaiSurdeanu:
Very clean. Good job!

@bsharpataz: to discuss Tue: let's use word embeddings to semi-automatically expand the triggers.

PR #58, opened 2018-02-02
Get console input to work in IDEs

kwalcock:
This annoying change, or something similar, seems to be necessary in order to allow for debugging within the IDEs on Windows.  The jline line reader is not quite right and is replaced with the standard Java Scanner when arranged.


kwalcock:
Just making sure you're on your toes toes.  This is part of getting the client/server version to work.

PR #59, opened 2018-02-02
Add reload functionality to shell

kwalcock:
This is supposed to support a quick edit/debug cycle.  Rules can be changed and reloaded while the shell and its processor continue to run.  Because files are loaded from resources, it may not work when run from a jar file (i.e., it might have to run in an IDE).  It seems to work with eclipse and supports debugging.

### DIFF
--- a/src/main/scala/agro/demo/RAPShell.scala
+++ b/src/main/scala/agro/demo/RAPShell.scala
@@ -77,14 +77,14 @@ object RAPShell extends App {
 
   val commands = ListMap(
     ":help" -> "show commands",
-    // ":reload" -> "reload grammar",
+    ":reload" -> "reload grammar",
     ":exit" -> "exit system"
   )
 
   // creates an extractor engine using the rules and the default actions
   val ieSystem = new AgroSystem()
 
-  var proc = ieSystem.proc
+  //var proc = ieSystem.proc // Not used
 
   println("\nWelcome to the RAPShell!")
   printCommands()
@@ -98,8 +98,7 @@ object RAPShell extends App {
         printCommands()
 
       case ":reload" =>
-        println("Not supported yet.")
-        // TODO
+        ieSystem.reload()
 
       case ":exit" | null =>
         running = false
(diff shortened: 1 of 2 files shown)


kwalcock:
Becky, it's probably good if there is someone to keep me honest.  Also this might be your code that I'm messing up and you should be in the loop.

kwalcock:
Noted.  I suspect that var is inherent in reload, but maybe not.  load => val, reload => var.

MihaiSurdeanu:
Ok to merge for now. But:
- All those vars should Option[...], so they are None by default and Some() when initialized.
- Please change all Agro* names to something more generic.

kwalcock:
Becky, had we (more likely, you) come up with a better name?  I should go ahead and finally use it.

Mihai, indeed I didn't do it the Scala way.  Do you know of a way not to force clients to deal with Option and at the same time not duplicate the initialization code?  My first attempt was to delay init() to the very next LOC.  There is a different, not pretty way using tuples (below).  Do you happen to know of a lighweight alternative?  Thanks!

  var (a, b, c, d) = init()
  
  def init() = (4, "hello", null, 4.5) // Something more complicated than this
  
  def reload() = init() match {
    case (a, b, c, d) => this.a = a; this.b = b; this.c = c; this.d = d;
  }

MihaiSurdeanu:
Tuples are easy to abuse in Scala. I think a tuple with more than 2 or 3 arguments is not easily readable.

I recommend:

- Defining a class to package what needs reloading, and implement an apply() method. Something along these lines:

class LoadableAttributes(val a:Int, val b:String)

object LoadableAttributes {
     |   def apply(): LoadableAttributes = {
     |     new LoadableAttributes(1, "one") // load stuff here
     |   }
     | }

Then, the code to create the object becomes simply:

lazy var loadableAttributes = LoadableAttributes()

Reload becomes:

loadableAttributes = LoadableAttributes()

- I recommend you use lazy for this. And for the processor as well.

kwalcock:
Thanks!  That works nicely.  There's not a null and hardly a var in sight.  It will be submitted soon on a different branch.

Is InfoExtractor OK for AgroSystem?  Also AgroEntityFinder => IEEntityFinder, AgroActions => IEActions, etc.?

PR #60, opened 2018-02-04
Rewrite reload to avoid nulls and vars

kwalcock:


kwalcock:
This code may be changed soon as processor access is rerouted to the server.  The compiler says that lazy vars are not an option.  Only vals can be lazy.  Agro is soon to be changed.

MihaiSurdeanu:
Nice.

PR #64, opened 2018-02-06
modified entity constraints to allow for JJ entities

bsharpataz:
also: 
- removed last old test :/
- added a trigger @ajaynagesh  will need (as long as I was at it...)


bsharpataz:
ps -- this will affect @mithunpaul08 's test(s) too

kwalcock:
I just wonder about the ignored tests.  Should the reason they are being ignored be documented or should they now be deleted?

bsharpataz:
They are on the old framework... I don't know the answer.  @MihaiSurdeanu  do you care?

bsharpataz:
I can also check to see if they're easily fixed...?

kwalcock:
Could there just be a comment added that says
// This test is being ignored because... it is designed for an old framework, namely...
Ignore is a red flag and I think a short comment will save someone in the future a lot of questions.

bsharpataz:
sure! I'll add!

PR #65, opened 2018-02-06
fixed P3S5

mithunpaul08:
-added "mean" (stem/lemma for meant, meaning, means) as a causal trigger
- also have added JJ into VALID_FINAL_TAG in EntityConstraints.scala. This  might be already done in becky's pull request merge before this.


kwalcock:
Was Becky removed because she saw the change in person?

mithunpaul08:
@kwalcock I had accidentally assigned this to becky. Then removed her when I noticed that.
@MihaiSurdeanu I had asked @bsharpataz  about this. So she had said add mean as a stem for meant/meaning etc. The trigger word per se in this sentence was meant though

MihaiSurdeanu:
@mithunpaul08: Can you please resolve this conflict first?
@bsharpataz: can you please take a look at my questions above? I still think "mean" should not be a causal trigger. At least not on its own. Maybe part of a longer trigger phrase?

bsharpataz:
@MihaiSurdeanu -- the JJ is fine (I added it actually, on a PR that was
merged, it allows for us to have 'economic' as an entity with a
Decrease('decline') from the text: 'economic decline'), the 'mean' as a
trigger is probably over matching-y... we added it bc of his sentence which
is something like: "The increased A has meant B decreased" (I mean, it's
not so simple, but that's the idea).  The hit to our precision (at the sake
of recall) may be more than we want to take, if so IMO some of his tests
would prob need to be removed.  Also, the precision-hit would likely be
worse with certain types of documents, in the UN reports, 'A means B' may
well be more likely causal than not...?

bsharpataz:
@MihaiSurdeanu -- what would the longer trigger phrase be?  As I mentioned,
in something like A has meant B, the word 'meant' is the whole trigger
phrase, it's just whether or not we think that 'mean/meant' is usually
causal or another relation and how much we want to favor precision vs
recall.  We can also add a score here -- an extraction with this trigger is
less certain than one with the trigger "cause"
I guess we can also lengthen it to "meant" and "means" to avoid matching
with the noun which will not have the same causal meaning...
We can also bump it to Affect_triggers (as the true causality is less
certain?)

mithunpaul08:
resolved conflicts. Mostly picked stuff from master branch over my version.  have changed "mean" to "meant" for the time being, until we decide another way out?

bsharpataz:
I think the newest decision on 'meant' is to remove the trigger and
**change** the sentence (and necessarily the tests) to have another, more
explicit trigger -- something like "driven"

mithunpaul08:
replaced meant in test sentence with caused. removed meant from triggers

mithunpaul08:
Got the latest from master+added in my changes (replaced meant in the sentence with caused). Please merge this branch with master.

PR #67, opened 2018-02-06
Fixing p2s2 

ajaynagesh:
- Added adverbs to Quantification attachments
- Changed keepMostComplete: to filter out subsumed Mentions. (currently using a hard coded sieve)
- Update the Testing framework
- Had to recreate some of the changes in https://github.com/clulab/eidos/pull/64 to make these pass

- also pushed into the same branch tests fixing p2s3 and p3s1 (I pushed the changes after creating the PR for p2s2)
- added rule for entityQuantification 
- copula rule
- new triggers and quantifiers

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ scalaVersion := "2.12.4"
 //EclipseKeys.withSource := true
 
 libraryDependencies ++= {
-  val procVer = "7.0.0"
+  val procVer = "7.0.1"
 
   Seq(
     "org.clulab" %% "processors-main" % procVer,
--- a/src/main/resources/org/clulab/wm/grammars/avoidLocal.yml
+++ b/src/main/resources/org/clulab/wm/grammars/avoidLocal.yml
@@ -1,8 +1,4 @@
-taxonomy:
-  - Entity:
-    - NounPhrase
-  # references, tables, etc.
-  - Avoid
+taxonomy: org/clulab/wm/grammars/entities/grammar/taxonomy.yml
 
 vars: org/clulab/wm/grammars/triggers.yml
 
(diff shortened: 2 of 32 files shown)


bsharpataz:
ok, so 'spread' is not marked as a causal trigger, it's included as an increase_trigger, which I feel is warranted.  
Re: the large number of changes -- several of them were included in my recent PR which was recently merge I think (including adding JJ in EntityConstraints, etc), so a fresh merge of master onto this branch would (?) make those not seem like changes (??)

bsharpataz:
@kwalcock I fixed the merge conflict, and I worked with @ajaynagesh on this, it's a lot at once but I think they're good changes.  The addition of adverbs to the quantification is important (though we may think about renaming them) to capture things like "unprecedented high levels of X."  Also important is the fix to keepMostComplete....  that said -- i'd like both you and @MihaiSurdeanu  to review again for sure

bsharpataz:
it has changed... where are you thinking about matching it -- the other
tests?
I know I have one test assigned to me that will need to have this notation
-- something about 'unprecedented high'

kwalcock:
It's no problem.  I was just describing a Quantification with jsonld and it looks like it has changed.

bsharpataz:
Ah, gotcha...

ajaynagesh:
Resolved the conflicts..

bsharpataz:
OK -- I fixed the bad merge (I seem to recall @kwalcock saying something about github not always being trustworthy...) :)  Anyway -- the travis build should work now -- where are we with the approving/not approving?  @kwalcock -- I'll do the same fixes with the merge on my other PR that is subsequent to this (https://github.com/clulab/eidos/pull/72)

kwalcock:
I'm never sure about merging when Mihai is a reviewer but hasn't responded.

MihaiSurdeanu:
let’s have this policy: if either one of the 3 of us approves, it’s Ok to merge. But please approve only if you understand the code.

PR #68, opened 2018-02-09
P6s1 last2

bsharpataz:
- added 'shrunk' as decrease trigger
- modified my sentence/tests to allow for not-broken parse


bsharpataz:
good call, done, please check :)

PR #71, opened 2018-02-10
Rename Agro

kwalcock:
This is just a suggested change.  Files and classes are renamed.  Some variables will follow.


kwalcock:
The code will also soon be broken into three parts: main, demo, webapp.

kwalcock:
That's no problem at all :)  Here's a counter-suggestion: how about pushing the names down a level (see below)?  If the project is ever given a different name, these classes do not need to change with it.  On the other hand, it might be difficult to combine two projects, especially since we're not using an eidos namespace anywhere.  Eidos is fine as well.

Actions -> BaseActions
AgroActions -> Actions

EntityFinder -> BaseEntityFinder
AgroEntityFinder -> EntityFinder

AgroSystem -> System (maybe too generic)

bsharpataz:
The Actions that AgroActions currently extends is in Odin, so a rename
there is not going to happen, if that influences the discussion

kwalcock:
That might be good.  org.clulab.wm.Actions will inherit from org.clulab.odin.Actions.

MihaiSurdeanu:
I would prefer to avoid classes with same name. This leads to code readability issues.
How about we just go with the Eidos prefix? It is fairly generic, and it's unlikely we'll change the project name.

kwalcock:
Eidos it is.  There will be an update shortly.

kwalcock:
I'm just going to start over because of the merge conflicts.

PR #72, opened 2018-02-10
P6s2 take2

bsharpataz:
OK -- so wanted to get this going, but....
NOTE: it needed some changes Ajay had submitted, so it is off of his branch.  When PR https://github.com/clulab/eidos/pull/67 is merged eventually, **I'll merge that newer master onto this** to make my actual changes more apparent

major changes:
- changed a couple of sentences (but documented it)
- **added a new type of test** to TestUtils for tests that we *currently* cannot pass, but with planned modifications/extensions of the system we should be able to.  I therefore don't want to delete them and forget the issues, but I also don't want to just leave it as "failingTest"... *thoughts??*

minor changes:
- added some quantifiers
- added another causal due_to rule
- tweaked the HomeController webapp display to fix the super annoying mismatch of the words and arrows!
- removed a println Ajay had added

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ scalaVersion := "2.12.4"
 //EclipseKeys.withSource := true
 
 libraryDependencies ++= {
-  val procVer = "7.0.0"
+  val procVer = "7.0.1"
 
   Seq(
     "org.clulab" %% "processors-main" % procVer,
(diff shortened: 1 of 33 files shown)


MihaiSurdeanu:
I like the idea of postponed tests!

I'll let you merge this into master when you think. It looks Ok to me.

PR #77, opened 2018-02-12
added directed/undirected to taxonomy for querying

bsharpataz:
to make the dumping to JSON-LD easier...


MihaiSurdeanu:
I thought we decided to merge Affect and Causal?
Remember that MITRE said that for them Affect is Causal when they don't know the polarity. Since polarity is captured in the states of entities, for us they would be similar.

bsharpataz:
Did we? We can also keep them separate and merge for use case... kind of
trying to keep things domain independent... or we can call them casual with
just a lower confidence score...

MihaiSurdeanu:
Ok… 

But if we accept MITRE’s definition, which I like, it really doesn’t make sense to separate...

PR #83, opened 2018-02-16
Flip cause/effect argument in ported_token_1_noun

bgyori:
I found that many example sentences were resulting in extractions where the cause/effect arguments were flipped. I traced it back to ported_token_1_noun-Causal. Here are some minimal examples:
```
List(Causal, EntityLinker, Event) => benefit boosts agricultural production
	Rule => ported_token_1_noun-Causal
	Type => EventMention
	trigger => boosts
	effect (NounPhrase, Entity) => benefit
	cause (NounPhrase, Entity) => agricultural production

List(Causal, EntityLinker, Event) => agricultural production drives food security
	Rule => ported_token_1_noun-Causal
	Type => EventMention
	trigger => drives
	effect (NounPhrase, Entity) => agricultural production
	cause (NounPhrase, Entity) => food security
```
I then flipped the cause/effect arguments in the pattern of this rule, and am now getting the right argument assignments:
```
List(Causal, EntityLinker, Event) => benefit boosts agricultural production
	Rule => ported_token_1_noun-Causal
	Type => EventMention
	trigger => boosts
	cause (NounPhrase, Entity) => benefit
	effect (NounPhrase, Entity) => agricultural production

List(Causal, EntityLinker, Event) => agricultural production drives food security
	Rule => ported_token_1_noun-Causal
	Type => EventMention
	trigger => drives
	cause (NounPhrase, Entity) => agricultural production
	effect (NounPhrase, Entity) => food security
```

### DIFF
--- a/src/main/resources/org/clulab/wm/grammars/linkersTemplate.yml
+++ b/src/main/resources/org/clulab/wm/grammars/linkersTemplate.yml
@@ -264,7 +264,7 @@ rules:
     type: token
     label: ${ label }
     pattern: |
-      @effect: Entity (?<trigger> [word=/(?i)^(${ trigger })/ & tag=/^NN/]) @cause: Entity
+      @cause: Entity (?<trigger> [word=/(?i)^(${ trigger })/ & tag=/^NN/]) @effect: Entity
 
 
   # this rule is needed because PP attachment of "by" is often wrong


bsharpataz:
so -- the rule is intended to capture nominal triggers, in the example "benefit boosts agricultural production," `boosts` is incorrectly tagged as NNS.  Let me look at the rule some more -- we may just want to tighten it up or remove it?  It's ported directly from REACH, I can also see if I screwed up the polarity when I ported it... more soon...

bsharpataz:
and when I look at the comment `#Original example: "mTOR activator Rapamycin"` -- I'm not sure that this rule is helpful in our domain?  Is this a common syntactic construction in the docs we're going to read?  I perhaps don't want to fix the rule to give desired results based on a bad parse, but I'm hesitant to remove bc I wasn't involved in REACH and I don't know how frequently a given rule X was used... @MihaiSurdeanu ?

bgyori:
I see, that makes sense in that in the example, Rapamycin is the one that activates mTOR. So this suggests that my "fix" isn't right. Instead, one has to somehow handle the fact that "boosts" and "drives" comes out as NNS.

MihaiSurdeanu:
Becky: We probably need to populate the POS post-processor with a set of
rules that fix POS tags...

bsharpataz:
yeah, probably.  I would also say that with a bit more text it handles boosts correctly, FWIW - like 'the rainfall boosts production", but yeah, we should help the POS tagger to do super weird things like try to make sure sentences have a verb.... wacky I know :)
Thanks so much @bgyori for pointing this out -- your PRs and attention are super helpful and appreciated!

bgyori:
Well, I'm happy to complain about anything that messes up my models :)

MihaiSurdeanu:
Actually, now that we have our own POS tagger and parser, maybe it’s
simpler to just retrain the tools with a few good examples in this domain?

MihaiSurdeanu:
@bsharpataz: can you please fix merge conflicts here?

bsharpataz:
@MihaiSurdeanu it's not a merge conflict that's the hold up here.  Do you want the rule changed or removed?  The only reason it's firing here is the bad parse.  I'd vote for removing instead, as this parse is less likely in the context of an entire sentence, right?

MihaiSurdeanu:
Remove for now.

I’d like to discuss a more rigorous process on Monday.

kwalcock:
Is this one of those Mondays when a more rigorous process could be discussed?  This PR is approaching an expiration date.

bgyori:
I think we concluded that this is not the right approach to fix the issue so feel free to close my PR.

PR #88, opened 2018-02-18
Ease Eclipse integration

kwalcock:
If these cause problems with any other environment, feel free to revert.


MihaiSurdeanu:
This should be fine. But IntelliJ is much better Scala IDE than Eclipse...

PR #89, opened 2018-02-18
rough draft of readme

bsharpataz:
note -- I will submit the PR with the new files that this references now
(PR https://github.com/clulab/eidos/pull/90)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,81 @@
 # eidos
 
-Machine reading system for World Modelers
+Eidos is an open domain machine reading system designed by the Computational Language Understanding (CLU) Lab led by [Mihai Surdeanu](http://surdeanu.info/mihai/) at [University of Arizona](http://www.arizona.edu) for the World Modelers program.  Eidos uses a small set of [Odin](https://github.com/clulab/processors) rules to extract causal events from free text.
 
-Some materials related to this project can be found at https://drive.google.com/open?id=1cHJIfQTr0XE2CEqbo4POSm0-_xzrDP-A. Access is temporarily limited to UA account holders.  Other documents are included in the /doc directory of the repository.
+# Usage
 
+## How to compile the source code
+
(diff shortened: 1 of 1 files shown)


MihaiSurdeanu:
"small set of rules" => "grammar"

kwalcock:
Also, good plan doing this PR separate from code changes!

kwalcock:
And FWIW .\shell also works on certain machines.

kwalcock:
If RAPShell => EidosShell, is there any problem leaving it in agro.demo, or should that be changed as well?

MihaiSurdeanu:
We shouldn’t use the “agro” name anywhere in the software.

bsharpataz:
Yes, I think we should move it, not just rename agro.  The actual directory
separation we have is largely meaningless. Keith, given that you were
thinking some subprojects I was hoping you steer me to repo goodness and
truth... or at least suggest an appropriate directory structure. :)

kwalcock:
In something just submitted, I renamed it to eidos.demo and decided it wasn't too bad to have these Apps packaged in the same jar.  They provide examples of how to use the API.   I'm not sure about having these outside the org.clulab.wm namespace, however. and that goes for Utils.  However, one step at a time.

We might want to merge this one even though not quite complete in order to avoid merge conflicts.

bsharpataz:
@kwalcock re: `And FWIW .\shell also works on certain machines.` -- I am not quite sure what you mean by this and how I should address it

kwalcock:
I was vague.  It is .\shell for windows.  There is a shell.bat to take care of this.  Windows users will figure it out, so it may not be worth mentioning.  Are you ready to merge?  It's more perfect than what we have.

bsharpataz:
I think it's ready to merge -- def better than what we have and @Mihai I
can continue to make any changes you want....?

MihaiSurdeanu:
Yep, please merge.

PR #95, opened 2018-02-19
At least rename Eidos before moving files around

kwalcock:


kwalcock:
I have to do this quickly, otherwise people work on files I'm trying to move.

PR #99, opened 2018-02-20
Make room for testing of evaluation documents

kwalcock:


kwalcock:
Here is the first version with room for testing of evaluation docs.  They have been numbered on Google Drive.

kwalcock:
@bsharpataz, this is what people can use to write their own tests, or at least it's the start of such a framework.  They can always ask me questions about any code they might need to copy and paste.

PR #112, opened 2018-02-23
Restructure

kwalcock:
ExtractFromDirectory to demo, ResourceUtils to utils, new EidosAttachment in attachments and changed packages and imports to match


kwalcock:
This is step 1.  GitHub starts getting confused if it is more complicated.  The next change is demo -> apps.  Anything with a main or derived from App will go here, whether demo or not.  That's a more cosmetic change.  Shell and README.md will change accordingly.  After that most changes should be internal to files.

PR #118, opened 2018-02-25
Consolidate attachment operations in EidosAttachment

kwalcock:


kwalcock:
This is in preparation for adding further attachments and including provenance with them

src/main/scala/org/clulab/wm/eidos/EidosActions.scala
```diff
@@ -46,23 +46,15 @@ class EidosActions(val taxonomy: Taxonomy) extends Actions with LazyLogging {
                     else
                         0
 
-      val attachmentsList = if (mention.isInstanceOf[TextBoundMention])
-                              Some(mention.attachments)
+      val attachmentsSet = if (mention.isInstanceOf[TextBoundMention])
```
MihaiSurdeanu: Why did you remove the Option here?

kwalcock: This is to avoid having to .get it right back out at line 56 in the old code or in the new code to have to distinguish between Some and None.  Instead of Some(Seq) or None, it's Seq or Seq.empty which can be used directly by map().

MihaiSurdeanu: Ok.

PR #122, opened 2018-02-26
Doc2 mihai

MihaiSurdeanu:


src/test/scala/org/clulab/wm/eidos/text/eval6/TestDoc2.scala
```diff
@@ -1,27 +1,190 @@
...
+    ignore should "have correct output #3" taggedAs(Somebody) in {
+      val hunger = NodeSpec("level of hunger", Quant("highest"))
+
+      tester.test(hunger) should be (successful)
+    }
+  }
+
+  { // Paragraph 4
+    val text = """
+                 |“We are very worried to see that food insecurity is spreading beyond conflict areas as rising prices,
```
bsharpataz: also a
`val insecurity = NodeSpec("food insecurity", Inc("spreading"))` ?
and... IDK, we really need to get these predicates, the thing being caused is really "families accessing food"... not just families, we need to talk about these Monday I think bc we need to add that functionality

bsharpataz: and if that -- the insecurity would be correlated with the other events... though we can't handle hyper edges yet

src/test/scala/org/clulab/wm/eidos/text/eval6/TestDoc2.scala
```diff
@@ -1,27 +1,190 @@
...
+    val text = """
+                 |“We are now seeing sharp spikes of need in new areas, such as Eastern Equatoria or Western
+                 |Bahr el-Ghazal, where malnutrition rates in some places are reaching dangerous levels.
+               """
+    val tester = new Tester(text)
+
+    behavior of "TestDoc2 Paragraph 8"
+
+    ignore should "have correct output #8" taggedAs(Somebody) in {
+      val need = NodeSpec("need", Inc("sharp", "spikes"))
```
bsharpataz: I think "sharp" and "spikes" should be switched...?

bsharpataz: and I'd argue there's a Correlation event between them -- btw -- I think I'm going to move Correlation to UndirectedRelations in the taxonomy (FYI)

bsharpataz: and re: malnutrition -- I think we don't know if Inc or Dec (without inference), so I'd argue a Quant or Unmarked (if we do move to those)

src/test/scala/org/clulab/wm/eidos/text/eval6/TestDoc2.scala
```diff
@@ -1,27 +1,190 @@
...
+    val text = """
+                 |The dramatic rise in malnutrition rates, means that in the first four months of the year UNICEF
+                 |has already treated 45 per cent of its planned 2016 caseload of 166,000 children.
+               """
+    val tester = new Tester(text)
+
+    behavior of "TestDoc2 Paragraph 9"
+
+    ignore should "have correct output #9" taggedAs(Somebody) in {
+      val malnutrition = NodeSpec("malnutrition rates", Inc("rise"), Quant("dramatic"))
```
bsharpataz: I know I havent' been consistent and the system's not 100% but *ideally* here we'd have ` NodeSpec("malnutrition rates", Inc("rise", "dramatic"))`... is there a correlation event here too...? the thing being correlated with is lame and hard to get though...

kwalcock:
Thanks for looking over these so carefully!

MihaiSurdeanu:
@bsharpataz, @kwalcock: I cleaned up my tests based on today's meeting.

bsharpataz:
@kwalcock mind if I merge this today/tomorrow?

kwalcock:
Go right ahead.  It was still red until you fixed it.  Thanks!

PR #127, opened 2018-02-27
Start on EidosMention

kwalcock:
This is all I can figure out to do at the moment.  I need to know more about how and when something like this will be used and what specifies the canonical form.  Mentions seem to move in and out of things like Odin and also State and I shouldn't be converting back and forth between an EidosMention and plain ones.


kwalcock:
Even though this is a rough draft, I'm going to merge it so that people can more easily try it out.  I'll put revisions in a new branch.

PR #129, opened 2018-02-27
More RAPS tests

hclent:
Added tests for sentences 2~10 of RAPs texts, and removed a duplicate "small" from the Quantifiers lexicon. 

**Edit** - there is no bug :) 

--- 

I believe that I also found a bug (!?!) in Eidos for attaching Quantifiers to entities like, "small increase in crop diversity" (example from `sent6`). If you put this into the Eidos Shell, you extract:

`crop diversity Attachments: Increase(increase,Some(ArraySeq(small)))` instead of `Increase(increase, Quant(small))`. Thus in the unit testing, this fails:

`val cropDiversity = NodeSpec("crop diversity", Inc("increase"), Quant("small"))`

Note that the Eidos shell will still extract `List(Quantifier) => small` outside of the "crop diversity" entity.

I think that the Attachment to the entity is not updating from a `Some(ArraySeq())` to a `Quant()` when the quantifier is modifying the increase/degrease trigger, rather than the found entity. This is because entities like "low education levels further restrict" (`val education = NodeSpec("education levels", Dec("low"), Quant("low"))`) properly attach the Quantifier. 

The bug I am describing can be observed in `sent6` and `sent7`. Let me know what you think, and I can also try to help fix this. If its not a bug, let me know :)

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/lexicons/Quantifier.tsv
+++ b/src/main/resources/org/clulab/wm/eidos/lexicons/Quantifier.tsv
@@ -2,7 +2,6 @@ significant
 significantly
 seriously
 low
-small
 acceptable
 acceptably
 adequate
--- a/src/test/scala/org/clulab/wm/eidos/system/TestRaps.scala
+++ b/src/test/scala/org/clulab/wm/eidos/system/TestRaps.scala
@@ -1,7 +1,7 @@
 package org.clulab.wm.eidos.system
 
 import org.clulab.wm.eidos.test.TestUtils._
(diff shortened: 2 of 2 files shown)


bsharpataz:
I think that the NodeSpec in the Testing framework here would be:
val cropDiversity = NodeSpec("crop diversity", Inc("increase", "small"))
such that "small" is modifying the increase, rather than the diversity

Thanks for working on these!

hclent:
@bsharpataz You are right! That fixed it. Thanks! I pushed updated tests for sent6 and 7 and now they are passingTests. Let me know if there's anything else you'd like me to add before merging these into master.

PR #136, opened 2018-02-27
Final changes to Doc3 in consultation with Becky.

ajaynagesh:
Also added a few potential rules that might be required to pass the test ..

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/grammars/entityQuantification.yml
+++ b/src/main/resources/org/clulab/wm/eidos/grammars/entityQuantification.yml
@@ -40,3 +40,23 @@ rules:
       trigger = [mention=Quantifier & tag=/^JJ/] #Find the quantifier
       theme: Entity = <xcomp (${agents})
 #      adverbs: Quantifier? = ${quant_modifiers} // todo: we should allow for modification here (e.g. 'more')
+
+  - name: quantification1b
+    priority: ${ rulepriority }
+    example: "moderate to heavy rainfall"
+    label: ${ label }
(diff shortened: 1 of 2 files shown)


bsharpataz:
@MihaiSurdeanu -- when Ajay and I sat, I convinced myself that "rainfall deficits" were a thing, not just "rainfall" + Dec("deficits").... looking back I'm not sure.  Perhaps it's because he had like 45 paragraphs that discussed rainfall deficits at length.  This likely ties into the convo about not splitting up MWEs (like, domain-specific `Proctor & Gamble`...)

MihaiSurdeanu:
Ok, but then how do you handle the fact that you don't include triggers in the entities currently?
Plus, I'm not sure this is true... Can you motivate this more?

bsharpataz:
@MihaiSurdeanu I am not 100% sure I'm convinced either... I guess it would depend on whether the experts think that a `rainfall deficit` is the same as decreased `rainfall`, perhaps a moot question if we're grounding to an ontology. FWIW there is a mechanism for not avoiding certain triggers (e.g., https://github.com/clulab/eidos/blob/master/src/main/resources/org/clulab/wm/eidos/grammars/triggers.yml#L7) in the entityFinder but we may or may not keep it long term...

bsharpataz:
@kwalcock I made some final changes, I think we're ready to merge.  I'll likely do it today/tomorrow if all checks pass

kwalcock:
Go ahead.  Thanks for working it out with Ajay and Mihai and the rule set and and and

PR #137, opened 2018-02-27
Same-As API

ajaynagesh:
- loading gigaword trained vectors (currently hardcoded path)
- loading the vectors only in eidos-shell with an optional parameter (not currently loading them in the webapp)
- made changes to EidosAttachments. Request @kwalcock to review these changes 

Todos:
---
- add a config parameter to load the vectors (once the config framework is ready in Eidos)
- try with other embedding vectors (i. retrofitted vectors ii. Elmo)
- add the vectors to the google-doc


bsharpataz:
hold up on this PR -- I think we're changing SameAs API per convo with Mihai

kwalcock:
I'll work on that configuration file very soon.  I wonder if some of this might move to the EidosMention now.

ajaynagesh:
@kwalcock: Made the required modifications for grounding the ontology. This is ready for your review now. 

3 things that require your attention and help: 

- The embeddings are put in [this location on google drive](https://drive.google.com/open?id=1tffQuLB5XtKcq9wlo0n-tsnPJYQF18oS) (we need to modify the readme to include this link)
- The vectors should not be reloaded when we reload the system (in the UI and shell). I was unsure how to achieve this in the new config setup. 
- Because of the EidosMention.remapOdinArguments, it seemed like we needed to pass those extra arguments throughout. Perhaps you could refactor this to make it cleaner

@MihaiSurdeanu: To ground a particular mention to the ontology (and subsequently choosing the topK), we are currently doing a linear-time computation over the entire ontology concept-preterminals. This is an O(n log n) computation with potentially a large constant time added to each of the similarity computations. Could be expensive if the ontology is large. We should think of smarter ways to do this topK similarity computation (trading off with a better data structure with more memory or doing a parallel topK similarity computation) in the future. Please let us know your thoughts.

fyi: We added an example in ExampleGenerator.scala to display the groundings. But it doesn't have to stay there .. 

Thanks!

MihaiSurdeanu:
@ajaynagesh: but why is runtime O(n log n)? It should be O(n), since you are scanning the list of pre-terminals once.

bsharpataz:
the sort I think

kwalcock:
It turns out I only have parallel code for KMeans clustering and not KNN.

kwalcock:
@ajaynagesh , I'm working on it.

ajaynagesh:
Thanks @kwalcock .

@MihaiSurdeanu : I think the sort makes it n log n as Becky pointed out.

MihaiSurdeanu:
I see. There are faster algorithms, O(n log k) I believe for this. We’ll fix this later.

kwalcock:
I think these have been addressed:

-The vectors should not be reloaded when we reload the system (in the UI and shell). I was unsure how to achieve this in the new config setup.
-Because of the EidosMention.remapOdinArguments, it seemed like we needed to pass those extra arguments throughout. Perhaps you could refactor this to make it cleaner

I will add a link for this, but we'll have to think about how to accout for our use, Travis, and people who just download our code.  I'm pretty sure that at least the path to the file needs to change because it would be expected to be found in the jar file otherwise.

-The embeddings are put in this location on google drive (we need to modify the readme to include this link)

PR #138, opened 2018-02-28
More Raps Tests (sent11-19)

hclent:
PR to include more unit tests for RAPs into codebase. 

For sent14-sent16, I did notice a pattern in some of the errors we are getting. See the "Add test sent16. Noticed pattern of failures." In short, Eidos fails to extract Increase Attachments to entities with sentences like "A and B will increase".

### DIFF
--- a/src/test/scala/org/clulab/wm/eidos/system/TestRaps.scala
+++ b/src/test/scala/org/clulab/wm/eidos/system/TestRaps.scala
@@ -7,7 +7,7 @@ import org.clulab.wm.eidos.text.{Inc, Dec, Quant, Causal, Origin, TransparentLin
 class TestRaps extends Test {
 
 
-  { //One Increase Event
+  { //One Increase attachments
     val sent1 = "Better and well-functioning agricultural credit and market services for both established and emerging farmers."
     val tester = new Tester(sent1)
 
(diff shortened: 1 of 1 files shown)


kwalcock:
P.S. Can you remind me what Raps is?  It must be from before my time.  I'm thinking that the file may eventually need to move in the collection of tests.  Thanks!

hclent:
@kwalcock There are just shy of 40 RAPs sentences that each need a test. I'm about half way done. For the ones that are failing, I thought (correct me if I'm wrong) that I should just add all of the new tests now, and then find patterns in the errors to integrate into the grammar.

"RAP" are "Representative Agriculture Pathways", (I think). They were the first data that the DARPA folks gave us during the seedling. You can find them [here](https://github.com/clulab/research/tree/master/wmseed/src/main/resources/org/clulab/wm/rapdocs) (assuming you have access to the research repo. If not, I can e-mail you a zip file).

Let me know if you want me to make some changes in order to have them pulled, or if you'd like me to wait until all ~38ish tests are done to make a PR.

kwalcock:
Thanks for the background.  Becky didn't object, so I'll merge.  The file is getting pretty large.  I wonder if it would help to break it into parts.  The location I was thinking of is org.clulab.wm.eidos.text.raps.  Do these texts come from a document that we should store with others on Google Drive for posterity or a rainy day?

hclent:
@kwalcock Good call on starting (a) new file(s) for these tests and also changing the file organization. I'll make sure to do that in my next PR! I think it would be a good idea to include these documents on Google Drive, so I'll go ahead and see if I can add them there.

PR #139, opened 2018-02-28
Make AntiNode and AntiEdge for things that shouldn't be there

kwalcock:


kwalcock:
I hope it's useful.  Just say if it could be more so.

Issue #52, opened 2018-02-01
Do not get the correct entities.


ZhengTang1120:
When parsing "the rapidly depreciating value of the South Sudanese Pound (SSP)", the system splitted the entity "value of the South Sudanese Pound" into "value" and "South Sudanese Pound" and failed the test.

Issue #44, opened 2018-01-30
Do not expand entities across clauses?


MihaiSurdeanu:
Becky,

We now expand entities across clausal dependencies. For example, in this sentence:
https://github.com/clulab/eidos/blob/master/src/test/scala/org/clulab/wm/CAG.scala#L27
the 3rd entity is "livelihoods being decimated", when it should probably be "livelihoods".

ZhengTang1120:
level of acute malnutrition -- test changed to match current "intended behavior," so will need to modify tests when issue closed

bsharpataz:
food insecurity in Unity

Issue #141, opened 2018-02-28
import vars


bsharpataz:
import the variables at the tops of grammar yml files (e.g. noun_modifiers)

PR #142, opened 2018-02-28
Expand ef

bsharpataz:
The real changes are 
- adding a couple of VALID_OUTGOING deps and pushing that functionality back to the base class (RuleBasedEntityFinder)
- increasing the max length of an entity and the maxHops to travel

The resulting change broke a test, and while I was poking around there I modified the test and tweaked some rules to make the extractions cleaner.  (@hclent fyi)

TB done -- need to do this -- https://github.com/clulab/eidos/issues/141

also, I think this Closes https://github.com/clulab/eidos/issues/44
closes https://github.com/clulab/eidos/issues/52


bsharpataz:
@kwalcock I went ahead and made the change, please merge when you're ready!

PR #143, opened 2018-02-28
Canonical form

bsharpataz:
feel free to close PR or request changes, but here as a PR in case it's helpful

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
@@ -142,37 +142,6 @@ class EidosSystem (
     res
   }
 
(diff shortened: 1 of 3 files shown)


kwalcock:
Thanks for the tests.  If it's OK, I'll move some code around to match what I was trying to describe.

bsharpataz:
yep!  And I'll take note and try to learn from you! :)

kwalcock:
Usually I would wait a day before messing with people's stuff, but we're in a hurry...  Common functionality remains in the base class and custom functionality is pushed down to derived classes, which obviates the need for the case statement.  You're welcome to change protected back to private.  I'm in the "private is premature pessimization" school, but it probably won't be important here.  Still, I could not resist.

bsharpataz:
Thanks Keith!!

PR #149, opened 2018-03-01
Get WMseed to work again

kwalcock:


kwalcock:
It has a configuration file now and a couple of documents to use and constructs EidosSystem in the new way.  If there once was an official set of documents, perhaps they can be added to the collection on Google Drive.

kwalcock:
This now includes cache for Travis to speed up builds.

PR #151, opened 2018-03-02
Make sure all lines make it to output file

kwalcock:
The double for loop was not working as expected.  Changed to nested foreach to "simplify".


kwalcock:
Just so that you know your code was changed.

Issue #147, opened 2018-03-01
JSON serialization has multiple instances of same event


bgyori:
For the sentence
```
The government promotes improved cultivar to boost agricultural production for ensuring food security.
```
The JSON output contains 4 copies of the same event, here with ID E:-1154077341:
```
    {
      "type": "EventMention",
      "id": "E:-1154077341",
      "text": "cultivar to boost agricultural production",
      "labels": [
        "Causal",
...

```
with the same foundBy, same arguments that have the same IDs, etc. Note that this is the JSON I obtain from `serialization.json.WMJSONSerializer`.

Is this a serialization issue or something else?

bgyori:
I noticed that I ran into this in a somewhat older version of Eidos, I'm testing with the latest to see if the issue persists.

bgyori:
Yes, issue persists with latest version.

kwalcock:
I was recently writing some serialization tests and it would not be
inconvenient for me to add this to the suite and try to see what is
happening.  If nobody objects, I'll proceed.

kwalcock:
FWIW there are actually two pairs with slightly different foundBy values

"foundBy" : "simple-np++Increase_ported_syntax_6_verb++Increase_ported_syntax_1_verb",
"foundBy" : "simple-np++Increase_ported_syntax_6_verb++Increase_syntax_1_verb_edited",

foundBy must not be used in calculating the id, and maybe that's a problem.  Why are there pairs, though?

MihaiSurdeanu:
Oh. Increase_ported_syntax_1_verb and Increase_syntax_1_verb_edited found the same event. We should remove duplicates, even when found by different rules.

kwalcock:
It doesn't appear to be the serialization.  The output seems to match the input that it gets.

kwalcock:
Here's the output from EidosShell for the sentence.  How many are considered duplicates?

entities:
List(NounPhrase, Entity) => government
	------------------------------
	Rule => simple-np
	Type => TextBoundMention
	------------------------------
	NounPhrase, Entity => government
	------------------------------

List(NounPhrase, Entity) => cultivar
	------------------------------
	Rule => simple-np++Increase_ported_syntax_6_verb++Increase_ported_syntax_1_verb
	Type => TextBoundMention
	------------------------------
	NounPhrase, Entity => cultivar
	  * Attachments: Increase(improved,None), Increase(promotes,None)
	------------------------------

List(NounPhrase, Entity) => cultivar
[...]

PR #152, opened 2018-03-02
Dup events

bsharpataz:
fix for duplicated entities and events in same span

I think fixes https://github.com/clulab/eidos/issues/147

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/EidosActions.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosActions.scala
@@ -34,24 +34,26 @@ class EidosActions(val taxonomy: Taxonomy) extends Actions with LazyLogging {
 
     val mention_attachmentSz = for (mention <- mentions) yield {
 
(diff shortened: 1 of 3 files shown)


kwalcock:
The test case is forthcoming.  Should it only check for EventMentions with duplicate IDs?

bsharpataz:
not id's prob... I guess events with the same label and span? if they have the same label and span, but diff triggers are we considering them to be duplicates? @MihaiSurdeanu

MihaiSurdeanu:
Events should be considered duplicates if:
- They have the same spam.
- They have the same label.
- They have the same arguments (so it's a recursive function).

The trigger doesn't matter.

MihaiSurdeanu:
(The trigger doesn't matter because it's captured in the event label already.)

bgyori:
I checked out this branch and tested it locally, it definitely resolves #147 thanks!

bsharpataz:
great!

PR #153, opened 2018-03-02
Add dummy constructor for Python integration

bgyori:
This was removed during a refactoring, I'm adding it back, this time with a comment.


kwalcock:
Ah, this one I understand.  Thanks for including the comment.  I only noticed unused code.

PR #159, opened 2018-03-05
first pass on adding semi-arbitrary predicate entities

bsharpataz:
@kwalcock this isn't quite ready -- there is a failing test and I can't figure out why (it's another case of the node looking fine, but the test fails....)


kwalcock:
I'll check if it works differently on my computer :)

bsharpataz:
in the meantime, I just marked it as failingTest so we can merge it

kwalcock:
21 hours later, I'm finally getting to it.

kwalcock:
The problem is that _two_ nodes are found:

  4: efficient input use
  5: service development
  6: efficient input use

It is hoping to find only one.  The tester might be extended to pick between the multiple possibilities by providing some extra information to the NodeSpec.  For instance, NounPhrase rather than Predicate.

List(NounPhrase, Entity) => efficient input use
	Rule => simple-np++Increase_ported_syntax_6_verb
	Type => TextBoundMention
	NounPhrase, Entity => efficient input use
	  * Attachments: Increase(increased,None)

List(Predicate, Entity) => efficient input use
	Rule => simple-vp++Increase_ported_syntax_6_verb
	Type => TextBoundMention
	Predicate, Entity => efficient input use
	  * Attachments: Increase(increased,None)

MihaiSurdeanu:
We need to make sure duplicate removal works...

bsharpataz:
yeah, ugh.  I keep thinking it's fixed and new ones pop up.... I'll look
into it.

kwalcock:
Somewhere I have a test for duplicates.  I'll add that sentence, assuming that it shouldn't be there.  It sitting in a PR.  I'll see if it works and then get it merged.

kwalcock:
Just to be clear, the two nodes are found by the Tester.  These nodes have different labels, List(NounPhrase, Entity) and List(Predicate, Entity), so they aren't the same as far as duplicate removal is concerned.  In the Tester, the NodeSpec (specification for a node) is only using the text and some attachments to identify the node.  That is insufficient in this case.  I will experiment with an additional argument that allows the user to add an extra filter.

kwalcock:
A filter argument has been added so that the person writing the test can choose between multiple matches of a NodeSpec.

kwalcock:
I merged with master because of a conflict, but now two tests are failing.  The cause of the first problem is the same as above, so I fixed that.  The second one I have to leave for now.  It lost a node.

[info] TestDoc7 Paragraph 6
[info] - should have correct singleton node 1 *** FAILED ***
[info]   List("Could not find NodeSpec [emergency food assistance]", "Mentions:
[info]   0: expand access to safe drinking water
[info]   1: implement livelihoods interventions
[info]   2: expand access to safe drinking water
[info]   3: FFP also partners with Catholic Relief Services to provide families in Jonglei State with emergency food assistance
[info]   4: implement livelihoods interventions
[info]   5: agricultural training for farming households
[info]   6: FFP also partners with Catholic Relief Services to provide families in Jonglei State with emergency food assistance
[info]   ") was not equal to List() (TestDoc7.scala:194)

kwalcock:
@bsharpataz when you get time (if ever), can you check on this PR?  It's getting worrisomly old.

bsharpataz:
'Food insecurity and conflict are also forcing many families to leave South Sudan for neighbouring countries.'

kwalcock:
See the "Details" link for the log of the Travis build

bsharpataz:
I'm closing this bc I think the top-down approach makes this obsolete.  We can reopen if I'm wrong

PR #162, opened 2018-03-05
Add saxon-dom to dependencies

### LINKED ISSUE #160: DOMSource cannot be processed
On certain strings, I am seeing a large number of these errors
```
Error 
  DOMSource cannot be processed: check that saxon8-dom.jar is on the classpath
```
when reading via the python interface which uses a fat JAR of Eidos and its dependencies. These errors flood the screen but don't seem to have a serious effect, the reading completes and returns results. 

I found that one example of a string that triggers this is "NOV". It looks like lowercase "nov" does too. So then I thought maybe it has to do with month names, and indeed "dec", "jan", "feb", etc. all result in the same error. Similarly, "november", "december", etc. result in the same.

So given this info, does anybody know what this could be?

bgyori:
Fixes #160 by including in build.sbt the dependency I got errors about. If this is not the preferred way to do this, please feel free to close the PR!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,6 +19,7 @@ libraryDependencies ++= {
     "ai.lum"        %% "common"                   % "0.0.8",
     "org.scalatest" %% "scalatest"                % "3.0.4" % "test",
     "com.typesafe"  %  "config"                   % "1.3.1",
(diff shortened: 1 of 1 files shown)


kwalcock:
A test case will be added from a different branch.

PR #165, opened 2018-03-06
filtering both nodes and canonicalName with stop and transparent words

bsharpataz:
This required me to mark some tests as "failing" -- I chose to do so by marking them as requiring inference (a special case of futureWorkTest?)


kwalcock:
A couple of emergency measures are in place here that will be addressed soon.

bsharpataz:
what about a separate config file, something like entityFilters.conf whose
location is found in the main config file?
I like the idea of having them "together" for easy reference and I like the
config file namespace-ness, but agree that it clutters the main config file

kwalcock:
I'd avoid the config file.  It seems unnecessary to do the formatting for what seems to be a word list and straight text is more portable.   Reloading would also work for free.  (The config file itself can't be reloaded, but things it refers to can.)  There's code in loadGradableAdjGroundingFile that can be borrowed to read a word list.  Related to that, I see that Quantifier.tsv contains the word "adaquately" with a typo.

bsharpataz:
but if there's a sep config, it could be  reloaded, and we'd have comments
available (like "theses should be lemmas" or something)

kwalcock:
It doesn't look like all the bugs of mine that you corrected are at the server.  Will you be sure to push them up and consider tagging the files that went into generating that output?  When you're not at some workshop thing...

kwalcock:
Room for comments sounds like a reasonable requirement.  Will # ever be part of a word?

bsharpataz:
don't know... it would also help contain the # of files in resources to
have a second config

bsharpataz:
(sorry -- shouldn't have used the # symbol there given the previous context
-- just meant 'number')

bsharpataz:
try now -- at the workshop all day for 3 days :)

kwalcock:
Enjoy!

kwalcock:
I fixed "adaquately" in Quantifier.tsv, so disregard.

PR #168, opened 2018-03-08
Remove gratuitous Unicode in resources and test for it

kwalcock:
Include not much used Sourcer to open files with UTF8 encoding

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/grammars/entityQuantification.yml
+++ b/src/main/resources/org/clulab/wm/eidos/grammars/entityQuantification.yml
@@ -29,7 +29,7 @@ rules:
     pattern: |
       @adverb: Quantifier?
       (?<trigger> [mention=Quantifier])  #Find the adverbial quantifier
-      @theme: Entity #Find the adjectival quantifier being quantified :)  
+      @theme: Entity #Find the adjectival quantifier being quantified
 
   - name: quantification_adjective_rule_2
     priority:  ${ rulepriority }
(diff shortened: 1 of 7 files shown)


kwalcock:
With this, it should not be necessary, as least as far as the rules are concerned, to do
export JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF8
The program will run even with ASCII encoding.  There is code in Sourcer to enforce UTF8 in case it is found in some text file that we process.  I have not checked in changes to code that makes use of Sourcer because I have not found a good way to test it yet.

kwalcock:
@bsharpataz, do you know why EidosActions tries first fromFile and then fromURL?  Are we allowing people to override the URL location with a file and falling back if it isn't there?  I'd like to make it the same as the others with just fromURL.

val source:BufferedSource = if (url == null) scala.io.Source.fromFile(new File(path)) else scala.io.Source.fromURL(url)

bsharpataz:
I think it may have been so it could be run both in the IDE and with
sbt...?  to help with finding the file in the jar...? I'm not 100% sure but
I recall that one way of running things had trouble finding files in the
resources?
but FWIW people should be able to provide their own grammar/resource files
as easily as possible

MihaiSurdeanu:
Fwiw, both the IDE and sbt should work with fromURL...

kwalcock:
I've gone ahead and rerouted Eidos file access through this Sourcer thing where our standard of encoding any necessary Unicode characters as UTF-8 is codified.  It shouldn't break anything, but I haven't found a case where it is actually necessary except second-hand in the docker container.  This might allow inclusion of the special characters I just got rid of in the yaml files without requiring the environment setting.  Everything is read as a resource via URL.  We may well have to make an exception for the file of vectors, however.  That's coming up yet.

PR #172, opened 2018-03-09
re-adding the fix

bsharpataz:


kwalcock:
OK.  In case anyone needs in meantime.

PR #178, opened 2018-03-12
Fixes #173

### LINKED ISSUE #173: Eidos not picking up entity
It seems that Eidos is not picking up entities when they are at the beginning of a sentence that is preceded by another sentence. 

- The sentence `Conflict causes displacement` works fine - the nodes and the edge are picked up.
- The sentences `X causes Y . Conflict causes displacement.` results in no entities or events picked up in the second sentence.
- If an article is added before 'Conflict', it works: `X causes Y . The conflict causes displacement.` Now, all entities and events are picked up.

These were tested using the EidosShell.

adarshp:


bsharpataz:
importantly fixes a bug in the filtering when running over docs with multiple sentences

PR #181, opened 2018-03-13
Clean up file operations

kwalcock:


bsharpataz:
@MihaiSurdeanu -- I really like working with @kwalcock !

kwalcock:
Do you mean that next to

def getTextFromResource(path: String): String =
def getTextFromFile(file: File): String =

I should e.g. add

def getTextFromFile(path: String): String =

So far I hadn't tried for completeness and only included what was being used.  The distinction between a file and resource is subtle and confusing, but I should probably add the missing ones preemptively.

bsharpataz:
@kwalcock maybe? I think I am speaking in terms of usage -- I'd love it if I can *either* specify a relative path from resources in the config *OR* an absolute path in the same line of the config (i.e. for something like a w2v file? or a stopwords file?) and the code handle either.  So I guess ideally the call in EidosSystem would figure out which it is and handle it... does it already with the leading "/"? seems like sometimes we use that "/" and sometimes we don't...?
e.g. (from EidosSystem line 67 on):
val   masterRulesPath: String = getPath(  "masterRulesPath", **"/org/clulab/wm/eidos/grammars/master.yml"**)
    val  quantifierKBPath: String = getPath( "quantifierKBPath", "/org/clulab/wm/eidos/quantifierKB/gradable_adj_fullmodel.kb")
    val domainParamKBPath: String = getPath("domainParamKBPath", "/org/clulab/wm/eidos/quantifierKB/domain_parameters.kb")
    val    quantifierPath: String = getPath(   "quantifierPath",  **"org/clulab/wm/eidos/lexicons/Quantifier.tsv"**)
    val   entityRulesPath: String = getPath(  "entityRulesPath", "/org/clulab/wm/eidos/grammars/entities/grammar/entities.yml")
    val    avoidRulesPath: String = getPath(   "avoidRulesPath", "/org/clulab/wm/eidos/grammars/avoidLocal.yml")
    val      taxonomyPath: String = getPath(     "taxonomyPath", "/org/clulab/wm/eidos/grammars/taxonomy.yml")

PR #182, opened 2018-03-13
DisplayUtils cleanup

bsharpataz:
I've been wanting to do this for a while.  In DisplayUtils over time we had accrued three "sets" of methods -- those that println-ed stuff, those that returned the string, and those that did almost the same as the println ones but for html (i.e. `<br>` and the weird html tab).  I finally consoldated them and switched to wrapper methods.  We can also remove the wrappers if people want... up to you all.

Also -- not sure what the diff between StringBuilder and StringBuffer is so I picked one....! But that can be changed if you care!

@kwalcock I used `protected` in this object, I don't even know if that's valid but seemed like something you would have wanted done, so I did, but I can remove that for sure :)

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/utils/DisplayUtils.scala
+++ b/src/main/scala/org/clulab/wm/eidos/utils/DisplayUtils.scala
@@ -8,202 +8,151 @@ import scala.runtime.ZippedTraversable3.zippedTraversable3ToTraversable
 object DisplayUtils {
 
 
-  def displayMentions(mentions: Seq[Mention], doc: Document, printDeps: Boolean = false): Unit = {
+  def mentionsToDisplayString(
+    mentions: Seq[Mention],
+    doc: Document,
+    printDeps: Boolean = false,
+    newline: String = "\n",
+    tab: String = "\t"): String = {
+
+    val sb = new StringBuffer()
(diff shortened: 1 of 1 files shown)


MihaiSurdeanu:
Looks ok!

kwalcock:
The code looks fine and I assume the output is good and I'll check it in.  If you would like to go overboard because you're not busy enough, you might try (parts of) this pattern:

new StringBuffer()
    .append("sentence #")
    .append(i)
    .append(" ")
    .append(nl)
    .append(s.getSentenceText)
    .append(nl)
    .toString

That's a good idea to combine the plain text and the html text.  However, rather than switching up the \t and \n, I'd just get the plain text and run it through a converter, in part because there are other symbols that need to be taken into account like <>&".  So

def webAppMention = asHtml(mentionToDisplayString(mention))

asHtml might look something like this (Java) code

public static String toHtml(String text) {
	return text
			.replace("&",  "&amp;" ),
			.replace("'",  "&#39;" ),
			.replace("\"", "&quot;"),
			.replace("<",  "&lt;"  ),
			.replace(">",  "&gt;"  )
			.replace("\r\n", "\n")
			.replace("\n", "<br/>\n")
			.replace("\t", "&nbsp;&nbsp;&nbsp;&nbsp;")
}

I'm still going to merge this so that you can better ignore this.

PR #183, opened 2018-03-14
fix for low and cut

bsharpataz:
we were matching "lowlands" as a decrease trigger bc the rules typically work as "starts with trigger string"

went ahead and fixed a copula rule (plus added a counterpart) while I was at it

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/grammars/modifiersTemplate.yml
+++ b/src/main/resources/org/clulab/wm/eidos/grammars/modifiersTemplate.yml
@@ -116,9 +116,21 @@ rules:
     label: ${ label }
     action: ${ action }
     pattern: |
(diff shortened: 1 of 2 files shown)


bsharpataz:
So the inc/dec gets attached to the right thing.  If there’s an outgoing
dobj, like with “A increases B” then the nsubj shouldn’t get tagged with an
increase (see companion rule added just below for this).  On the otherhand,
with something like “A increases with B” (note no dobj) etc we would want A
to be tagged (which this rule will now only fire on).

MihaiSurdeanu:
Ah, good catch. I remember seeing this ambiguity in reach a lot.

kwalcock:
Great job, linguists!

bsharpataz:
hahaha -- "by your powers combined..." @MihaiSurdeanu any captain planet in
Romania?

PR #184, opened 2018-03-14
Turn groundings and word2vec into classes and move out of EidosSystem

kwalcock:
There are now AdjectiveGrounding and OntologyGrounding which are nearly self sufficient, leaving EidosSystem with its more important things to do than manage them.


kwalcock:
This passed the tests and also ran through the DocX.txt test documents with the exact same output as master, so I'll go ahead and check it in.  Some of these changes are getting in the way of the next updates.

PR #187, opened 2018-03-15
modified extraction to make a corpus of all annotatedDocs, not one pe…

bsharpataz:
…r annotatedDoc

I think this fixes the issue that @adarshp found tonight


kwalcock:
You beat me to it!  Thanks for using FileUtils.  Well, actually if it is Sourcer instead of Source, then UTF8 standard will be maintained.  Besides, I had to fix my extra bug.

kwalcock:
The 200,000 line output for DocN.txt is a little hard to work with.  I broke the file up into ten pieces and am putting them on Google Drive.  This makes more sense for ExtractFromDirectory.  We should have other programs for ExtractFromFile.

PR #188, opened 2018-03-15
Add logging for config paths and file access

kwalcock:
Eidos diagnostic output should look like that from other components


kwalcock:
I guess it depends on whether it it's really output or just some kind of commentary.  These first cases were fairly clear.

PR #189, opened 2018-03-15
Use Real- and Fake- WordToVec instead of ifs

kwalcock:
Reconcile eidos.conf values and defaults in EidosSystem.  Prepare to document in readme.


bsharpataz:
what happens if you try to use a fakeW2V?

kwalcock:
It just returns 0 for similarity and Seq.empty for similarities without ever instantiating a real w2v which doesn't like being empty.  Functionality shouldn't have changed.

kwalcock:
FWIW it's a null object pattern.

PR #195, opened 2018-03-16
Merge ents

bsharpataz:
Key changes:
- modifications to entitiy extraction methods and avoidLocal.yml
- temp disabled 2 RAP tests (they will pass again with final PR)
- added a lot of stop words
- smarter isCoord for splitting (or not) coordinated entities.... we can change this later if we want!


kwalcock:
At some time I can imagine STOP_NER being moved out of source code and something like

def removeNER(ner: String) = EidosSystem.containsStopNER(ner)

but not now.

bsharpataz:
k, sounds good!

PR #197, opened 2018-03-16
Mithun testdoc2

mithunpaul08:
7 passing tests . rest are dependent on @ZhengTang1120  or @bsharpataz 's fixes


bsharpataz:
I think this handles the conflicts and issues with tests.  I fixed a Vikas test with a new rule while I was at it... hope you all don't mind

kwalcock:
Yipee.  This is an old one.  There are some comments in TestDoc2 that I wonder about like
//this is passing from mithun's side, but will pass once zheng fixes the "merging entities bug"
and some others where your name is mentioned.  What's the status on those?

bsharpataz:
I’ll try to check today

bsharpataz:
I will merge as is (assuming travis passes, which I think it will based on local runs) even with the comments the way they are.  I tried to simply mark the failingTests that were dependent on the merge code as passing, but it's a bit more complex apparently.  The merge code is fine-ish, but maybe needs some work, or the rules, or some combo of the 2

kwalcock:
In all your spare time would you like to take a peek at the 15 branches and see if you recognize any that are now superfluous or that need to be reincarnated?  I should have asked everyone this morning.
I'm really glad to see this one moving along.

bsharpataz:
sure thing!

PR #199, opened 2018-03-18
"add 'level' as Transparent, re-order rules in entityQuantification.yml

Fan-Luo:
I revise test: val malnutrition = NodeSpec("levels of malnutrition", Quant("truly", "alarming")) to
val malnutrition = NodeSpec("levels of malnutrition among children", Quant("alarming", "truly"))

Does it make sense?
I do not know how to attach adverbs, "truly" in this case.


bsharpataz:
Yes -- changing to Quant("alarming", "truly") is a good catch
re: adverbs -- likely rules need to be adjusted to capture them similarly to how quantifiers are captured with inc/dec rules

Fan-Luo:
Sure, I will

Fan-Luo:
Should all adverb (-ly) be sorted after adj? or should they also be alphabetized as adj?

kwalcock:
Fan, let's alphabetize them all together rather than in two categories.  If that's the wrong answer, I'll fix it.  Only worry about the ones you've added or modified.  If it looks like other ones are out of order, we'll run it through sort.  Thanks.

Fan-Luo:
ok. I'll simply sort all of them, which is easier.

bsharpataz:
I much prefer this change over the one involving transparent.yml -- I'm planning on removing that file @kwalcock FYI

PR #200, opened 2018-03-18
Move from ec2 to gce for testing

kwalcock:
This gets us more memory for the test and .sbtopts is compatible.  It costs more startup time, but the tests run in ~1/2 the time.


MihaiSurdeanu:
Nice. Is this part of the free plan?
If so, can you please add it to processors and reach too?

kwalcock:
Although it is documented at travis-ci.com rather than travis-ci.org, it does seem to be free.  Processors and reach were actually already using it and I hadn't copied it over.  What they aren't yet using is the cache.

PR #203, opened 2018-03-21
Combine several rounds of doc8 manipulations into something for discussion

kwalcock:
These should not actually be merged.  They are probably more for practice than for production.


kwalcock:
This stuff is almost certainly incorrect, but mark it down as practice.  Maybe we can discuss them if you have another tutorial type meeting.

Past participles used as adjectives are too often taken for past tense verbs.  Here some unambiguous adjectives are used.  There is also a correlation which is not done correctly.

PR #207, opened 2018-03-22
Added warning about tildes in README

adarshp:


kwalcock:
I will also improve the error message so that the cause, file not found, is obvious.

PR #208, opened 2018-03-22
Improve error handling for missing files

kwalcock:
Make sure Sourcer is used


kwalcock:
This should produce a FileNotFoundException and show the unfound file/path, complete with any ~ which will not have been replaced.  It should produce the error before bothering to load the EidosSystem unnecessarily.

kwalcock:
Fair enough.

PR #209, opened 2018-03-22
Add output table for words in sentences

kwalcock:
This is so scala session is not necessary for debugging
Rename last agro to eidos
Escape HTML


kwalcock:
The useful change may not be desired and can be disabled.  I wanted to see the additional information about the words without starting a Scala session.  The other changes are more cosmetic.

PR #211, opened 2018-03-22
Modified ExtractFromDirectory app to process files in parallel (taking advantage of multiple CPU cores)

adarshp:


kwalcock:
Do you know offhand whether

println(s"Extracting from ${file.getName}")

is thread safe?  Will the output be intermixed?

kwalcock:
Looks like it is documented as synchronized, so disregard.

kwalcock:
@ajaynagesh, you were making a similar change.  Does this work well with what you have?

ajaynagesh:
@kwalcock I noticed that if all the threads write to the same file on disk, the parallelization is very slow due to locks for write being shared. Better to write to separate files. At least this is what I observed in my case.

kwalcock:
Thanks.  I think this one has a separate output file for every input file.

kwalcock:
Somewhere in EidosSystem there is an ordering dependency.  If multiple texts are fed into the system, the result will be dependent on their ordering.  The parallel version can result in a different order of files and it can get a different answer on two runs.  Some kind of state is being preserved.

bsharpataz:
the result of "canonicalName" ?

kwalcock:
The difference is
        "entity" : "O",
versus
        "entity" : "LOCATION",

One version
      }, {
        "@type" : "Word",
        "@id" : "_:Word_378",
        "tag" : "NNP",
        "entity" : "O",
        "startOffset" : 496,
        "endOffset" : 501,
        "lemma" : "Green",
        "chunk" : "I-NP"
      }, {
        "@type" : "Word",
        "@id" : "_:Word_379",
        "tag" : "NNP",
        "entity" : "O",
        "startOffset" : 502,
        "endOffset" : 506,
        "lemma" : "Belt",
        "chunk" : "I-NP"
      }, {

Other version

      }, {
        "@type" : "Word",
        "@id" : "_:Word_378",
        "tag" : "NNP",
        "entity" : "LOCATION",
        "startOffset" : 496,
        "endOffset" : 501,
        "lemma" : "Green",
        "chunk" : "I-NP"
      }, {
        "@type" : "Word",
        "@id" : "_:Word_379",
        "tag" : "NNP",
        "entity" : "LOCATION",
        "startOffset" : 502,
        "endOffset" : 506,
        "lemma" : "Belt",
        "chunk" : "I-NP"
      }, {

kwalcock:
There is test code on a different branch that attempts to isolate the problem, not very successfully.  The problem isn't with ExtractFromDirectory itself.  So, I'm going to check this in.  These notes are hopefully sufficient paper trail.

ajaynagesh:
@MihaiSurdeanu : Is the output of the histograms due to the effect mentioned in this thread ?

kwalcock:
Did you get some strange results?

PR #215, opened 2018-03-23
grammar changes from mitre stuff

bsharpataz:
@kwalcock @MihaiSurdeanu 
This is the last of my PRs -- sorry so late.  I tried to keep this current with master, so hopefully minimal issues?


bsharpataz:
resolved the merge conflicts, I was pretty careful, but there is one grammar rule merge that may cause a travis issue with someone's test.  If so, I'll fix it, but fingers crossed!

bsharpataz:
@kwalcock can you help me understand what the necessary change in DisplayUtils is?  I think these mods were from you (and I"m sure they were good!) but now I don't see how to specify the tab etc in the methods (i.e. to give the flexibility for the html version).  Thanks!

bsharpataz:
ok @kwalcock it's **finally** fixed!! 😄 👍 😢

kwalcock:
Here goes.  Thanks for dealing with the conflicts.  Fingers crossed.

kwalcock:
Do we have a strategy for dealing with comparatives and superlatives?  In the increase triggers I see greater, but not greatest, and then now highe(r|s) which is probably intended to match highest.  That seems like the only superlative so far.

MihaiSurdeanu:
I think we should add superlatives, no?

bsharpataz:
No real strategy, greater should likely be great, as rules match ^trigger.
For short words like high and cut, i saw us matching *other* words like
highlands, so I added specificity.  That said, if we were much more
careful/robust/consistent (myself very much included) that would be awesome!

bsharpataz:
for most we have just the stem and superlatives would kinda fall out (i.e.
if we have ^large we get larger and largest for free.  We should go through
and make them more consistent/explicit though. I can do this when I add the
new triggers (working a little on the trigger expansion this weekend)

PR #219, opened 2018-03-25
Check both directions for undirected relations

kwalcock:


kwalcock:
This *should* handle the issue.

PR #221, opened 2018-03-26
Stopfilter

bsharpataz:
the moving filtering methods to OntologyGrounder


kwalcock:
@ajaynagesh This has been checked in and will probably help with the New Zealand and FAO problem.  I am still writing tests for it, though.

bsharpataz:
sure -- yeah with this as a record of our decision

kwalcock:
Sorry to nag.   Was the conclusion to do stopword filtering before the finding events?  So, in "Australia causes kiwi production." we filter out Australia before it can be used in the causal relationship.  Is that correct?

bsharpataz:
yes -- it happens here:
https://www.dropbox.com/sh/ncx9kpqxkzxr865/AADSgizBCNhI-cALSnn1XCQOa?dl=0

PR #222, opened 2018-03-26
Write filter test

kwalcock:
Add ORGANIZATION to list, alphabetize list, streamline filter to not produce a collection


kwalcock:
@bsharpataz Thanks for sorting out the merge problem.  Your code was changed slightly for efficiency.  IntelliJ keeps telling me to use "exists" more.  Some of this functionality is not so obviously part of the Grounder anymore.  Moving it elsewhere is not a high priority, though.

PR #223, opened 2018-03-26
Test doc7 egoitz

EgoLaparra:
I have marked a Correlation in TestDoc7 Paragraph 1 as futureWorkTest because there is not a clear path between the entities.


kwalcock:
It may be that in the meantime the webapp shows this information that you were looking at.  

P.S.  If you want to recycle fors, they can be written together: for (c <- s.lemmas; cs <- c) println(cs)

-    for(s <- doc.sentences){
-      for (c <- s.lemmas) {
-        for (cs <- c) {
-          println(cs)
-        }
-      }
-      for (c <- s.chunks) {
-        for (cs <- c) {
-          println(cs)
-        }
-      }
-    }

PR #226, opened 2018-03-27
Pipeline

bsharpataz:


kwalcock:
Is this the one where you noted some return statements?  I'm pretty sure that the ones in isCAGRelevant are superfluous.

PR #228, opened 2018-03-28
Man-Nature and Natural Resources additions

ajaynagesh:
- Keeping the toy_ontology.yml as it is
- Adding the un_ontology.yml file (which can be used in place of toy_ontology (in code currently) when the un_ontology is in good shape)
- Added todos in node additions for easy checking


MihaiSurdeanu:
@bsharpataz: Ok with you too?

bsharpataz:
@MihaiSurdeanu @ajaynagesh added some comments but OK for moving forward for sure

PR #229, opened 2018-03-28
Do not include triggers in mention duplicate test

kwalcock:


kwalcock:
@ZhengTang1120, it looks like you noticed this being ready.

PR #234, opened 2018-03-28
Zheng' tests, merge mentions.

ZhengTang1120:


kwalcock:
It is my turn to be confused.  There is only a very small number of changes in this PR.  Did that Revert not work as expected or are the local and server versions somehow mixed up?

ZhengTang1120:
I am confused too, I assume that it only shows the changes compare to my previous PR(which was reverted). Because I checked the my unit tests, they are labeled as "failingTest" in master but "passingTest" in my PR.

kwalcock:
The changes haven't been merged into master yet, so they wouldn't be changed to passing there.  You are expecting a lot of changes to EidosActions.scala, right?

ZhengTang1120:
Yes

kwalcock:
This did not get all the changes.  I'm trying again.  The version you want is in the zheng_test branch with the number 160d2d1, I assume.

ZhengTang1120:
I think I got the problem here. Because GitHub thinks that my branch is behind master, so it treats all my changes as happened before last commit in master.

kwalcock:
I'm not sure about that, but I manually moved the changes to a branch kwalcock_redo_zheng_test that have all the changes up through 160d2d1 and it has passed the tests locally.  Your branch can be behind because Becky has since checked things in and that would explain that part.

kwalcock:
I take that back.  She didn't do a PR yet on master.

ZhengTang1120:
I found a way to solve this, I made a new branch from master and then apply all the changes to this new branch. It seems works.

ZhengTang1120:
One thing I want to make sure is that besides the un_ontology.yml, there is no changes from other people, right?

kwalcock:
I was wrong again.  Becky had made the change you found.  Yes, I think it's the only one.  I'll compare your result with mine before OKing the PR.  Thanks.

ZhengTang1120:
Sorry for the inconvenient and all these extra work. I think my new branch just did the same thing as you do. I already put a PR there. Thanks!

bsharpataz:
Guys, thanks for working through this!

kwalcock:
I'll merge it momentarily after double checking then look at what else is waiting.

PR #236, opened 2018-03-29
GPEs and fixing comments of prev. PR

ajaynagesh:


ajaynagesh:
Adding reviewers .. forgot yesterday ...

MihaiSurdeanu:
This is fine and should be merged. But we need to come up with a consistent strategy for handling singular vs plural nouns... This should be discussed at the next reading meeting.

PR #237, opened 2018-03-29
Zheng test

ZhengTang1120:


kwalcock:
It matches what I had except for one commented out line which is removed in one version.  So, here goes.  Thanks for double checking.

PR #238, opened 2018-03-29
Filter for no-NN entities

bsharpataz:
I added a filter for entities found in the entitiyFinder that don't have a noun:
- before expansion, I check to see if the entity contains a noun OR
- if it ends with an adj and the *next* word is a noun (i.e., if it got filtered out bc it's a trigger needed later)

I also added a test in TestEntityFinder to demo the desired behavior, but please let me know if it's obviously incomplete.

Thanks!


bsharpataz:
yeah -- I thought about that, but it would be such a looooong condition,
seemed more readable that way... but I could maybe still split it on lines
with comments?

PR #239, opened 2018-03-29
add brokenSyntaxTest as an type of ignore

ZhengTang1120:


bsharpataz:
@kwalcock I asked for this, Mihai approved at the mtg and I would like access to it :) so I'm going to merge....

PR #240, opened 2018-03-29
Update README.md with Github syntax highlighting

adarshp:


kwalcock:
Cool.  I learn something every day.  The list is at https://github.com/github/linguist/blob/master/lib/linguist/languages.yml, so maybe I'll add a few more for completeness.

adarshp:
Yep, I knew about a few other languages, but only discovered that it could hand JSON-LD today!

PR #242, opened 2018-03-29
Update README.md with new function calls

adarshp:
Changed the function calls in the beginner example to match current code


kwalcock:
Although I wonder whether it should match the currently released/published version instead.  In any case we're about to update the release, so this is a good thing, I think.

PR #244, opened 2018-03-29
Mark this as passing in the meantime

kwalcock:
There should be some better tests, probably calling mergeAttachments directly, but those are forthcoming.


bsharpataz:
thanks! it's a start!

PR #250, opened 2018-03-31
sanitizing canonical name

bsharpataz:
fyi


kwalcock:
What does sanitizeWord do in 10 words or less?

bsharpataz:
it lower-cases and converts numbers to something like `xnumx` (normalizes) -- does not lemmatize or stem though

kwalcock:
I'm rejecting your answer for excessive length.  You are welcome to revise and resubmit.

PR #251, opened 2018-03-31
Test doc8 egoitz

EgoLaparra:
My last round:

- The main change of the PR is _argTokenInterval_ in EidosActions. It seems to work without breaking any test. I have had to change a test in TestDoc2 but it was failing after adding _nmod_among_ to INVALID_OUTGOING in RuleBasedEntityFinder.
 
- I have added the variable _passive_mark_verbs_ to explicitLinker for verbs like "marked by". I have only added four of these and I'm not sure if they are correct. Maybe this could be improved by a native speaker :sweat_smile:

- I have marked three of Ajay's tests as passing because they have been solved along with one of mines.


EgoLaparra:
To remove underspecified event mentions we were grouping the events by the full token interval. But this didn't work for cases like the following:

`Exacerbated by climate extremes and a reduction in planted area, the national cereal gap has widened`

In this case, there are two causal relations but they have the same token interval, between _Exacerbated_ and _gap_, so we were grouping both as single one.  Using the intervals between the arguments seems to be more accurate.

MihaiSurdeanu:
Ok, thanks!

PR #252, opened 2018-03-31
Adding Product pre-terminal .. however I am not sure where to stick it in the ontology

ajaynagesh:


kwalcock:
Hi Ajay.  I don't know the answer to your question, but I will trade you for a different one.  What is a "presentation material information product"?

ajaynagesh:
Hi Keith, 

All I know is that, it is some canonical entity which occurs 722 times in the corpus. There is a lot of noise. It sounds weird :-) . But since it related to product, I put it in here. However, we could remove it.

MihaiSurdeanu:
@bsharpataz: can you please see my comments on this PR? Should we merge this AFTER we release v.0.2?

bsharpataz:
Yes, lets merge this after.  I was planning to ground to the old ontology,
but can ground to the current (pre-merge) new one if you prefer

ajaynagesh:
@bsharpataz : removed the product node from the ontology 

also checked the current ontology to see if it does not barf (due to various comments etc ... ) 
I ran the test `TestOntologyGrounder` and also `apps/ExampleGenerator.scala` .. both go through file with the current changes (and the product pre-terminal removed ) ..

MihaiSurdeanu:
Ok. Please merge.

PR #259, opened 2018-04-02
Added a guard to prevent out of bounds exceptions in EidosEntityFinder

adarshp:


kwalcock:
Thanks.  I should have caught that.  If the entity goes up to the end of the sentence, the next tag won't exist.

adarshp:
I took the liberty of merging it since it seemed like both of you approved of the change.

kwalcock:
Thanks.  This problem probably showed up quickly if you have sentences without the final period.

PR #260, opened 2018-04-02
Added a new app to enable extracting JSON-LD output from a single file. The reason for this is that I frequently want to test Eidos output for a single file, and it is a pain to make a new temporary directory with a single file in it every time I want to do that. My intuition tells me that this may not be the optimal way to do it, and there is some redundancy in two places - one, in the code of ExtractFromDirectory and ExtractFromFile, and two, between the app names ExtractFromFile and ExtractFromFile2. Any thoughts? Should we keep these separate or do some refactoring?

adarshp:


adarshp:
Sorry for the insanely long pull request message - I did it from the command line, and I thought that it would split automatically.

adarshp:
I like the second (`ExtractFromFile2 -> ExtractJsonLDFromFile`)

How about this for the first: 

`ExtractFromDirectory -> ExtractJsonLDsFromDirectory` (I added an _s_ to emphasize that multiple JSON-LD files will be produced)

As for the third, I think it's good to put Directory in there, but the Txt part is not so informative since the first function (`ExtractFromDirectory/ExtractJsonLDsFromDirectory`) also takes in .txt files.

How about `ExtractSingleJsonLDFromDirectory` instead?

Alternatively, should we combine the apps and use a command line argument to specify whether to output a single JSON file or multiple JSON files?

kwalcock:
Would it be OK to hold off on the renaming of the old programs (in case someone is using them, for example, and so that a little more planning can be done) and for the time being just rename yours in the style that is planned for the others?  Other people would probably find it useful to have.

adarshp:
That's a good idea. I'll rename `ExtractFromFile2->ExtractJsonLDFromFile`.

adarshp:
How about these changes?

kwalcock:
Those are fine.  Not being able to keep my hands off things once I open them, I was making a few more changes.  With 9 Apps, sometime soon we might want to start formalizing and produce syntax errors for unexpected output, etc.  For now I am just making sure we're using UTF-8 and logging file accesses, etc.  I'll check it in soon.

kwalcock:
I merged very clumsily, but this is what I had in mind so far.  There are only minor changes to the main App.

adarshp:
Looks good to me!

Issue #262, opened 2018-04-03
IndexOutOfBoundsException when reading multiple documents


adarshp:
When I try to read the 52 MITRE documents with ExtractFromDirectory, I run into the following error after a certain number of documents are processed.

```
java.lang.IndexOutOfBoundsException: 4
	at scala.collection.immutable.Vector.checkRangeConvert(Vector.scala:123)
	at scala.collection.immutable.Vector.apply(Vector.scala:114)
	at org.clulab.wm.eidos.EidosActions.$anonfun$mergeAttachments$4(EidosActions.scala:190)
```
(I have suppressed the rest of the stack trace for readability)

This occurs whether I keep or remove the parallelism in ExtractFromDirectory. The issue seems to be in the mergeAttachments function in EidosActions - in this line:

https://github.com/clulab/eidos/blob/676944d2afde1a4fa75298b98dc34117aca88923/src/main/scala/org/clulab/wm/eidos/EidosActions.scala#L190

the index is out of bounds. Was it @ZhengTang1120 who wrote the function? If so, could you (@ZhengTang1120 ) take a look at it and see what's going on?

kwalcock:
Please continue to pound on it.  Can you remind me where these 52 documents are stored?  Thanks.

adarshp:
It's in the shared Google Drive folder called WorldModelers: https://drive.google.com/open?id=1cHJIfQTr0XE2CEqbo4POSm0-_xzrDP-A

it's under evaluations/6month-eval/mitre-17Feb2018/DocumentSubsetArchive.zip

bgyori:
@adarshp are you running on the txt files that I uploaded here? https://drive.google.com/drive/folders/1ECQNUej2O3_W9NmlA-wyPkwY1UhW_qnD
I have read all of those with Eidos via INDRA (which calls Eidos classes directly rather than the CLI you are using here) and it worked well, though it takes a bit of time.

adarshp:
@bgyori Yes, I am running it on those. When did you run Eidos on the documents? I was able to do it a couple of weeks ago just fine, but when I tried to do it again today I was not able to. The mergeAttachment function where the error seems to be popping up was  first added 6 days ago and has undergone more recent modifications.

bgyori:
I ran them on 3/27 so yeah, it's possible that something changed since then.

kwalcock:
[...]

PR #263, opened 2018-04-03
Try a fix for issue #262

kwalcock:


kwalcock:
This probably isn't quite ready and should have some tests.  The first time we did this I was just happy to get the merge to work.  I left in at least one TODO.  AFAIK, filterAttachments returns multiple attachments in the order it received based on the keys of a groupBy function.  It is not clear to me that the first one should be chosen to locate the "template".  It's unfortunate that we don't have a good Mentions constructor for this case and that the attachments have to be added on.  I think that the existing ones may need to be removed before the winners are added back... unless it is important to keep the other ones because they are guaranteed to match the rule used and other internal data.  Maybe .head should just be replaced by something better then.

kwalcock:
This is the draft for which I will write a few simple sanity checks.

kwalcock:
OK, @bsharpataz @ZhengTang1120 I'm ready to let this go.  You might want to take a quick look to see if it still does what you intended.

bsharpataz:
OK -- I'll run some sentences thru tomorrow -- but it doesn't crash on that
doc anymore I presume?

kwalcock:
Not on my computer ;-)  All the tests still pass, but it might get different answers on some edge cases.

bsharpataz:
yep!  still does what we wanted, so yay, let's merge and see how long it takes for the next bug! :)

kwalcock:
Good point.  I should spell those out then: less than and greater than or equal.  Maybe it should have been ge.  I'll change it.  Doesn't everyone learn assembly language?

kwalcock:
I spelled them out.  This code will soon move to the Attachment class as a compareTo function, I think.  Some Attachments are inherently bigger than others.

PR #268, opened 2018-04-06
Entities

bsharpataz:
poorly named branch, but oh well.

@bgyori simply an fyi -- though of course I welcome your comment/review


bgyori:
This looks okay to me though I can't fully tell what effect it will have on sentences. If you merge it I can quickly assemble a JAR and run some more reading.

kwalcock:
I'm still at Becky's mercy on the rule changes, but it looks reasonable and the tests still pass.  Here you go, Ben.  Thanks, Becky.

Issue #269, opened 2018-04-07
Hearst patterns extracted as causal


bgyori:
I noticed that the `ported_syntax_noun_Hearst-Causal` rule extracts Hearst patterns that could be useful for finding hyponyms (I just learned about what all this means in the last 10 minutes) but I don't think these should be extracted as `Causal`. For instance,
- dietary alternatives, such as other livestock
- emergency livestock support including provision of feed

currently produce causal relations.

bsharpataz:
Ha! Yeah no joke, guess the “porting” was a bit careless at times... I’ll
def remove/move it soon

PR #271, opened 2018-04-07
hearst++

bsharpataz:
- removed the problematic hearst rule as well as some commented out ones.  
- added a file for storing the unused rules in case we want them later, it is not called by the grammars.  
- fixed a direction of a rule

@bgyori fyi

closes https://github.com/clulab/eidos/issues/269


bsharpataz:
hahaha :) I think about this most of the times I use my initials somewhere

Issue #257, opened 2018-04-02
Canonical name handling formatting-related characters


bgyori:
I found that the canonicalName in the JSON-LD output sometimes contains things like list bullet points. I'm sorry I don't understand how canonical names are constructed deeply enough to make this change myself, so I'm just reporting the issue.

Here is a case where the bullet point character is not stripped off in the canonical name:
```jsonld
       "@type": "Entity",
       "@id": "_:Entity_207",
       "labels": [
         "NounPhrase",
         "Entity"
       ],
       "text": "\u2022 The wetter",
       "rule": "simple-np",
       "canonicalName": "\u2022 The wetter",
```

Here is a more problematic case where the actual string is lost and only the bullet point (arrow) character remains as the canonical name:
```jsonld
      "@type": "Entity",
       "@id": "_:Entity_1780",
       "labels": [
        "NounPhrase",
         "Entity"
       ],
       "text": "\u27a4\u27a4 Across Central Asia",
       "rule": "simple-np",
      "canonicalName": "\u27a4",
     ...
```

bsharpataz:
thanks!  feel free to go ahead and assign me

On Mon, Apr 2, 2018 at 12:15 PM, Benjamin M. Gyori <notifications@github.com

PR #273, opened 2018-04-07
added weirdly parsed unicode lemmas to the avoidlocal

bsharpataz:
to help with not sucking them into entities in the first place

I think this closes https://github.com/clulab/eidos/issues/257

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/grammars/avoidLocal.yml
+++ b/src/main/resources/org/clulab/wm/eidos/grammars/avoidLocal.yml
@@ -37,7 +37,7 @@ rules:
     type: token
     pattern: |
       # avoid xrefs
-      [tag=/NN|CD/ & word=/\\/] [word=/^u[0-9]/]
+      [tag=/NN|CD/ & word=/^\u27a4/] | [word=/^\u2022/]
 
 
   # OURS HERE:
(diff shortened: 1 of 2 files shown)


kwalcock:
I'm going to try to add some unit tests for this on based on #257.  Encodings can be tricky and as we saw earlier, even change based on environment variables.  The escaping often goes wrong, too.  I'd like to be sure that this one doesn't stop working without us noticing.

kwalcock:
@bsharpataz, to what extent could you observe this working?  I probably have a bad test.  As far as I know, the word is only represented by characters including a "u" when it is serialized to JSON or is sitting in source code.  word=/^u[0-9]/ should not be able to match an internal representation.  I would think it is more likely something like word=/^\\u2022/ for the bullet, but I haven't seen that work yet, either.  The test includes some diagnostic output.  Can you take a peak before I search farther?  It still gets EventMentions with the symbols in the canonicalName (on my computer).

bgyori:
I just also tested it on one of the actual documents and the fix didn't seem to work there.

kwalcock:
Then it's not just my computer and I'll investigate further.  Thanks, @bgyori.

bsharpataz:
ok -- sorry I was basing my diagnostic on the webapp outputs
so... @kwalcock are you taking over?

kwalcock:
Sure, as long as it seems I haven't misunderstood.  I'll check the webapp output and compare to println.  Thanks.

kwalcock:
I didn't change much, but the test switched from failing to passing.  There aren't any canonicalNames among the 52 files that include the special characters and everything else still passes.

PR #274, opened 2018-04-08
Fixed one unit test

adarshp:


kwalcock:
That looks like a good idea to mark the old test as brokenSyntaxTest and then to also include a similar test that does pass.  I probably wasn't so thorough myself and at some point should probably look through and see where the syntax might have been fixed for some of these tests as we change between processors.

PR #275, opened 2018-04-09
Ajay tests: TestDoc-3 and TestDoc-5

ajaynagesh:
- 1 passing in TestDoc-3
- 3 passing in TestDoc-5
- new rule for causal expressions " .. has contributed to ..."
- minor changes in triggers.yml


kwalcock:
It looks like there was a side effect on test Raps_28.  Can you check it out?  Are you able to verify it locally?

ajaynagesh:
Yes .. I am able to replicate this .. 

Its because "intens" is added as an increase trigger .. however "intensive cultivation" is added as a NodeSpec... Changing them to 

`val cultivation = NodeSpec("cultivation", Inc("intensive"))` 

Will this work @bsharpataz ?

bsharpataz:
sounds reasonable to me!

PR #276, opened 2018-04-09
tech, finance/economic (some of them), infrastructure

ajaynagesh:


kwalcock:
My name isn't on this one, so this is just a comment.

Is it easier to include the quotes around all the strings?  Wikipedia says, "Strings (scalars) are ordinarily unquoted, but may be enclosed in double-quotes ("), or single-quotes (')."  It seems like our files are diverging a little in format.  taxonomy.yml and transparent.yml do not use the quotes.  Some other files are mixed internally.

Have we standardized on British or American spelling or should we include both?  In this file I see British centre and harbour but American mechanization.  Does it matter than livelihood appears twice?

bsharpataz:
re: quotes -- consistency is our friend, so I'd be all for going quoteless
re: british vs american, we just want to help the text lock onto a node,
I'd prob include both to give best odds to w2v (but may be wrong here!)
re: livelihood -- prob ok... again, we're mainly trying to establish a good
"word cloud" to help w2v lock on.  for that matter, since we have both
"food security" and "food insecurity" we have "food" twice, etc

ajaynagesh:
Thanks @kwalcock and @bsharpataz for your inputs! Removing the quotation marks around the examples in the ontology

kwalcock:
Is "nAtuRAl ReSouRceS" some kind of test data?  It looks like you are hoping for Mihai's opinion, so I'll not merge it yet unless someone says to.

PR #279, opened 2018-04-10
Change .sbtopts to .jvmopts because it works on Windows

kwalcock:
Can some people try this out?  Namely, rename .sbtopts to .jvmopts on your Macintosh or Linux computer and see if it works just as well?  It appears to be one way to keep the source distribution compatible across operating systems, assuming it works.  Thanks for trying.


bsharpataz:
bsharp@Beckys-MacBook-Pro ~/github/eidos (bullets)]$ sbt test
Unrecognized option: -J-Xmx3g
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

kwalcock:
Sorry, just -Xmx6g.

bsharpataz:
I think it works -- set it to 1g and got a bunch of java heap space OOM
errors :)
not the *most* thorough diagnostic, but.....

bsharpataz:
and back at 6G works swimmingly

PR #282, opened 2018-04-18
Cross compilation for scala 2.11 and 2.12

marcovzla:
This PR adds cross-compilation to the release process so that two artefacts are published, for scala 2.11 and 2.12. Depends on clulab/processors#237

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ dist: trusty
 language: scala
 scala:
   - 2.12.4
+  - 2.11.11
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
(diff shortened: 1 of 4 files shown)


marcovzla:
Tests will fail until a cross-compiled version of processors is released. Please do not merge yet.

kwalcock:
I'm about to check something in using 7.2.2.  That part seems to work well.

kwalcock:
There is code which is incompatible with Scala 2.11.11.

kwalcock:
It is very likely code that I added without knowing of any requirement for backward compatibility.

kwalcock:
Will this then double the time that Travis needs to run its tests?  That would be unfortunate.

marcovzla:
The tests should run in parallel

PR #286, opened 2018-04-22
W2v access

bsharpataz:
@kwalcock feel free to refactor/request stuff.  we wanted to give HMS easier access to w2v for string similarity that isn't a part of the regular pipeline.  I thought about putting the method in the Grounder stuff (i.e., adding a version of the method to the trait and then each of the classes that extend it, but I wanted it to barf if the w2v wasn't loaded.  All that to say -- that's why I put it here, but feel free to move.

@bgyori -- sorry I took so long to PR this, I had it done a few days ago and just forgot.  PLease forgive the delay!  Here's a simple example of usage:
```
object testme extends App {
  val es = new EidosSystem()
  val sim = es.stringSimilarity("cat in the house", "dog in the house")
  val sim2 = es.stringSimilarity("cat in the house", "dog in the yard")
  val sim3 = es.stringSimilarity("groundskeepers are cool", "dog in the house")
  println(s"$sim\t$sim2\t$sim3")
}
```
This prints the following:
```
0.4281447618301553	0.35720187497529654	0.17503918311177405
```

similarity is based on average pairwise similarity, and each similarity ranges from -1 to 1 (higher = more similar)

hope it's helpful!

@MihaiSurdeanu fyi

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
@@ -1,7 +1,6 @@
 package org.clulab.wm.eidos
 
 import com.typesafe.config.{Config, ConfigFactory}
-
 import org.clulab.odin._
 import org.clulab.processors.fastnlp.FastNLPProcessor
 import org.clulab.processors.{Document, Processor, Sentence}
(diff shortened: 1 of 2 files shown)


bgyori:
Great, thanks! I will try this after Thursday and let you know if I run into any issues.

bgyori:
Just tested this and it works as expected, thanks!

PR #287, opened 2018-04-23
Sort attachments for consistent output

kwalcock:


kwalcock:
This PR will probably have more added to it, because output is still not consistent.

Issue #288, opened 2018-04-27
Classpath problem with sbt


kwalcock:
The kwalcock-classpath branch has a test TestSerialization with a few lines that cause problems.  (See PR #289.)  They are commented out in the master branch.  Basically, the line `val copy = decoder.readObject()` causes

```
[info] TestSerialization:
[info] Standard Serializer
[info] - should serialize and deserialize mentions *** FAILED ***
[info]   java.lang.ClassNotFoundException: org.clulab.struct.DirectedGraph
[info]   at java.net.URLClassLoader.findClass(Unknown Source)
[info]   at java.lang.ClassLoader.loadClass(Unknown Source)
[info]   at java.lang.ClassLoader.loadClass(Unknown Source)
[info]   at java.lang.Class.forName0(Native Method)
[info]   at java.lang.Class.forName(Unknown Source)
[info]   at java.io.ObjectInputStream.resolveClass(Unknown Source)
[info]   at java.io.ObjectInputStream.readNonProxyDesc(Unknown Source)
[info]   at java.io.ObjectInputStream.readClassDesc(Unknown Source)
[info]   at java.io.ObjectInputStream.readOrdinaryObject(Unknown Source)
[info]   at java.io.ObjectInputStream.readObject0(Unknown Source)
[info]   ...
```

when run under `sbt`.  However, it works for IntelliJ and Eclipse.  I assume these have taken over the classpath and magically solve the problem.  Our users, most importantly Travis, need to be able to make do with `sbt`.

Another possible symptom of a `build.sbt` which may not be quite right is that when I import the project into IntelliJ, it gets an extra `eidos` module beside the `core`, `core-build`, `webapp`, and `webapp-build`.  The IDE wil not let me right click and run tests until the extra module is removed.  If I manually make a configuration to run the test, I recall that it also has classpath problems until the module is removed.

It seems like build.sbt has some kind of implicit module (or lack of one) that is causing some problems.  It does seem structured differently than the same file for processors and reach.

kwalcock:
@marcovzla, you seem to have some expertise in this area.  I don't think this has high priority, but if you  happen to be looking at `build.sbt` for some reason, like for Spark integration, please keep an eye out.

marcovzla:
i'm working on it

marcovzla:
I believe this is now solved, so I will close this issue.
Feel free to reopen if needed.

kwalcock:
Thanks!

PR #289, opened 2018-04-27
Uncomment some problematic lines

kwalcock:


marcovzla:
The problem is that ObjectInputStream is not using the right ClassLoader.
I used [ClassLoaderObjectInputStream](https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/input/ClassLoaderObjectInputStream.html) to specify the correct one.

Closes #288

kwalcock:
It did indeed fix the test and on short order.  Thanks.  I wish that it didn't involve yet another dependency.

marcovzla:
That dependency was already present, because ai.lum.common uses it.
I just added it to build.sbt to make it explicit

PR #290, opened 2018-04-29
Redesign attachments

kwalcock:
Add (unused) room for mentions on an attachment to track provenance, remove need for attachment on an AttachmentSpec.


kwalcock:
I can imagine conflicts, which is why I'm trying to push it sooner.  It's only going to get worse when the provenance is added.  Sorry.  I'll double check that I didn't comment out any of the important stuff so that API is reasonably consistent.

PR #293, opened 2018-04-30
add WorldDev Indicators ontology

ZhengTang1120:


kwalcock:
I'm looking into it.  Just adding a file shouldn't cause any problems.

kwalcock:
Unless maybe zheng_test2 wasn't passing in the first place.  it looks very old.  Can you make a new branch off master and add the file to that instead?

kwalcock:
The file may need to change some, but it is used only when configured so I'll check it in if only as a placeholder for the good version.

PR #296, opened 2018-05-01
lab's favicon

marcovzla:
added lab's favicon to webapp


marcovzla:
this is useful for people that keep many browser tabs open

kwalcock:
Thanks.  Should favicon.png be removed?  It's some strange triangle.  Does @bsharpataz know where it came from?  It must not have been hooked up to the server.

marcovzla:
That was added by play when we generated the project. It may be used by the error pages displayed by play, but I'm not sure.

PR #295, opened 2018-05-01
Allow for grounding to multiple ontologies

kwalcock:

PR #302, opened 2018-05-03
Multi-ontology groundings

ajaynagesh:
- Reading multiple ontologies slightly differently
- Ontologies with examples are read as before
- Ontologies with descriptions: filter the descriptions to contain only nouns, verbs and adjectives and then read as before. 

@kwalcock: Have not yet tested loading the fao_variable_ontology.yml . Can I run a specific test for this ? Also, I assume now the framework for grounding (with your new modifications) is able to create a HashMap [onto_name, [grounding_list] ]. I have not tested explicitly. 

Will replace the gigaword embeddings with GLoVE in a subsequent commit.

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/fao_variable_ontology.yml
+++ b/src/main/resources/org/clulab/wm/eidos/fao_variable_ontology.yml
@@ -1,2706 +1,3980 @@
-- events:
-  - Agriculture orientation index:
-    - Gross Fixed Capital Formation (Agriculture, Forestry and Fishing):
-      - "Agriculture orientation index Gross Fixed Capital Formation (Agriculture,\
-        \ Forestry and Fishing)"
-    - DFA Commitment to Agriculture, Forestry and Fishing:
-      - "Agriculture orientation index DFA Commitment to Agriculture, Forestry and\
-        \ Fishing"
-  - Agriculture orientation index US$, 2005 prices:
-    - Gross Fixed Capital Formation (Agriculture, Forestry and Fishing):
-      - "Agriculture orientation index US$, 2005 prices Gross Fixed Capital Formation\
-        \ (Agriculture, Forestry and Fishing)"
-  - Amount excreted in manure (N content):
-    - Chickens, broilers:
-      - "amount excreted in manure (n content) broiler chickens"
-    - Cattle:
-      - "Amount excreted in manure (N content) Cattle"
-    - Goats:
-      - "Amount excreted in manure (N content) Goats"
-    - Poultry Birds:
-      - "Amount excreted in manure (N content) Poultry Birds"
-    - Sheep:
-      - "Amount excreted in manure (N content) Sheep"
-    - Chickens, layers:
-      - "amount excreted in manure (n content) layer chickens"
-    - Mules and Asses:
-      - "Amount excreted in manure (N content) Mules and Asses"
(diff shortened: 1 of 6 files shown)


ajaynagesh:
@kwalcock : Thanks! I also tested loading the FAO ontology. Seems to be working fine. 

I also changed the path of the embedding file to use GLoVE emebeddings (https://nlp.stanford.edu/projects/glove/) 

Specifically I am using : Common Crawl (840B tokens, 2.2M vocab, cased, 300d vectors, 2.03 GB download): [glove.840B.300d.zip](http://nlp.stanford.edu/data/glove.840B.300d.zip) 

@MihaiSurdeanu: 
I have uploaded the file (with the header line attached for processors word2Vec module) in the google drive. I noticed that the GLoVE embedding file is ~13 larger than the vectors.txt. I would definitely have a performance impact when we initially load the vectors (previously it was taking under 30s, now it has not yet completed since ~5mins when I am loading locally). Just fyi.

ajaynagesh:
Loading is still not complete after 20 mins. This is causing me some concern! ...

kwalcock:
So org.clulab.embeddings.word2vec.Word2Vec knows how to deal with this different file format?  It could be loaded in a different thread and the main thread could show some kind of "heartbeat", but unless there is a callback of some kind, the main thread wouldn't have any idea how much more time is needed.  Can you check how much memory is being used?  Getting close to the limit can slow things down quickly.

ajaynagesh:
I maxed out of my heapspace and got an exception .. but then my .jvmopts had 6G .. trying with 25G now ..

ajaynagesh:
I have uploaded the GLoVE [here](https://drive.google.com/open?id=1k4Bc3iNWId8ac_fmkr9yFKhQEycdwOVk) 

Also when I change the .sbtopts to have 25G
- Gigaword embeddings = ~30s
- GLoVE embeddings = [will update when done]

bsharpataz:
do we *need* 25G?  is it loadable with <16G?

ajaynagesh:
@bsharpataz : I tried with 10G in .jvmopts. It goes through fine. But I'm not sure if it uses up all the memory which might be needed for other stuff. I just loaded the GLoVE vectors. Takes  ~4.5 mins to load now.

ajaynagesh:
@MihaiSurdeanu : I think those are part of @kwalcock 's earlier set of changes related to multi-ontology grounding (https://github.com/clulab/eidos/pull/295)

kwalcock:
The code was checked in earlier, but I don't know what "description embeddings not the ones from examples." means.

ajaynagesh:
Yes.

kwalcock:
I'm simplifying OntologyGrounder slightly so that it is easier to understand.  Is this PR otherwise safe to merge?  Have all doubts been removed?

kwalcock:
@ajaynagesh, I made the code a bit more complicated than it needed to be.  The replacement should be equivalent, but it would be good for you to take a look.

ajaynagesh:
Looks good to me.

kwalcock:
If word2vec is false by default and the vectors file is not in GitHub, I suspect that most people will be running without grounding.  Many have only 8GB total in their machines, so 10GB might be a problem.
Tests still pass, so I'll merge.  More changes in the next PR then.

PR #304, opened 2018-05-04
bad path to wdi ontology

bsharpataz:


kwalcock:
Thanks.

bsharpataz:
np -- realized later it's also fixed in another bigger PR that isn't merged
yet -- so sorry for the extra work!

PR #305, opened 2018-05-04
Complete JSON-LD (1)

kwalcock:
The items highlighted on the JSON-LD wiki page are updated.  A more nicely formatted list is being added in the comments.


kwalcock:
In addition, positions are always lists of Intervals.  Right now we just have a list of a single interval, but it's a list just the same.

adarshp:
I don't use the plain JSON output - @bgyori: do you?

kwalcock:
As far as I know, the output can take place either before or after CAG-ification.

kwalcock:
The plain JSON does not have the groundings now, so it's missing some things.  On the other hand, the JSON-LD probably has some things that nobody (else) is interested in but which are included unnecessarily for completeness.  At the very least I don't want to drop something that somebody is currently using.  I notice that the JSON-LD serialization for the OdinMentions needs to be updated for the EidosAttachments.  Is anybody using that?  That will be in a separate PR.

PR #309, opened 2018-05-07
Cleaned up README by moving JSON outputs to wiki

adarshp:


kwalcock:
I should check to what extent the other formats really have ordering issues, but that's for a different day.

PR #314, opened 2018-05-16
added 'forc' to cause triggers

bsharpataz:
didn't wanna wait for the big guy and added something to a slide that uses this trigger, so I wanted master to be able to "generate" the output I claimed it did...


kwalcock:
Thanks, both.

PR #317, opened 2018-05-28
Test JSON-LD format using external library

kwalcock:


kwalcock:
From the email:

It would probably be a good idea for me to grab one of the JSON-LD libraries (in Java and which we aren't using at all and therefore an unfurtunate addition dependency) and write a test that just writes one of our files and reads it back, checking for exeptions.  In the past I've copied and pasted into a web browser to check, but that's not automated enough.

[Keith]

On Thu, May 24, 2018 at 11:37 PM, Ajay Nagesh <ajaynagesh@gmail.com> wrote:
Hi Maria, 

I was wondering if it would be possible to recreate this scenario in our team. 

copying this mail to Keith Alcock and Becky Sharp. We will work on this on our end and let you know soon. 

@Keith: Would it be possible for me to reproduce this at our end ? Has this error been encountered already before ? Thanks. 

Thanks!
Ajay 

On Thu, May 24, 2018 at 2:06 PM, Maria Stoica <Maria.Stoica@colorado.edu> wrote:
Hi Ajay,

 

It’s Maria here, working on ontology development for the ISI team, and I’m currently trying to figure out a way to link concepts from the UAZ CAG to our variables ontology. I’m trying to read in the json-ld file from the website using Python’s rdflib-jsonld library but I get the following error … do you know what the issue is? Should I be using a special format setting other than utf-8?

Issue #316, opened 2018-05-18
Scalability


kwalcock:
As discussed at a recent Monday morning meeting, some of our Apps are expecting input files to  have one paragraph per line.  Each paragraph is treated as a separate document and processed.  This works well with the files that were converted from PDFs by hand, like the 8 or 10 we have been using for testing.  With the full set of 52 which were automatically converted, however, this doesn't work well.  The PDF converter seems to have output one line per PDF line and not attempted to make paragraphs.  These lines are usually not even complete sentences.  The way to handle this is to treat the entire file as a document and let the Eidos system extract the sentences.  This isn't working well.  Many files never run to completion.  I made a small and informal test program that uses a text of approximately 4KB from CAG.fulltext and am timing linear increases in that text length.  It doesn't look good.  A 16KB file is taking around 30 minutes and the largest of the 52 test files is still 10 times larger.

![image](https://user-images.githubusercontent.com/8679738/40209647-5c30905e-59f5-11e8-826f-280ee6225874.png)

MihaiSurdeanu:
Thanks @kwalcock!

We've seen this behavior before: the text produced from the PDF reader has huge sentences because of lack of punctuation. Some of the components we have depend on sentence length, so they are completely off for very long sentences.

One simple solution I suggest: insert an artificial period whenever the text contains *two* new lines wo/ any text in between. This usually indicates a new paragraph, and we should treat it as such, even when no punctuation is recognized. 
Also, I'd like to see an example of these files that take forever to parse.

kwalcock:
We could have the problem you mention, but this test is being performed with known, fairly well behaved text.  It's pasted below.  To make the document longer, I just concatenate another copy.  We are probably doing more searching and sorting than previously and perhaps I should comment it out to see if that's the culprit.

Food insecurity levels are extremely alarming throughout the country due to conflict, a collapsing economy, low cereal production, poor rainfall in southeastern areas, and the exhaustion of coping capacities after several years of crisis. In Eastern Equatoria - one of the drought-hit areas - the latest food security and nutrition assessments in November 2016 revealed that households experiencing poor food consumption nearly doubled from 19 to 37 percent when compared with the same time in 2015. Rainfall deficits were widely reported in the area, with 93 percent of households indicating this was a major shock during the last agricultural season, impacting pasture and water availability and local food production. Climatic forecasting indicates these areas are likely to experience depressed rainfall between March to May 2017.
[...]

PR #319, opened 2018-06-01
Do not process files line by line but by their entire content, and do so efficiently and with output properly encoded

kwalcock:
Some pdf to text converters were producing one line per line in the PDF.  These partial sentences, usually, were becoming single documents when processed line by line by various Apps.  That's not what we want.  Unfortunately, the larger texts were a performance problem and had to be addressed mostly in an update to processors.  Some small efficienty improvements to eidos are included.

Also, programs that were checking text files for comments, removing them, and joining the lines cause problems with the provenance.  Comments are therefore no longer allowed in the text files.  The Google Drive copies of some of the documents were updated to remove comments.

Some output was also not being produced in UTF-8 format.  This can cause problems for consumers of the JSON-LD output in particular.  It is addressed here.


kwalcock:
See also Issue #316

kwalcock:
People are tagged just to keep them up to date.  Maybe you're notified already.

PR #311, opened 2018-05-09
Update of DomainOntology

kwalcock:
This supports an OntologyNode (among other things)

PR #321, opened 2018-06-05
Change ontology file format

kwalcock:


kwalcock:
This is extending PR #311.  It doesn't really belong here, but this is where we know to find it.

kwalcock:
Regarding the IDs...  I'm not sure where they came from or what they are intended to do.  Are they supposed to be unique within an ontology, between our ontologies, between ontologies built by anyone?  At some point we started adding a top level to the path (-UN:, -FAO:, -WDI:).  This in combination with the remaining unique path gives us uniqueness across at least our ontologies.  This ID is so far never output anywhere.  Is the ID supposed to just be shorter than the entire path or a simple way to search in the file for an entry (difficult in a tree)?  Can we just call them UN1, UN2, UN3, etc.?  If we edit an ontology, our IDs probably won't be consistent across different versions of it.  An insertion or deletion would mess things up.  UN200 would be "foo" in version 1 of the ontology and may be "bar" in version 2.  Maybe that doesn't matter.  I wonder whether the ID will be used whatsoever.  Any advice?

kwalcock:
This is being turned into an issue where it better belongs.

PR #324, opened 2018-06-07
Add some tools to get reachable mentions and check uniqueness

kwalcock:
This is a PR on a dev branch not intended for master yet.


bsharpataz:
looks good to me -- to be clear, which issues on dev_hashiness does this solve/leave unsolved?  I think it digs in and gets mentions from within events that are not in the State.  However, it creates a duplication issue?  If so, can that be solved with a hash-based equivalence/filtering?

PR #329, opened 2018-06-10
huge PR from the MITRE submission

bsharpataz:
@kwalcock -- this is so overdue I am not planning on splitting it (BUT I can if you tell me you want me to!!)

@kwalcock there are currently 2 failing tests -- both in the TestUnicode and I don't remember any more how that was supposed to be fixed.  I dug into it a bit and indeed we're just slurping up the unicode.... if you remind me what actually worked before (I think my approach didn't, right?) I can work on it or you can if you want (that would be great!)


kwalcock:
FWIW I'm looking at this now.

kwalcock:
It's definitely failing on Travis, but it says it passes on my computer.  However, if I run it in the debugger, it seems to end the test early.  Something strange is going on.

kwalcock:
The test sentence in that file was no longer producing the events that were being tested or any at all for that matter.  That might not be a good thing itself.  I made up a different sentence, but the unicode characters are being incorporated into the entities that are found.  That's what we're trying to avoid.  The    "unicode-icons" rule is still there, but it doesn't seem to be working.

MihaiSurdeanu:
Do you expect us to review all this? :)
Did you regenerate the iteration 2 results? Did they look the same? If so, I propose to merge.

bsharpataz:
Yep, about the same! Woohoo!

And yes, please review all 51 files, line by line is preferable

kwalcock:
I'm crossing my fingers and will endeavor to check out the details a little at a time.

kwalcock:
P.S.  I'll work on that unicode issue, of course.  I agree that rules are not the best place for it to happen.

PR #330, opened 2018-06-12
Fix ontologies to pass tests

kwalcock:
@ZhengTang1120, can you study the changes in case you have to change a script somewhere?  Pretty soon we'll have a place on GitHub for you to store it/them.  Here are some notes:

fao: Descriptionless node is removed.

toy: Indentation corrected, December added.  Some files are definitely getting cut off.  Examples pluralized.

un: December added.

wdi: Unicode characters removed.  Consequently internal quotation marks escaped.

It was decided that descriptions should be plural, but that will be changed later.

These changes are already in my branch which should be headed to master shortly.


kwalcock:
I found this page helpful: http://blogs.perl.org/users/tinita/2018/03/strings-in-yaml---to-quote-or-not-to-quote.html

MihaiSurdeanu:
What's the TL;DR on that document? When should we use quotes?

kwalcock:
My takeaway from the document is that a library should be used to escape things propertly.  I say that reluctantly as a do it yourself kind of person who doesn't like dependencies.  I only read it close enough to know what to do with quotes within quotes, though.

PR #332, opened 2018-06-12
New ontology format and code to read them

kwalcock:


kwalcock:
I double checked that the merge from master with @bsharpataz seemed to work correctly.  I could be mistaken.  I guess that's what version control is for, though.

PR #337, opened 2018-06-15
Release todo

bsharpataz:
do we also need to change version.sbt to not have "SNAPSHOT" in it?

please let me know what else should be done prior to release, but ideally we can release this tonight?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,6 @@
 # Changes
++ **0.2.1** - Switch to top-down extraction (entities now found only when they participate in events), grounding
+to multiple grammars (high- and mid-level), scalability improvements, updates to JSON-LD export format
 + **0.2.0** - Grammar improvements, grounding to ontology, canonical names for extractions, 
 entity filtering, added example output of system on 10 debugging documents
 + **0.1.5** - Added missing organization id to build file
--- a/example_output/example-0.2.1.jsonld
+++ b/example_output/example-0.2.1.jsonld
@@ -1,0 +1,2248 @@
+{
+  "@context" : {
+    "Corpus" : "https://github.com/clulab/eidos/wiki/JSON-LD#Corpus",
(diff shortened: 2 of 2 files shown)


kwalcock:
I guess it's technically still tonight.  Is it master you are going to release?  I've been holding off on e.g. processor changes and temporal context until a release of some kind.  I *think* the release process asks about a version number, changes the current one in github, releases, and then updates the version number with SNAPSHOT.  However, if one updates the version by hand, that seemed to work as well.

bsharpataz:
So... what are my action items here? Do i need to mege these as is? Should
i go to v 0.3.0 Or is 0.2.1 ok? Do i need to make version.sbt match?  Who
is releasing?

kwalcock:
Did you do it before?  I got all signed up with accounts, passwords, keys, and everything else and can do it if necessary.  Maybe I even remember which commands to use.

bsharpataz:
I've never released before and I'm not set up currently. Can either you or @MihaiSurdeanu once we are sure that the questions above are fine?

kwalcock:
I can sure try.

MihaiSurdeanu:
Keith, can you please release?

kwalcock:
I vote for 0.3.0.  If someone else votes in the next five minutes, I'll count that as well.

MihaiSurdeanu:
Sorry, I only saw this now.
I vote for 0.3.0. I think this is a big change.
Do you want me to adjust CHANGES and release? Or have you done this already Keith?

kwalcock:
It is already done at 0.2.1.  Somewhere above it says merged.  That doesn't prevent us from having 0.3.0 as well.  It's only one command and there haven't been commits in the meantime (we were waiting just in case).  Just let me know.

MihaiSurdeanu:
If it’s done, it’s OK. We’ll bump the number later.

kwalcock:
There should be some good opportunities with the change to CluProcessor or the update to the JSON-LD format.

PR #338, opened 2018-06-16
adding properties from icasa spreadsheet

ajaynagesh:
I am no domain expert. But I tried to adding some of the variables from the ICASA spreadsheet into the properties node of the un_ontology. The examples are made up by me based on what looked like commonsense. Feel free to ignore these variables if not necessary

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/ontologies/un_ontology.yml
+++ b/src/main/resources/org/clulab/wm/eidos/ontologies/un_ontology.yml
@@ -47,6 +47,33 @@
       examples:
         - liter
         - gallon
+    - OntologyNode:
+      name: solar_radiation
+      examples:
+        - solar radiation
+        - sunlight
+        - solar energy
+    - OntologyNode:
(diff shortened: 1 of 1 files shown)


MihaiSurdeanu:
I am not sure solar_radiation is a property? Hard to combine it with anything...
@ajaynagesh, @bsharpataz: what do you think?

ajaynagesh:
@MihaiSurdeanu Had a similar feeling too. Since you mentioned in the mail, I thought it should be included. I could remove just this one. Others seem ok, except for depth (maybe) ..

adarshp:
Solar radiation seems like an object rather than a property to me  - I think of it in the same way as rainfall, etc...

MihaiSurdeanu:
I agree with Adarsh: it should be included in the ontology, just not as a property.
Maybe under weather, for the lack of a better non-terminal?

ajaynagesh:
Moved the solar_radiation to weather...

PR #339, opened 2018-06-17
Update for Eidos & Windows

kwalcock:


kwalcock:
@EgoLaparra, you were very right in that I had overlooked the `@transient` annotation and now just to be sure, I merged your changes into mine.  Now it's time to merge mine back into yours, if you approve.
`@transient` will prevent the non-serializable fields from causing immediate problems, but of course the result is that the objects aren't serialized.  We don't have any tests right now that check for this, so we're really cheating.  We should find a way to get them serialized.
The model file is small enough to fit into GitHub, but using it makes the tests run too slowly for everyday use.  A boolean variable may need to be added, similarly to how the grounding works, to turn this on and off.
The JSON-LD representation will probably be changing as well after multiple other people look at it.
However, the tests are passing and I should get these changes back into the branch you are working on.

kwalcock:
Additional people will have to study this before it goes into main.  They'll have another chance.

MihaiSurdeanu:
I kinda think this docCreationTime should be added to the processors base Document class. Even if it's not really used in processors (for now). This would simplify the architecture in eidos a lot. There would be no need for the EidosDocument class. Should I do it, @kwalcock, @EgoLaparra?

kwalcock:
I'm not quite sure what @MihaiSurdeanu means above.  Where will the times (`@transient val times = new Array[List[TimeInterval]](sentences.length)`) all go then, or is this exactly what Egoitz is working on and it is assumed there's a place for them among the sentences or mentions or somewhere else in processors?  It seems like EidosDocument can't disappear completely until that happens.

MihaiSurdeanu:
What I had in mind was that times gets stored in Sentence.norms. Note that this field already stores such values, when they are produced by CoreNLP. 
However, Sentence.norms is an Array[String]. Let's make sure the code that populates EidosDocument is solid first. Then we can think about how to transfer it.

PR #341, opened 2018-06-17
Use mkDocument instead of entire annotate

kwalcock:


kwalcock:
This brought the time for one test down from 8 minutes to 3 minutes and it used noticeably less memory.  That's at least the right direction.  This is for after the impending release, just in case.

bsharpataz:
Works at least for now, and if we want more annotation than just POS we can add.  Is the 3 mins including time to load the vectors?  If so, we can def work on that

bsharpataz:
Specifcally, @marcovzla suggested a way to... index? the vectors

kwalcock:
Thank you!

kwalcock:
@MihaiSurdeanu, is this any better?  It may eventually need to be handled by processors.

PR #344, opened 2018-06-18
Ontology becky

bsharpataz:
minor change -- added biotic_resources non-terminal node to match abiotic branch and fixed minor indentation issue where abiotic_resources wasn't indented --

but please review!
do any tests check for indentation issues?


kwalcock:
There is a TestDomainOntology that will check the syntax of the ontologies even if they aren't configured to be used in the program.  As long as it's valid syntax, though, it won't mind semantically incorrect files which might be the result of some indentation errors.
I can't usually say much more about the content of the ontology files, though.  I'll mark approve but wait until someone else does as well or until you merge it.

kwalcock:
OK, I'll wonder about capitalization as well, especially because of that nAtuRAl ReSouRceS from earlier.  The strings in the ontologies go through wordToVec.makeCompositeVector -> w2v.makeCompositeVector -> matrix.get(s) where matrix is Map[String, Array[Double]].  I don't see anything obvious that standardizes the case.  It seems like this should have case insensitive lookup.  Maybe I should write a test case for it just to be sure.  Someone let me know if I've obviously misunderstood.

bsharpataz:
I don't think he meant for the examples/descriptions, and whether we want
case insensitive lookup is a separate question (i.e. the diff between the
semantics of Smith and smith and China vs china... it *can* help with
disambiguation -- not that we necessarily need that).  But he was asking me
to make the case of the OntologyNode names consistent,  which I can do soon
(but maybe not this afternoon)

kwalcock:
There's a PR out to fix the capitalization on the non-terminals.  They've been made lowercase.  The change will show up here shortly in order to merge.

bsharpataz:
did https://github.com/clulab/eidos/pull/348 fix this?

kwalcock:
Probably, but I'm not sure what this is.  There were only about 3 changes.  Can you click on 7a83a1a?

bsharpataz:
@MihaiSurdeanu I reviewed the capitalization of file after @kwalcock fix (thanks Keith!!) and seems good (granted, a quick review).  I removed the bad GPE comment (thanks @ajaynagesh !)

I think it's ready -- if you approve it we can merge it, then merge the other ontology stuff so we don't get too log-jammed

Issue #345, opened 2018-06-20
Eidos throws exceptions on files in our larger collection


kwalcock:
Documents 10, 11, 17, 25,  26, 28, 30, 31, 42 cause java.util.NoSuchElementException or java.lang.RuntimeException: what? that need to be looked into.

```
Could not process 26_South_Sudan_2018_Humanitarian_Needs_Overview.txt
java.util.NoSuchElementException: key not found: cause
        at scala.collection.immutable.Map$EmptyMap$.apply(Map.scala:97)
        at scala.collection.immutable.Map$EmptyMap$.apply(Map.scala:95)
        at org.clulab.wm.eidos.EidosActions.$anonfun$createEventChain$7(EidosActions.scala:72)
        at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
        at scala.collection.Iterator.foreach(Iterator.scala:929)
        at scala.collection.Iterator.foreach$(Iterator.scala:929)
        at scala.collection.AbstractIterator.foreach(Iterator.scala:1417)
        at scala.collection.IterableLike.foreach(IterableLike.scala:71)
        at scala.collection.IterableLike.foreach$(IterableLike.scala:70)
        at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
        at scala.collection.TraversableLike.map(TraversableLike.scala:234)
        at scala.collection.TraversableLike.map$(TraversableLike.scala:227)
        at scala.collection.AbstractTraversable.map(Traversable.scala:104)
        at org.clulab.wm.eidos.EidosActions.$anonfun$createEventChain$5(EidosActions.scala:69)
        at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:241)
        at scala.collection.immutable.List.foreach(List.scala:389)
        at scala.collection.TraversableLike.flatMap(TraversableLike.scala:241)
        at scala.collection.TraversableLike.flatMap$(TraversableLike.scala:238)
        at scala.collection.immutable.List.flatMap(List.scala:352)
        at org.clulab.wm.eidos.EidosActions.createEventChain(EidosActions.scala:67)
```

```
Could not process 25_2016_HNO_South Sudan.txt
java.lang.RuntimeException: what?
        at scala.sys.package$.error(package.scala:27)
        at org.clulab.wm.eidos.EidosActions.expandIfNotAvoid(EidosActions.scala:534)
[...]

PR #347, opened 2018-06-20
Bugfixes

marcovzla:
This aims to fix the issues reported in #345 

There were two main issues:
- we were assuming all mentions have a syntactic path between a trigger and an event, but this is not the case for mentions found by a surface rule
- there were overlaps between some triggers and their arguments. i added a few "avoid" rules, but i'm not sure if they affect other extractions

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/grammars/avoidLocal.yml
+++ b/src/main/resources/org/clulab/wm/eidos/grammars/avoidLocal.yml
@@ -102,3 +102,35 @@ rules:
     pattern: |
       # avoid passive like in "marked by drastic hyperinflation"
       [tag=VBN] "by"
+
+  - name: avoid_given
+    label: Avoid
+    priority: 3
+    type: token
+    pattern: |
+        # `given` is a correlation trigger defined at `syntax_explicit_Correlation_given`
+        given
+
+  - name: avoid_syntax_2_verb-Correlation_triggers
+    label: Avoid
+    priority: 3
+    type: token
+    pattern: |
+        # In Greater Upper Nile and Greater Bahr el Ghazal , limited activity is occurring on most trade routes due to combination of insecurity and seasonal constraints , as the rainy season limits road functioning .
+        # functioning is a correlation trigger that should be avoided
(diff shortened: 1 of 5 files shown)


marcovzla:
@bsharpataz i hope you find this useful

bsharpataz:
@marcovzla yes! thanks!! and sorry I was only sometimes available today!
about the avoiding -- can you please walk me through why those were necessary?
Ideally we won't need to have specific triggers being avoided, more like whatever the _current_ trigger is should be avoided, but others are fine -- where does this break down?
(and thanks!!)

marcovzla:
they were necessary because they were being matched by entity and trigger rules. Maybe there was a better way of filtering them but I couldn't figure it out.

For example, in the sentence `However , the mechanized sector in Upper Nile State is expected to produce some 7 000 tonnes of sesame from 23 000 feddans -LRB- 9 600 ha -RRB- that is likely to be purchased immediately from the field by Sudanese traders .` we are getting the originl entity (before expansion) `expected to produce` by the `simple-vp` rule and then the trigger `produce` with the rule `ported_syntax_5_verb-Causal`. The code doesn't expect the trigger and arg to overlap and throws an exception. I added an `Avoid` rule for that trigger, but maybe there is a better way to handle this?

marcovzla:
this is where it breaks: https://github.com/clulab/eidos/blob/25f0b7bc5c69210f3ac7a8eac1f24e5714101128/src/main/scala/org/clulab/wm/eidos/EidosActions.scala#L534-L542

bsharpataz:
Thanks, that’s helpful! Ummm, ok, let me think it through...

kwalcock:
Tests are in the pipeline.  Thanks, @marcovzla, for working so hard on figuring it out.

marcovzla:
i think moving away from chunking makes sense

marcovzla:
where should i add unit tests for this issue?

kwalcock:
Maybe test/org.clulab.wm.eidos.system where these is already a TestEidosActions.  Moving it isn't difficult.

PR #349, opened 2018-06-21
Update sbt and plugins

kwalcock:


MihaiSurdeanu:
@marcovzla, @kwalcock: did one of you test these versions? Especially the release one?
If so, this is Ok.

marcovzla:
I haven't released anything, but based on the docs I think this should work.

kwalcock:
Is there an easy way to test that other than by releasing?  The update to processors 7.4.0 is being tested and switching to that will involve a release shortly.

marcovzla:
not that i know of

kwalcock:
This is waiting until after some important demos.

PR #350, opened 2018-06-21
maria's comments on entities vs. events.. preceeded with Maria:

ajaynagesh:
Placeholder for discussion of a broader question of when some ontology nodes should be treated as entities or events (processes etc.). The definition is not clear yet either to me or Maria Stroica ...

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/ontologies/un_ontology.yml
+++ b/src/main/resources/org/clulab/wm/eidos/ontologies/un_ontology.yml
@@ -65,8 +65,8 @@
       name: temperature
       examples:
         - temperature
-        - degree celcius
-        - farenheit
+        - degree celsius
+        - fahrenheit
         - kelvin
         - C
         - F
@@ -106,8 +106,11 @@
             - contaminant
             - contamination
             - soil pollution
-            - deforestation
             - soil erosion
+        - OntologyNode:
+          name: climate_change
+          examples:
+            - deforestation
             - desertification
     - OntologyNode:
       name: crisis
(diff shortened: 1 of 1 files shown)


ajaynagesh:
Further discussions ...
 
Top-level nodes according to Maria
--------------
- Property (characteristics of an entity/event … which cannot be very easily quantified .. e.g.: aridity, ability, availability)
- Event (entities undergoing processes, e.g.: deforestation, snowfall, rainfall)
- Entity ( anything that has an attribute or a property .. and typically static)
- Units
- Attribute (characteristics of a system property value pair, e.g.: arable_land=%_of_land)
- Operator (ratio / percentage … )

Phenomenon (e.g: gravity) vs Event (hurricane Katrina ...)

kwalcock:
It should be easy enough to run a spell checker over our ontologies.  Let's make that part of the process.
These are just in the comments, but other mistakes haven't been.
occuring -> occurring
govenrment -> government

ajaynagesh:
@kwalcock sure ... 

Here is a subset of variables which Maria gave me from the GSN ontology that are relevant and could be added to augment our un_ontology: 

https://docs.google.com/spreadsheets/d/1gGPJiyXtE-70Af0Kh7bu6svc7GbIrP-qdoKWuaOnQ_k/edit#gid=1770194572

kwalcock:
Both spellings of anemia/anaemia were present, so there was really no need to choose.

ajaynagesh:
I noticed that this PR was merged. Just wanted to point out, some of the comments (for instance reorganization of some of nodes as entities or events) need more thought and discussion, before incorporating those comments.

bsharpataz:
if this is a good place to mention this -- the reorganization or top-level
branches (i.e., units, etc) should be thought through in terms of how we
will ground to things/represent groundings before we implement in the
ontology

PR #352, opened 2018-06-22
Pull proc740 after testing on 52 documents

kwalcock:


kwalcock:
This is waiting until after important demos.

kwalcock:
Things have been pretty quiet recently, so I wonder whether now is a good time to merge this into master.  @MihaiSurdeanu, @bsharpataz, what do you think?

bsharpataz:
i think so!  this will allow the temporal stuff to go in for real too

kwalcock:
This will be replaced shortly by kwalcock-proc741 and not itself be merged.

PR #356, opened 2018-06-27
fixing the maria stoica comments ..

ajaynagesh:
Removed all # maria .. comments and reorganized the ontology. @MihaiSurdeanu: FYI

Also ran the system end-to-end on the 10 mitre documents and getting similar groundings...


bsharpataz:
merging, but won't be used for the 6mo eval bc of the multitude of downstream effects on stuff that other performers have already done

PR #357, opened 2018-06-29
weighted version of the mention grounding 

ajaynagesh:
using (i) synHead-distance and  (ii) IDF

PR this change although it results in a slightly poorer performance on the UN grounding (see experiments [here](https://docs.google.com/spreadsheets/d/1STJ4C-GBFqp_VE8xXqhNriX5QXu0-hhytWTUtfT7H7Y/edit#gid=842605389)) 

the weight computation is done as follows: 

For example, in the extracted mention: 

"price of oil in Sudan": 

synHead weights:
----
wt_syn(price) = 1
wt_syn(oil) = 1/2
wt_syn(Sudan) = 1/3

IDF weights:
---
wt_idf(price) = 1 / log (freq(price))
wt_idf(oil) = 1 / log (freq(oil))
wt_idf(Sudan) = 1 / log (freq(Sudan))

To find freq(X) : 

consider the corpus of tokens of ontology terminals (tokenized by space or _) : description + examples

Also we might have to smooth the freq(X) appropriately. 

Then, wt(term) =  wt_syn(term) + wt_idf(term) [todo: should we use multiplication here?]

---- 

Now cosine similarity computation has to be modified appropriately: 

Say we are grounding to /entity/resource/oil: 

which contains: 

- oil
- natural gas
- crude oil
- petroleum 

Then : 

cos("natural gas", "price of oil in Sudan") = 

Sigma [  Vec(natural) * Vec(price) * wt(price) ,  Vec(natural) * Vec(oil) * wt(oil) ,  Vec(natural) * Vec(Sudan) * wt(Sudan) ,  Vec(gas) * Vec(price) * wt(price) , .... ]

Basically all pairs.

I have used `word2vec.interpolate()` to do this instead of `makeCompositevector()`. please see function `makeCompositeVectorWithInterpolate()` in `EidosWord2Vec.scala` 

Adding @kwalcock as a reviewer as well, for I changed some of the EidosWord2vec functions...


kwalcock:
There are apparently offline discussions about this PR.  It would help me to know the outcome so that I can decide what to do.

ajaynagesh:
Hi @kwalcock : I am working on an extensive error analysis .. experimenting with a bunch of issues. Making notes of this offline. Will update here once some clarity is reached ...

kwalcock:
@bsharpataz, should these be closed without merging?

bsharpataz:
@kwalcock, check with @ajaynagesh  and @MihaiSurdeanu , I think that yes this should be closed without merging bc it didn't work, right? or did it @ajaynagesh ?

ajaynagesh:
Yes ... it did give us better results .. compared to what we already have ..

bsharpataz:
closing, we can always revisit

PR #358, opened 2018-07-01
Serialize mentions

bsharpataz:
OK -- so we were having app-plosion... I added this app to facilitate (a) reading from a directory and subsequently (b) exporting into one or more formats.  This removes the need for the MakeMITRE app, so I removed it (and moved the methods over).
Current formats:
- JSONLD (for obvious reasons)
- MITRE table for eval
- serialized *odin* mentions (for importing into influence search)

I added some content to the eidos.conf to make the magic happen, but it's not too cumbersome, IMHO

@ the three of you -- Please do review-review if you're willing and let me know what changes I should make, surely I did something in a less than scala-tastic way but I can make changes (or you can)

ps -- fixed a bug in the MITRE table export where I was printing the FAO grounding for the effect in place of the cause... !!

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -8,8 +8,8 @@ EidosSystem {
 //     avoidRulesPath = 
 //       taxonomyPath =
 //            maxHops =
-      wordToVecPath = "/org/clulab/wm/eidos/w2v/vectors.txt"
-//      wordToVecPath = "/org/clulab/wm/eidos/w2v/glove.840B.300d.vectors.txt"
+//      wordToVecPath = "/org/clulab/wm/eidos/w2v/vectors.txt"
+      wordToVecPath = "/org/clulab/wm/eidos/w2v/glove.840B.300d.vectors.txt"
 // topKNodeGroundings =
 //      stopWordsPath =
 //    transparentPath =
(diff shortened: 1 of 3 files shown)


marcovzla:
Thanks @bsharpataz!
I will check this out and get back to you.

kwalcock:
This is a great idea!  My only minor reservation is that the config file might limit its applicability to sbt environments.  If the arguments were taken from the command line, they could be more easily manipulated.  However, I don't think this is the place to address the issue.  Somewhere else we can check the command line for parameters and then fall back to defaults in the config file, etc., and then finally call  a method with arguments, an exposed API.

bsharpataz:
sure! I will make the changes here soon, thanks for finding this!

marcovzla:
One way to override the config from the command line is using java system properties, e.g., `-Dapps.inputDirectory=/path/to/input`

see https://github.com/lightbend/config#standard-behavior

kwalcock:
Thanks, @marcovzla.  I can probably use that when running from a GUI instead of having to edit the config file and then change it back.

bsharpataz:
@kwalcock I think this addressed your requested changes

kwalcock:
Thanks, @bsharpataz.

bsharpataz:
:)

PR #360, opened 2018-07-03
Fix the correlations

kwalcock:


kwalcock:
Before, a correlation was recorded as a DirectedRelation with "Correlation" in the labels.
```
    "@type" : "DirectedRelation",
    "@id" : "_:DirectedRelation_3",
    "labels" : [ "Correlation", "UndirectedRelation", "EntityLinker", "Event" ],
    "text" : "Trade worsened during the conflict as there was no exchange of goods from Sudan\nand the other states of South Sudan",
    "rule" : "syntax_2_verb-Correlation",
    "canonicalName" : "Trade be exchange goods",
```
After, it matches the documenation and is recorded as an UndirectedRelation with the same labels.
```
    "@type" : "UndirectedRelation",
    "@id" : "_:UndirectedRelation_1",
    "labels" : [ "Correlation", "UndirectedRelation", "EntityLinker", "Event" ],
    "text" : "Trade worsened during the conflict as there was no exchange of goods from Sudan\nand the other states of South Sudan",
    "rule" : "syntax_2_verb-Correlation",
    "canonicalName" : "Trade be exchange goods",
```

kwalcock:
A trigger was also added to the UndirectedRelation.  It wasn't in the documentation before.

bsharpataz:
thanks @kwalcock !

kwalcock:
It seems that the modellers were not explicitely using correlations, but they may have done so unintentially if the labels weren't checked.  So, this change could result in fewer relations being modelled.  Also, the new UndirectedRelation objects in the JSON might confuse the import.  Whether we want to ever use this patch is questionable.  An update to the new JSON format is already finished.  This code could be merged just before that code so that at least github has a record of it.

bgyori:
We actually checked for the "Causal" label already but of course this is a welcome fix!

kwalcock:
If anyone is JLD-serializing plain Odin Mentions, then this patch is necessary to get the same UndirectedRelation as seen for EidosMentions.  Both of these will probably disappear with the transition to the new JLD format.

PR #363, opened 2018-07-05
host scaladocs

marcovzla:
added sbt command to publish scaladocs.

Type `sbt ghpagesPushSite` to push scaladocs.
They are hosted here:
https://clulab.github.io/eidos/latest/api/org/clulab/wm/eidos/index.html

Some other things would be nice, like a redirect from https://clulab.github.io/eidos so that we don't need to remember the whole url. However, this already works.


marcovzla:
Note that the `gh-pages` branch must exist for this to work.

kwalcock:
You are a genius.

Does the gh-pages branch need an index.html like the one below?  Will it get overwritten or cause problems?
```
<html>
	<head>
		<meta http-equiv="Refresh" content="2; url=https://clulab.github.io/eidos/latest/api/org/clulab/wm/eidos/index.html">
		<title>Eidos Reference Documentation Redirector</title>
	</head>
	<body>
		You should momentarily be redirected to the Eidos reference documentation at
		<a href="https://clulab.github.io/eidos/latest/api/org/clulab/wm/eidos/index.html">https://clulab.github.io/eidos/latest/api/org/clulab/wm/eidos/index.html</a>.
	</body>
</html>
```

kwalcock:
That looks good to me, except that I wonder why the redirect is done in JavaScript.  Is that the new way of doing things?  If for some reason we need to have other pages, the redirect could become a web page with multiple links.  Would you like to add a redirecting src/site/index.html to this branch before it gets merged?

marcovzla:
Sure I can add it. And we don't need to use the javascript redirect. I'll try your version first.

kwalcock:
I should have had a document type, encoding, language, etc.

marcovzla:
Added the redirect. You can now visit https://clulab.github.io/eidos and be redirected to the scaladocs.

marcovzla:
@bsharpataz if you want to see the scaladoc without publishing it (e.g. in some branch other than master) you can use `sbt previewSite`

bsharpataz:
@marcovzla and @kwalcock when do we get to merge this bad boy?

marcovzla:
I think this is complete. Unless you want me to add something else?

PR #366, opened 2018-07-14
Grounding webapp

bsharpataz:
ok, so this PR is a little of a mixed bag, but primarily it is focused on the webapp:

- adding grounding info
- getting nested args so they would show always
- changing the webapp loading graphic
- and... while I was at it I added a constraint to several rules....

@marcovzla you don't need to review-review... this is fyi, but if you want to make sure it's all kosher, your input is always appreciated....


kwalcock:
@bsharpataz, although the rule changes don't break any tests, they seem to change the output for the 52 documents quite a bit.  Is that expected behavior?

bsharpataz:
Um... idk.  Any particular pattern you notced?

PR #367, opened 2018-07-14
add polarity in ontologies

ZhengTang1120:


bsharpataz:
@kwalcock this is per my request, currently the change doesn't *do* anything, but eventually it will be used in the ("offline") indicator mapper, even if no where else... Currently, the polarity doesn't need to go to the jsonld -- i don't know if that will change... but my current guess is no... except that everything changes? :)

kwalcock:
@bsharpataz When will it be "safe" to merge all these pull requests and not risk adversely affecting a demo?  It seems like there are a couple of demos to go yet.

bsharpataz:
IDK for sure -- the evaluation demos are over, but the system will be demo'd at the PI meeting too -- @MihaiSurdeanu when should we merge new content? (i think this includes processors 7.4, this polarity thing, um.... others @kwalcock ?)

We should DEF merge the PR I submitted (#366) as soon as it's checked bc that's the branch I used for demo 1 anyway...

kwalcock:
Recent versions of IntelliJ seem to be more agressively reformating code in ways that don't match our traditional format.  We might have to change some settings so that fewer irrelevant changes are registered.

PR #376, opened 2018-07-26
Add example output for new version

bgyori:
This PR adds an example JSON-LD for the version 0.2.2 of Eidos. The structure has changed since 0.2.1 so this is useful to have.


kwalcock:
Thanks, @bgyori.

PR #380, opened 2018-07-27
Checkbox to toggle debug/demo modes

### LINKED ISSUE #375: add filter for just VBN entities

bsharpataz:
Previously we were showing all concepts in the webapp bc it's primarily a debug tool.  But, when we demo, it begs the question "Why are you getting that???" 
So -- now this is optional.

I also added a constraint to require that a concept must contain something besides just a VBN (helps prevent "adjectivey" only concepts)

Minor change:
 - I rearranged the order of some methods in HomeController, and to make things semantically cleaner I swapped the polarity of the boolean for filtering CAG relevant in the EidosSystem extractFromText method

closes #375


marcovzla:
well, once the tests pass

kwalcock:
It seems to work even on my computer.

PR #382, opened 2018-07-27
Add some dots to config, double check for UN grounding

kwalcock:


bsharpataz:
Ha! i literally just made the same changes today, chez moi.  I guess I don't need to PR it after all?

kwalcock:
Then it's probably worth taking a second look at.  Your version likely checked to see whether the sequence of groundings was empty before taking the head.  It should not be necessary to strip /examples or /description: they are not used in the path.  I regret using simple types to express what it happening, resulting in grounding(namespace).grounding.  If it gets more complicated, I'll make proper classes for it.

PR #390, opened 2018-07-30
Cache onts

bsharpataz:
To save time.  This is a semi-temporary solution perhaps.  This code has the following affects:
- Now EidosSystem has a param to load cached onts.  if you use them, you need to give an absolute path
- the config file reflects this
- if you want to have cached ontologies, you need to first run org.clulab.wm.eidos.apps.CacheOntologies


kwalcock:
I'll be studying this closely and testing it for Windows tomorrow.

kwalcock:
More comments are forthcoming, but keep #406 in mind.

kwalcock:
Great idea and well done, it is (or at least it was before I mucked around with it).  On one informal test it saved 25 seconds out of around 90.  Is that about what you see?

I made some slight changes that you may wish to veto.  Somebody hasn't in the past liked println in the more library-like code, so they were changed to log statements.  The Serializer in Processors should probably be updated (and it looks like Marco already did), not just because of the class loader but also because it can potentially leak a resource.  I hope that Scala soon has an official way to "try with resources".  In order to avoid the unofficial ways, it does use a no-no null.

It turns out that the configuration values can be fished out of an EidosSystem, so I rearranged CacheOntologies quite a bit, probably unnecessarily.

PR #412, opened 2018-08-16
Fixed expansion edge case failure

### LINKED ISSUE #364: Cause and effect start on same word
In some cases were seeing cause and effect overlap, and in fact starting on the same word.  Six of the 52 test files exhibit the problem.

So far this has been noticed to happen with three rules and these counts:
7 - syntax_explicit_Correlation_given
1 - ported_syntax_5_verb-Causal
1 - dueToSyntax2-Causal

Diffs of the zipped files attached will lead to the sentences, rules, etc. in question.  One example is

"In Leer, available evidence from the recent FSNMS data conducted during the lean season July/August suggests some households were in Catastrophe (IPC Phase 5) at the peak of the lean season, and given the likelihood that not all households have access to a harvest, it is expected that some households could still be in Catastrophe (IPC Phase 5)."

which results in

List(Correlation, UndirectedRelation, EntityLinker, Event) => given the likelihood that not all households have access to a harvest, it is expected
    Rule => syntax_explicit_Correlation_given
    Type => EventMention
    trigger => given
    cause (Concept, Entity) => likelihood
    effect (Concept, Entity) => likelihood that not all households have access to a harvest, it is expected

[OverlappedCauseAndEffect.zip](https://github.com/clulab/eidos/files/2168361/OverlappedCauseAndEffect.zip)

bsharpataz:
When argument overlaps with the trigger, backoff from expansion to the
original.  Gives a debug statement instead of a error.

@MihaiSurdeanu fyi

if travis is happy, i think closes #364 
??

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -23,12 +23,11 @@ EidosSystem {
 }
 
 apps {
-  inputDirectory = ""
-  outputDirectory = ""
-  inputFileExtension = "txt"
+  inputDirectory = "."
+  outputDirectory = "."
+  inputFileExtension = ".txt"
   exportAs = ["serialized", "jsonld", "mitre"] // valid modes: jsonld, mitre, serialized
   groundTopN = 5
-
 }
 
 
--- a/src/main/resources/org/clulab/wm/eidos/grammars/avoidLocal.yml
+++ b/src/main/resources/org/clulab/wm/eidos/grammars/avoidLocal.yml
@@ -56,8 +56,6 @@ rules:
     priority: 1
     type: token
     pattern: |
-      # avoid coordinations
-      # we want any coordinated entities we might encounter to be split
       [word = /(?i)^(${increase_triggers}|${decrease_triggers}|${noncausal_increase_triggers}|${noncausal_decrease_triggers}|${cause_triggers}|${correlation_triggers}|${reverse_direction_cause_triggers})/] # triggers used downstream
       # This is if we want to no avoid the increase/decrease triggers at all...
       #[word = /(?i)^(${cause_triggers}|${correlation_triggers})/] # triggers used downstream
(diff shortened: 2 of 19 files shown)


bsharpataz:
I can do this too -- the one sentence I saw was a really broken syntax
sentence... but i can poke around some more

bsharpataz:
hold up -- please don't merge quite yet...

bsharpataz:
This is a cleaner fix -- but, i don't know which sentence I should make a test for.... there was a lack of specificity in some triggers (one issue on the trigger side, another with a rule), so...? Should I just grab any of the previously failing sentences?

kwalcock:
I don't know any better.  I was just going to verify that the code has been visited once and doesn't throw a new exception or otherwise blow up.

kwalcock:
So, as we discussed I'm starting another run across the 10 or so problem documents, the collection of 52, and then the 10K to check for issues.

bsharpataz:
woohoo! vaya con dios!

kwalcock:
It looked good on the first half last night, but waiting until morning I find error reports for these files:
2009_Document_a-bq992e.txt
2017_Document_a-i8219e.txt
2005_Book_a-am390e.txt
2013_Meeting_a-bk305e.txt
1964_Document_a-bm438e.txt
1987_Document_a-bm745e.txt
1978_Document_a-bm334e.txt
2008_Meeting_a-k2554e.txt
2017_Reports_a-bu002e.txt
Don't hesitate to request more details or help if it would be useful or if this can't be reproduced.  And ignore this until Monday, of course.

bsharpataz:
ok -- so, how about this.... i removed the exception bc Odin seems to gracefully handle the situation all by itself, so who am I to complain, right?
Aside from that, this journey did illuminate some issues with missing triggers so... yay?

@kwalcock please check for me and see, since i removed the exception, not it just gives a debug message and moves on, is this ok?

kwalcock:
The latest version made it through 10K+ documents without complaining and not causing other obvious problems due to its not complaining.  So, I'll merge it momentarily.  Thanks for working on it.

PR #415, opened 2018-08-22
Update to processors 7.4.1

kwalcock:


kwalcock:
Aren't we supposed to be changing from FastNLPProcessor to CluProcessor?  Is that to coincide with this update?

MihaiSurdeanu:
@kwalcock: We're not quite ready to make the switch to CluProcessor... But I would appreciate it if you told me what % of unit tests fail under CluProcessor now.

kwalcock:
Of the 353 tests run, 282 succeed and 71 fail with CluProcessor.  That's 20% failing.  Not all of these tests are related to the processor, though.

MihaiSurdeanu:
Thanks! What about FastNLPProcessor? How many fail with it? I'm trying to
understand how many fail because of change in processors vs something else.

kwalcock:
I hope that FastNLPProcessor is the one we're supposed to be using, because it's in the code.  None fail.  Wait!  That's with certain tests being intentionally ignored, which I completely forgot about.  If everything is turned back on, the numbers are

| Processor | Total | Fail | Fail% |
|-|-|-|-|
| CluProcessor | 550 | 240 | 44% |
FastNLPProcessor | 550 | 176 | 32% |

Standard settings, which were changed to all "it":
```
    val passingTest = it
    val failingTest = ignore
    val brokenSyntaxTest = ignore
    val futureWorkTest = ignore // added to mark the tests that are not currently passing, but with planned changes to the
                                // framework, they will be achievable
    val inferenceTest = ignore  // type of futureWorkTest -- added for tests which are now failing because of entity
                                // filtering, basically because inference or coref would be needed
    val tempBrokenEntitiesTest = ignore
    val affectEventTest = ignore
```

PR #417, opened 2018-08-28
Temporal attachment

EgoLaparra:
Failing tests:

```
- TestDoc3
-- Paragraph 14
--- "should have correct edges 4"
--- "should have correct edges 5"
```

To reproduce the error:

1. Copy "timenorm_model.hdf5" file into src/main/resources/org/clulab/wm/eidos/models/
2. Change "B-Time" to "B-Timex" in line 42 of EidosDocument.scala
3. Change "B-Time" to "B-Timex" in line 20 of TemporalAttachement.yml

The tests pass with one of the following changes in the text of paragraph 14:

- Ta'izz -> Taizz
- and/or -> and


kwalcock:
Thanks for excellent instructions.  I can reproduce this, but haven't gotten to the bottom of it yet.

kwalcock:
@EgoLaparra, I'm not having much luck with this, so I may as well write some notes and see if it helps.  Maybe even @bsharpataz will read this and have a bright idea.

As far as I recall, the only step necessary for me to produce the problem is to activate the timing system with `Copy "timenorm_model.hdf5" file into src/main/resources/org/clulab/wm/eidos/models/`.  Right now I'm using `D-Dinosaur` and `I-Dinosaur`.  Addressing the problem by changing the text seems to work because there can somehow be long distance affects that change how one sentence is parsed based on a previous sentence.  We had this problem earlier with POS.  I'll attach a screenshot of this happening when the same sentence was appended to itself and parsed

I am concerned about the overwriting of norms in EidosDocument.parseRealTime.  I have changed some code there to
```
case false if inTimex(0)._1 == 0 && mayOverwriteNorm(norm) => "D-Dinosaur" // keith was ere
case false if inTimex(0)._1 != 0 && mayOverwriteNorm(norm) => "I-Dinosaur"
...
// def mayOverwriteNorm(norm: String): Boolean = { println("This should fail"); true }
def mayOverwriteNorm(norm: String): Boolean = { println("This should pass"); norm == "O" }
```
If the value is not overwritten, then your rule cannot trigger and that might be why this works.  However, overwriting the value, but changing your rule so that it doesn't trigger, also fails.  It seems that overwriting is important.  It does change the hash value of the sentence and that is a potential problem because it could affect equality tests on objects that reference the sentence.  However, things like mentions are not yet in play here.

It's a long shot, but I have observed that the ordering of sets is extremely variable.  If this somehow affects the order of some set whose members are evaluated for being the best fit, then if there is a tie that goes to the first comer, the outcome may vary.  That ordering may depend on a hash value, come to think of it.

The closest I've come to seeing what was happening is with the two attached files.  One is produced with the passing test and one with the failing test.  I renumbered and reordered things to get the minimal difference.  The only difference in the data incoming into Eidos is that the passing test has `"norm" : "XXXX-07"` and the failing test has `"norm" : "B-Dinosaur"`.  This messes up the relations for mysterious reasons still.  I'm still stumped.

twoparse
[renumberedpass.txt](https://github.com/clulab/eidos/files/2338339/renumberedpass.txt)
[renumberedfail.txt](https://github.com/clulab/eidos/files/2338340/renumberedfail.txt)

kwalcock:
@EgoLaparra, have you seen any diagnostic output like this?

```
08:28:01.074 [pool-7-thread-4] INFO  org.clulab.wm.eidos.EidosSystem$ - timeNormModelPath: /org/clulab/wm/eidos/models/timenorm_model.hdf5
08:28:01.076 [pool-7-thread-4] INFO  o.d.nn.modelimport.keras.Hdf5Archive - Unexpected end-of-input: was expecting closing '"' for name
 at [Source: {"class_name": "Model",
...
"retu; line: 1, column: 4001]
```
and
```
08:28:01.077 [pool-7-thread-4] INFO  o.d.nn.modelimport.keras.Hdf5Archive - Unexpected end-of-input: was expecting closing quote for a string value
 at [Source: {"class_name": "Model",
...
line: 1, column: 8001]
```
or
```
08:28:01.079 [pool-7-thread-4] INFO  o.d.nn.modelimport.keras.Hdf5Archive - Unexpected end-of-input: was expecting closing '"' for name
 at [Source: {"class_name": "Model",
...
line: 1, column: 12001]
```
It looks like there might be a bug in the library.  Could I have a bad model file?  It's one I found sitting on my computer from sometime in the past.

For that matter, do you see any of these:
```
08:28:01.088 [pool-7-thread-4] WARN  o.d.n.m.keras.utils.KerasModelUtils - Could not read keras backend used (no backend field found)
08:28:01.088 [pool-7-thread-4] WARN  o.d.n.m.k.l.e.KerasEmbedding - Masking in keras and DL4J work differently. We do not completely support mask_zero flag on Embedding layers. Zero Masking for the Embedding layer only works with unidirectional LSTM for now. If you want to have this behaviour for your imported model in DL4J, apply masking as a pre-processing step to your input.See https://deeplearning4j.org/usingrnns#masking for more on this.
08:28:01.089 [pool-7-thread-4] WARN  o.d.n.m.k.l.e.KerasEmbedding - Masking in keras and DL4J work differently. We do not completely support mask_zero flag on Embedding layers. Zero Masking for the Embedding layer only works with unidirectional LSTM for now. If you want to have this behaviour for your imported model in DL4J, apply masking as a pre-processing step to your input.See https://deeplearning4j.org/usingrnns#masking for more on this.
08:28:01.181 [pool-7-thread-4] WARN  o.d.nn.modelimport.keras.KerasModel - Model cannot be trained: output layer timedistributed_1 is not an IOutputLayer (no loss function specified)
08:28:01.182 [pool-7-thread-4] WARN  o.d.nn.modelimport.keras.KerasModel - Model cannot be trained: output timedistributed_1 is not an IOutputLayer (no loss function specified)
08:28:01.182 [pool-7-thread-4] WARN  o.d.nn.modelimport.keras.KerasModel - Model cannot be trained: output layer timedistributed_2 is not an IOutputLayer (no loss function specified)
08:28:01.182 [pool-7-thread-4] WARN  o.d.nn.modelimport.keras.KerasModel - Model cannot be trained: output timedistributed_2 is not an IOutputLayer (no loss function specified)
08:28:01.182 [pool-7-thread-4] WARN  o.d.nn.modelimport.keras.KerasModel - Model cannot be trained: output layer timedistributed_3 is not an IOutputLayer (no loss function specified)
08:28:01.182 [pool-7-thread-4] WARN  o.d.nn.modelimport.keras.KerasModel - Model cannot be trained: output timedistributed_3 is not an IOutputLayer (no loss function specified)
```
?

kwalcock:
Pursuiant to the request at yesterday's meeting for more details...

@MihaiSurdeanu, at issue is the test in ord.clulab.wm.eidos.text.eval6.TestDoc3.scala which fails when @EgoLaparra adds his timing code.  It is specifically the tests
```
    val text ="""
        |In Yemen, second season (July to September) rainfall season started on time in July and has been above
        |average across western, central, and parts of eastern Yemen. Vegetation conditions are currently below
        |across many western coastal and mountainous areas of the country, with above-average vegetation conditions
        |in parts of Ta'izz and Ibb governorates. Although, the short-and long-term rainfall forecasts are favorable
        |for agricultural areas of the western and central highlands, production of most crops is likely to be
        |limited due to insecurity and lack of availability and/or access to farm inputs due to ongoing conflict.
        """

    val availability = NodeSpec("availability", Dec("lack"))
    val longFarmInput = NodeSpec("access to farm inputs due to ongoing conflict", Dec("lack"), Quant("ongoing"))
    val production = NodeSpec("production of most crops", Quant("likely"), Dec("limited"), Quant("most"))

    passingTest should "have correct edges 4" taggedAs(Somebody) in {
      tester.test(EdgeSpec(availability, Causal, production)) should be (successful)
    }
    passingTest should "have correct edges 5" taggedAs(Somebody) in {
      tester.test(EdgeSpec(longFarmInput, Causal, production)) should be (successful)
    }
```
These particular tests have been unstable and I had to tweak them most recently for the update to processors 7.4.1.  That didn't seem to be suspicious in itself because there are parentheses and "and/or" in the text which the newer processors tokenizes differently.  We could just tweak them again, but it seems better to figure out what is happening.

Egoitz points out above that very small changes in the text will work around the problem.  I reported that the same sentence parsed twice in context will give different results.  I don't think that either of these are necessarily errors.  Yesterday @bsharpataz found that just adding a space before the third sentence will change the Eidos output.  This seems like an error.

This "error" might not have anything to do with Egoitz' issue, but it seems to be at least confusing the matter, muddying the water, so that we can't tell what it going on.  It seems therefore worth pursuing.

To be continued...

kwalcock:
Regarding "I'm sorry, but i don't understand. Can you please explain the problem in plain language?":

Yesterday @bsharpataz briefly described the issue in this email attachment, below.  If an extra space is added in front of a sentence (the same one Egoitz is having problems with), the Eidos output changes.  A causal relationship, the circled one, disappears.  The parse coming from processors is exactly the same except for the offsets (which consequently change the hash value).  The table in the attachment has been expanded in a PR to verify this similarity, showing the offsets and also the raw word, which was recently added to processors.  This descrepancy with the space wasn't an issue with the previous processors, but then the parse was different anyway, so I don't think that an easy conclusion can be drawn, although I hoped that to be the case yesterday.

We're getting closer.

[issue.pdf](https://github.com/clulab/eidos/files/2354188/issue.pdf)

MihaiSurdeanu:
Thanks!

@kwalcock: can you please check if the second sentences are identical in the two scenarios (without or with additional spaces), other than the different offsets?
If they are, I would happily blaim the hashing function, and/or the way we use it in Eidos...

kwalcock:
All three input sentences are identical.  At times I've suspected three different culprits, but upon closer inspection they were innocent.  I'm getting close (famous last words) by running the two instances side by side.  Halfway through the execution and at some large number of stack frames deep I see the difference appearing and there aren't too many frames to go before it bottoms out.  Cautious optimism is the mood.

MihaiSurdeanu:
God speed @kwalcock!

kwalcock:
The likely problem has been identified and a fix is in the works.  It is within Eidos, so there shouldn't be wider repercussions.  Thanks for waiting.  I think it was worth it.

bsharpataz:
@EgoLaparra  and @kwalcock -- when I use this branch locally it doesn't work with the webapp, am I missing something?  Or is there something that needs to be fixed?

kwalcock:
Do you have the model file?

bsharpataz:
yep :)

bsharpataz:
ps -- locally I merged #425 onto this branch and tried re-creating the issue above and I _think_ it's fixed..... fyi

EgoLaparra:
Is the webapp still failing? What error do you get?

kwalcock:
Did you do these things?

Change "B-Time" to "B-Timex" in line 42 of EidosDocument.scala
Change "B-Time" to "B-Timex" in line 20 of TemporalAttachement.yml

Does the test fail when the webapp is not involved?

kwalcock:
@EgoLaparra, can this be checked in as is, or would you like, for example, to use Timex instead of time?

bsharpataz:
I mean -- i only merged it locally, I didn't push it up bc it's its own PR

bsharpataz:
huh? yes, I merged the other PR on, made the changes and ran with sbt
"testOnly *TestDoc3"

the webapp wasn't involved in this at all, sorry if I confused things with
that

EgoLaparra:
@kwalcock time is ok.

kwalcock:
@EgoLaparra, I'm about to to a PR on your branch to update the configuration and testing.  I thought it should go in before the merge.

kwalcock:
Does @EgoLaparra have a feel for the performance repercussions?

kwalcock:
@EgoLaparra, how do you usually run eidos with this enabled?  If it is run in an application, i.e., not as a test or in the webapp, then sbt puts your model file into a jar, which it doesn't look like you can use.

EgoLaparra:
It seems that something changed between sbt versions that affects this. With the version we are currently using in master (1.1.6),  I am also getting a FileSystemNotFoundException, but if I change to the version we were using before (0.13.16), it works.

Could you try changing sbt.version to 0.13.16 and run something like:

`sbt "runMain org.clulab.wm.eidos.apps.ExtractJsonLDFromFile example_output/debug_docs_example/demo-example.txt"`

and tell me if it works?

kwalcock:
@EgoLaparra, yes that works for me.  Is appears that the newer version of sbt puts all the files into a jar and into a temporary directory and tries to run from the jar, which means that the model can't be treated as a normal file anymore.  That can probably be fixed by finding the temporary directory and copying the file next to it, but that's not nice.  People accessing eidos via API will need to do something special about the file.  I only tracked down the problem to the point of Hdf5.file (?) where the methods started to be marked as native.  From that point I assume a file is necessary.  Do you happen to know whether there is a different interface to the model in that code somewhere?  If necessary, we could copy the file's contents out of the jar and into a temporary file somewhere.  The code may have to watch for jar: where there would otherwise be file: in that URL and deal with it.  Maybe @MihaiSurdeanu knows whether this has been an issue in other projects.

EgoLaparra:
The last version of deeplearning4j allows to import the model as an ImputStream. I'm trying to update the parser.

kwalcock:
Excellent, @EgoLaparra.  Thanks for looking into it.

PR #418, opened 2018-08-29
Cross sent coref

bsharpataz:
I think this is producing the desired result, but am not 100% sure it's what people want.  At the very least we should talk to @bgyori about how these mentions would be handled downstream (they're low freq I think).

If this is looking good, I can also add a unit test?


bgyori:
Happy to take a look but based on the PR alone I'm not sure what this will look like in practice. Could you post an example or a unit test that I can learn from? Thanks!

MihaiSurdeanu:
This looks fine to me.

Separate note: now that we do all this SD to UD changes, I wonder if we should a single class/object with POS/dependency labels, so all the representation-specific strings are in one place?

kwalcock:
That's a lot of code.  Congratulations.  Here's one nitpick: If there's still code like `val corefDeterminers = Set("this", "that", "these", "those")` inside a def, can it be moved to an object?

kwalcock:
I'm running this over a few documents (as a substitute for understanding everything that is happening in the code).  In the output of Doc5.txt it looks like one rule changed from
```
"rule" : "gradable-lexiconner++simple-np++quantification2++Increase_adjective_rule_1",
```
to
```
"rule" : "simple-np++gradable-lexiconner++quantification2++Increase_adjective_rule_1",
```
where the ordering simply changed.  That seems suspicious and I'm searching for suspicious things because of @EgoLaparra's difficulty.  Is this change explainable?

kwalcock:
I should make that Tester thing work for a coreference.  In a near future PR, I'll try to add it.

kwalcock:
There are some coreferences in the wild among the 52 document collection. 21, 22, 26, 28, 31, 42.

kwalcock:
In document 19, this extraction has changed from
```
    "@type" : "Extraction",
    "@id" : "_:Extraction_6",
    "type" : "concept",
    "subtype" : "entity",
    "labels" : [ "Concept", "Entity" ],
    "text" : "average conditions",
    "rule" : "gradable-lexiconner++simple-np++quantification2",
    "canonicalName" : "average conditions",
    "provenance" : [ {
      "@type" : "Provenance",
      "document" : {
        "@id" : "_:Document_1"
      },
      "documentCharInterval" : [ {
        "@type" : "Interval",
        "start" : 2162,
        "end" : 2179
      } ],
      "sentence" : {
        "@id" : "_:Sentence_13"
      },
      "positions" : [ {
        "@type" : "Interval",
        "start" : 5,
        "end" : 6
```
to
```
    "@type" : "Extraction",
    "@id" : "_:Extraction_7",
    "type" : "concept",
    "subtype" : "entity",
    "labels" : [ "Concept", "Entity" ],
    "text" : "early wetter than average conditions",
    "rule" : "simple-np++gradable-lexiconner++quantification2",
    "canonicalName" : "early wetter than average conditions",
    "provenance" : [ {
      "@type" : "Provenance",
      "document" : {
        "@id" : "_:Document_1"
      },
      "documentCharInterval" : [ {
        "@type" : "Interval",
        "start" : 2144,
        "end" : 2179
      } ],
      "sentence" : {
        "@id" : "_:Sentence_13"
      },
      "positions" : [ {
        "@type" : "Interval",
        "start" : 2,
        "end" : 6
```
in that the order of rules have changed and the text now starts earlier.  Any idea why?

bsharpataz:
@kwalcock  -- I have no idea why a rule name would change... :(  I just looked through the code again and don't see anything that would create a change.  Is it consistent across diff runs? (perhaps we can isolate some sentences and check only them?)

re: the strange ordering effects -- i looked at that too, both in the email you sent and in the other temporal PR and I have no idea what is happening at the moment....

bgyori:
@bsharpataz thanks! I understand, and I think given a link between the coref and the entity it resolves to in the JSON-LD, we will be able to use these. One thing I was confused about is, why are there two "Found Relations" blocks in your example? The second block looks correct but the first one lists "more recession" which wasn't in the original sentences - where does that come from?

bsharpataz:
@bgyori haha, just checking to see if you were paying attn.... yeah, i meant to paste over that, not add below... it was from the orig example, but i figured having "recession" in 2 places wasn't a good idea when trying to demonstrate the idea, so i changed it.  But thanks for letting me know, I removed the first block :)

kwalcock:
@bsharpataz, it is consistent across different runs.  That's good news.  I'll try to isolate an example.  If we (I) don't understand what is happening, it may as well be a neural network.

kwalcock:
@bsharpataz, I'm trying not to merge this until after the other issue is fixed for Egoitz, just in case this and the next few changes complicate that.  However, if the delay becomes a problem, I'll back down.  A fix for that other issue might also affect the rule change in the output noted above.

kwalcock:
@bsharpataz, could you take a look at the conflicts in EidosActions?  There aren't many, but they are important.

bsharpataz:
sure

kwalcock:
@bsharpataz, it's probably just a question of whether the foundBy still needs to change in EidosActions.  I'm paranoid of making changes in that file.

kwalcock:
Thanks, @bsharpataz.  It still passes all tests, even with the time being extracted.

PR #419, opened 2018-09-02
Rule sheets

bsharpataz:
To make the files that get will get posted on google sheets for annotating rules.  Very similar to what Ben has, but now in house and can be adapted to suit our needs and/or diff aspects of annotation (i.e. entities, grounding, polarity, etc)


kwalcock:
You aren't supposed to be working on Labor Day weekend!

PR #420, opened 2018-09-03
updated ontology to include interventions

bsharpataz:
fyi @ajaynagesh @ZhengTang1120

### DIFF
--- a/src/main/python/interventions_to_yaml.py
+++ b/src/main/python/interventions_to_yaml.py
@@ -10,12 +10,35 @@ yaml.add_representer(type(None), represent_none)
 import sys
 from collections import defaultdict
 
-def ont_node(name, examples):
-    return {'OntologyNode': None, "name": name, 'examples': examples, 'polarity': 1.0}
+# todo: check??
+syn_relations = {"Synonym", "B is a subset of A", "A is a subset of B", "Rough synonym"}
 
-def load_intervention_types(fn):
+def ont_node(name, examples, keywords):
+    # Make sure the node name is added to the examples to be used for grounding
+    examples.append(name)
+    d = {'OntologyNode': None, "name": name, 'examples': examples, 'polarity': 1.0}
+    if keywords is not None:
+        d['keywords'] = keywords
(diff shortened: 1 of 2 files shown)


bsharpataz:
- the '' I just took care of, thanks  
- The ones that seem "incorrect" come from the spreadsheet, e.g.:
image
(not that you can read that)
so... i am unsure what we should do, as this theoretically comes from the experts?
- why is `"hiv/aids" (line 185)` incorrect? it's under the antiretroviral node
- i merged the `i.e.` and `e.g.` lines up... but am not sure we should merge the e.g. ones... they're examples right?

MihaiSurdeanu:
@bsharpataz: Should I review this again?

code should live under src/main/PROGRAMMING_LANGUAGE/...

MihaiSurdeanu:
Ok

kwalcock:
What are these things: `- ''`?  It seems like the rest of the file was made so that the names of the non-leaf nodes contain no spaces, but these do.  The directory tree is on a per project basis.  I'm starting to think that there needs to be a separate project for the preparation of materials.  I need to add a program to normalize vectors, another to remove duplicates from ontologies, maybe another to pre-serialize vectors and ontologies,...

bsharpataz:
didn't all the '' get removed?

bsharpataz:
@MihaiSurdeanu I made the changes we talked about, namely to move keywords to their own field and to add the synonym info from the other sheet.  I opted to do less processing of text here (i.e. stop words) and allow grounding to have at it.  Please review, at least to some degree and let me know if this will pass muster :)

bsharpataz:
@kwalcock are we waiting on something before we merge this one?

kwalcock:
Sorry, I hadn't notice the next set of changes.  Are keywords used by the program?  I assume this unknown field is ignored.  Should "e.g." or "i.e." or "and" be part of them, does it matter when they end with a comma or are missing spaces: "counselling,pep,"?  If you say it's OK,  I'll just click the button.

PR #421, opened 2018-09-03
Expand better

bsharpataz:
Improving entities -- by allowing to expand on incoming `compound` and `amod` we can recapture the whole NP chunk (which was prevented in the current entityfinder bc we are avoiding triggers).  If we remove the reliance on avoiding in the EntityFinder, this will not be necessary, but until then...

Updated tests to match.

Example of change:
OLD:
image

NEW:
image


bsharpataz:
@kwalcock are we waiting on something for this merge?

kwalcock:
No, just concern that a change might somehow mask the problem with time.  There's probably a different way to test that, but I'm trying to be appropriately lazy.

PR #422, opened 2018-09-04
Kwalcock compact word2 vec

kwalcock:
Make a cached version of the vectors file


kwalcock:
@bsharpataz, you once wrote
```
// NOTE: The cachedOntologiesDir MUST be an absolute path!
```
I can't figure out why.  At least for my purposes, it's working as a relative path on all three operating systems.  You may had a different purpose.  Can you explain?  Thanks.

kwalcock:
@MihaiSurdeanu, can you check again whether this is OK?  As discussed at a recent meeting, the strategy of having the user cache the ontologies and vectors may soon be replaced.  In the meantime, some people (Ajay, Zheng, Becky) can save lots of runtime with this version.  Parts may also soon be transferred to Processors.

The directory used is ./cache/, which is created if necessary.  There's no longer a requirement to start the name with /, but ~ can't be readily used.  People can run CacheOntologies out of the box (if they have access to the vector files).  Eventually something can be done about that as well, but for the time being...

PR #423, opened 2018-09-05
Expand table output of webapp for easier diffing

kwalcock:


kwalcock:
Output now looks like the picture below.  It's at the bottom of the screen, so I hope it doesn't disturb anyone.

image

kwalcock:
Having both the (raw) text and the extracted word helps verify the tokenization.  The dependencies column is incomplete, but if the dependency hashes of two sentences match, then the complete dependencies should also match.

PR #425, opened 2018-09-10
preserve path for expanded arguments

marcovzla:
addresses problem related to different extractions depending on spaces

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/EidosActions.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosActions.scala
@@ -74,7 +74,7 @@ class EidosActions(val taxonomy: Taxonomy) extends Actions with LazyLogging {
       } else {
         // odin mentions keep track of the path between the trigger and the argument
         // below we assume there is only one cause arg, so beware (see require statement abov)
-        val landed = m.paths(arg2).values.head.last._2 // when the rule matched, it landed on this
+        val landed = m.paths(arg2)(m.arguments(arg2).head).last._2 // when the rule matched, it landed on this
         // when the rule matched, it landed on this
         // println("---------")
         // println(s"rule: ${m.foundBy}")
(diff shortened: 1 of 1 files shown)


kwalcock:
Thanks for working so long, so late on this.  I can't say that I understand it (yet).  I would have expected changes to createEventChain, but the combination of changes to expandArguments and reintroduction of keepMostComplete must force createEventChain to produce the correct/singular answer.  Still, "head" there makes me worry and so does "hopefully," although I like the positive attitude.  At some point, perhaps after all the experimentation is finished, this code can be documented more thoroughly.

kwalcock:
So, if it still looks good to you all this morning, I'll merge it :-)

bsharpataz:
I'm for it, I'd love for @marcovzla  to see the general form on lines 496 to 499 and give some thumbs up or down... also, how easily can we check to see if applying this takes care of #417?

kwalcock:
One could rewrite the hash function for Mentions, calculateHashCode, in a private copy of processors and then rerun Eidos on a large number of documents to make sure the answer doesn't change.  Adding - to the hash code might reverse all conversions from maps to lists.  That seems painful, though.

kwalcock:
FWIW I'm doing the above, so nobody else needs to be.

kwalcock:
I ran the unit tests and did quite a bit of regression tests on hundreds of files with the normal hash function and then the negation of it and see no differences in output with the fix, so I'm satisfied and will merge it finally (well, it was within a day, but it's holding up other older things).  @bsharpataz did need to adapt a test above, thanks, and there are differences in the output between before the fix and after the fix, but not very many.

For the GitHub record let me report that I also used a hash function that returns zero, creating a collision for every Mention.  This breaks the tests and changes quite a bit of the output.  The probable cause is this code in Mention:
```
  override def equals(that: Any): Boolean = that match {
    case that: Mention => that.canEqual(this) && this.hashCode == that.hashCode
    case _ => false
  }
```
It has been there for many years and I've just been crossing my fingers that there is never an important collision.  It doesn't seem to be noted as an issue in processors, so it will be added.

PR #426, opened 2018-09-11
Update configuration and testing

kwalcock:
Adjusted testing and configuration so that path need not be duplicated.


kwalcock:
@EgoLaparra, I just want to make sure these changes look reasonable to you.

EgoLaparra:
Looks great!

PR #428, opened 2018-09-13
Portuguese2

bsharpataz:
- Compartmentalized resources by language
- added resource.conf --> eidos.conf configuration file setup (which can be overridden on the command line -- see below).  
- Accordingly, changes to the loading of args from the configs and the imports in the grammar files.   
- Added Portuguese rules.  
- Since the Portuguese processor produces true UD POS tags (i.e., NOUN instead of NN etc), explicit tags were updated in several places to be inclusive.

to run the webapp in portuguese, override the language setting on the command line:
`sbt -DEidosSystem.language=portuguese  webapp/run`

(and other settings/params can be overridden in the same way)

@kwalcock -- tests for unicode fail, but we need to be able to have unicode in Portuguese and we should also be able to just gracefully handle it in English too (rather than require not having it, right?)


kwalcock:
I'll adjust the unicode check to probably skip Portuguese resources.  It's there because we had a lot of unnecessary unicode without ever knowing it was there.

bsharpataz:
But what if we need accents in the english grammars?  Should we have a
warning instead, if the issue is the fact that we didn’t know?  I thought
the issue had been the unicode in docs we were reading, not the resources,
or am i remembering wrong?

kwalcock:
There's a PR somewhere that allows the two files through.  There have been Unicode characters in the English resources and there still are in ones we don't control, although they won't necessarily cause a problem.  I don't know the best solution.  Maybe the test can become an app and be run once by people who write rules.  A similar app can be run by people who edit ontologies and might accidentally create duplicate nodes, etc.  Perhaps an "ignore" is called for and it can be changed to "it" once in a while.  I don't know to what extent various pieces can deal with Unicode and several times we have forgotten to open files in a way that can handle it.  I suspect that warnings will be ignored.

kwalcock:
I *really* like reference.conf and lack of file names in EidosSystem!

bsharpataz:
what's the status on this PR?

kwalcock:
I'm waiting for more of the reviewers.  Will also run a few more files shortly.

gcgbarbosa:
@kwalcock I will start writing tests tomorrow.

PR #429, opened 2018-09-13
Allow some Portuguese Unicode

kwalcock:


bsharpataz:
adding @marcovzla bc he designed Odin, so he can weigh in about what Odin can/can't handle.  If Odin is fine (I think it is?) then I feel like we prob shouldn't add a restriction/limitation in Eidos?

kwalcock:
I was thinking more of Processors and everything that it depends on, like Stanford NLP.  If we analyze "The α and ω are beginning and end." the words are "The alpha and omega are beginning and end" and any grammar with α or ω will probably? not match.  Someone has to "sign off" in order to get the test to pass.  But change as needed, including removing the test.

bsharpataz:
Yeah, i guess I was thinking that sometimes you want to make your grammar able to handle anything, unicode or not.  @marcovzla can you help decide when you get a chance?  You have the Odin/Reach experience to throw behind this convo :)
btw -- thanks @kwalcock for being thorough in what you do!

kwalcock:
This issue doesn't seem important enough to delay the Portuguese PR.  I'm getting tempted to merge this just so the other one can pass its tests.

bsharpataz:
yeah, let's go ahead and do it this way for now, and later we can revisit
the issue of unicode in the resources in general (perhaps at a meeting)

PR #430, opened 2018-09-13
Include the time model as resource

kwalcock:


kwalcock:
The added file needs to be accessible outside of a jar in order to be used, but it is now included with the source code.

MihaiSurdeanu:
@EgoLaparra: Can this file be loaded as a resource from the classpath in dl4j?

PR #431, opened 2018-09-17
Make Compact version of DomainOntology

kwalcock:
The cached copy of the MeSH CompactDomainOntology can be read in under one second.


kwalcock:
FWIW, this is likely to be checked in fairly soon, reviewed or not, because more changes need to be made on top of it.

MihaiSurdeanu:
This is nice. Not urgent, but at some point, this should be ported to processors...

PR #432, opened 2018-09-17
Hedging

bsharpataz:
and let there be hedging!

added heading component from Reach (please check docstring @MihaiSurdeanu to see if sufficiently ref'd), added some terms for our domain.  Also added a simplistic test for good measure! 

@kwalcock I have no idea how to dump what I would like to dump in terms of the attachment (ideally an Int that says the number of hedging terms found in the extraction such that downstream folks could say >0 = hedged).

@bgyori -- I hope that soon we will be dumping something helpful to you in the JSONLD -- heads up!


kwalcock:
Thanks for including a test :-)  Is this something that needs to be added to the CAG kinds of tests where these are found in relevant texts and should continue to be found?

PR #434, opened 2018-09-18
fix path issue

bsharpataz:


kwalcock:
You are right about the typo.  It should be corrected in the PR just before this one.

kwalcock:
Travis failed on some network operation, but this fix is included with the previous PR, so I'm just going to close without merging.  Thanks for trying!

PR #437, opened 2018-09-19
passing filename thru to document

bsharpataz:
and updates to the string sanitation in the export apps

fyi @marcovzla  :)


kwalcock:
Thanks, @bsharpataz!

bsharpataz:
thanks @kwalcock -- per a convo with @bgyori today,  when you run eidos, the docid should be something id-like, and the filename isn't quite it... we can change this apparatus to use docids, esp if we have metadata, this is kind of a placeholder perhaps?

Issue #435, opened 2018-09-19
Canonicalization of texts differs


kwalcock:
In EidosMention we have
```
  protected def canonicalFormSimple(m: Mention): String = {
    def isContentTag(tag: String) = tag.startsWith("NN") || tag.startsWith("VB")
    def removeNER(ner: String) = StopwordManager.STOP_NER.contains(ner)
    val contentLemmas = for {
      (lemma, i) <- m.lemmas.get.zipWithIndex
      tag = m.tags.get(i)
      ner = m.entities.get(i)
      if isContentTag(tag)
      if !stopwordManaging.containsStopword(lemma)
      if !removeNER(ner)
    } yield lemma

    if (contentLemmas.isEmpty)
      m.text // fixme -- better and cleaner backoff
    else
      contentLemmas.mkString(" ").trim.replaceAll(" +", " ")
//    println("  * result: " + contentLemmas.mkString(" "))
  }
```
while in TreeDomainOntology we have
```
    protected def realFiltered(text: String): Seq[String] = {
      val sentences = getSentences(text)

      sentences.flatMap { sentence =>
        sentence.words.zip(sentence.tags.get).filter { wordAndPos =>
          // Filter by POS tags which need to be kept (Nouns, Adjectives, and Verbs).
          wordAndPos._2.contains("NN") ||
            wordAndPos._2.contains("JJ") ||
            wordAndPos._2.contains("VB")
        }.map(_._1) // Get only the words.
      }
    }
```
One works on mentions and the other on text, but the effect should be the same: same filters, use of lemmas instead of words.  @kwalcock can make this happen, but will double check with assignees first.

MihaiSurdeanu:
I think this needs to be aligned between the two.

kwalcock:
@bsharpataz, should they both use the 3 parts of speech or just the two?

bsharpataz:
i'll prob make it a method in an object and use it both places
i can get to this soon (or you can...)

kwalcock:
I'm almost done, but should they both be
`tag.startsWith("NN") || tag.startsWith("VB")`
or
```
          wordAndPos._2.contains("NN") ||
            wordAndPos._2.contains("JJ") ||
            wordAndPos._2.contains("VB")
```

bsharpataz:
Whatever is in canonical name is prob more recent... but "should" is such a
strong word....

kwalcock:
I'm will go with 2 then.

MihaiSurdeanu:
[...]

PR #439, opened 2018-09-20
Canonicalize texts and ontology nodes the same way

kwalcock:
Namely,
```
  protected def isContentTag(tag: String) =
      tag.startsWith("NN") ||
      tag.startsWith("VB")

  def isCanonical(lemma: String, tag: String, ner: String): Boolean =
    isContentTag(tag) &&
    !stopwordManaging.containsStopword(lemma) &&
    !StopwordManager.STOP_NER.contains(ner)
```

Also, not only do descriptions pass through this check, but so do examples and even node names.  Node names are now also added to the values against which the nodes are grounded.  _ characters are replaced with spaces in node names.

The vectors had been converted to Floats, but now so have the outputs of word2vec operations.  This is done to conserve memory, especially since incoming data is at less than float precision.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,13 @@ language: scala
 scala:
   - 2.12.4
   - 2.11.11
-before_cache:
-  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
-  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
-cache:
-  directories:
-    - $HOME/.gradle/caches/
-    - $HOME/.gradle/wrapper/
-    - $HOME/.ivy2
-    - $HOME/.m2
-    - $HOME/.sbt
+#before_cache:
+#  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
+#  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
+#cache:
+#  directories:
+#    - $HOME/.gradle/caches/
+#    - $HOME/.gradle/wrapper/
+#    - $HOME/.ivy2
+#    - $HOME/.m2
+#    - $HOME/.sbt
--- a/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
@@ -238,7 +238,7 @@ class EidosSystem(val config: Config = ConfigFactory.load("eidos")) extends Stop
   def populateSameAsRelations(ms: Seq[Mention]): Seq[Mention] = {
 
     // Create an UndirectedRelation Mention to contain the sameAs grounding information
-    def sameAs(a: Mention, b: Mention, score: Double): Mention = {
+    def sameAs(a: Mention, b: Mention, score: Float): Mention = {
       // Build a Relation Mention (no trigger)
       new CrossSentenceMention(
         labels = Seq("SameAs"),
@@ -338,7 +338,7 @@ class EidosSystem(val config: Config = ConfigFactory.load("eidos")) extends Stop
   /*
       Wrapper for using w2v on some strings
    */
-  def stringSimilarity(string1: String, string2: String): Double = wordToVec.stringSimilarity(string1, string2)
+  def stringSimilarity(string1: String, string2: String): Float = wordToVec.stringSimilarity(string1, string2)
 
   /*
      Debugging Methods
--- a/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
@@ -97,14 +97,16 @@ class EidosSystem(val config: Config = ConfigFactory.load("eidos")) extends Stop
     def useTimeNorm: Boolean = eidosConf[Boolean]("useTimeNorm")
     def    useCache: Boolean = eidosConf[Boolean]("useCache")
 
+    val canonicalizer = new Canonicalizer(EidosSystem.this)
+
     protected def mkDomainOntology(name: String): DomainOntology = {
       val serializedPath: String = DomainOntologies.serializedPath(name, cacheDir)
 
       name match {
-        case   UN_NAMESPACE =>   UNOntology(  unOntologyPath, serializedPath, proc, useCache = useCache)
-        case  WDI_NAMESPACE =>  WDIOntology( wdiOntologyPath, serializedPath, proc, useCache = useCache)
-        case  FAO_NAMESPACE =>  FAOOntology( faoOntologyPath, serializedPath, proc, useCache = useCache)
-        case MESH_NAMESPACE => MeshOntology(meshOntologyPath, serializedPath, proc, useCache = useCache)
+        case   UN_NAMESPACE =>   UNOntology(  unOntologyPath, serializedPath, proc, canonicalizer, useCache = useCache)
+        case  WDI_NAMESPACE =>  WDIOntology( wdiOntologyPath, serializedPath, proc, canonicalizer, useCache = useCache)
+        case  FAO_NAMESPACE =>  FAOOntology( faoOntologyPath, serializedPath, proc, canonicalizer, useCache = useCache)
+        case MESH_NAMESPACE => MeshOntology(meshOntologyPath, serializedPath, proc, canonicalizer, useCache = useCache)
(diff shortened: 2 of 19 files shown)


kwalcock:
@ajaynagesh, Mihai wanted some kind of evaluation of whether these changes help or hurt performance and claimed that you have a way to evaluate it.  Can you do so?

It might also be good to test whether adding "JJ" to the acceptible parts of speech helps.  It was in the code for the ontologies but not for the document texts.

Also I finally notice that the old code for text used startsWith("NN") while the code for ontologies has contains("NN").  I'm not familiar enough with the tagset to know the repurcussions, so it would be great if @ajaynagesh and @bsharpataz weighed in.  Contains is of course more general.  Thanks.

ajaynagesh:
Certainly @kwalcock! Will let you know about this soon ...

kwalcock:
Also, in un_ontology.yml I notice a lot of `provision of provision of` and wonder if there was some sort of search and replace error.  I'm not sure what that phrase means.

kwalcock:
@ajaynagesh, are this and #435 going OK?

bsharpataz:
> Also, in un_ontology.yml I notice a lot of `provision of provision of` and wonder if there was some sort of search and replace error. I'm not sure what that phrase means.

Sorry -- meant to reply to this sooner.  This comes from the way I converted the sheet from Kimetrica into a format for us.  In general their specs recommend for them to avoid including the "provision of" in the node name, however, for grounding, I was thinking it would be helpful (to ground to the _provision of_ food rather than _food_ itself?), so I inserted it.  However, they didn't always follow their own spec, so sometimes it's duplicated.  This could be fixed, but IDK if it matters in terms of the way it's all used in the grounding algorithm...

kwalcock:
I haven't seen that, but can imagine what is going wrong.  I'm checking.

kwalcock:
Sorry about that.  It should be fixed.

ajaynagesh:
Thanks @kwalcock , It is working currently. I am evaluating both. I have run `ExtractAndExport` on both `master` and `kwalcock-canononical` and evaluating the results. I think there is a slight change in the ontology, so I have regenerate the gold standard grounding for this again. 

I will report the numbers soon. Upon preliminary examination of the grounding outputs, there is little difference between both the versions. In a few cases, the kwalcock-cononical has better groundings compared to the master.

ajaynagesh:
Ok here are the results: 

master: 69.27% 
kwalcock-canonical: 66.67% 

Performance is more or less similar. But the master has a slightly better performance. Also I noticed that with some of the new intervention nodes in the ontology, there seems to be more errors in grounding. This is present in both master and kwalcock-canonical. 

The detailed spreadsheet is present at: https://docs.google.com/spreadsheets/d/1CNb0swt1EBJOX7rD74S39GU0NCUZQvGM2Ky14ZxELLg/edit#gid=0

kwalcock:
Thanks, @ajaynagesh.  I wonder if this means that adjectives should be added back.  The code was once
```
          wordAndPos._2.contains("NN") ||
            wordAndPos._2.contains("JJ") ||
            wordAndPos._2.contains("VB")
```
We've also got code soon to be included that says
```
  def contentWords(ws: Seq[String]): Seq[String] = {
//    def contentTag(s: String): Boolean = {
//      s.startsWith("N") || s.startsWith("V") || s.startsWith("J")
//    }
//
//    val doc = proc.mkDocument(ws.mkString(" "))
//    proc.tagPartsOfSpeech(doc)
//    for {
//      (w, i) <- doc.sentences.head.words.zipWithIndex
//      tag = doc.sentences.head.tags.get(i)
//      if contentTag(tag) && w != "provision"
//    } yield w
    ws//.filter(!_.startsWith("provision"))
  }
```
We have differing parts of speech, either startsWith or contains, and words or lemmas.  It seems like these should all be the same.

ajaynagesh:
Hi @kwalcock , @MihaiSurdeanu : 

Ran with the adjectives included in kwalcock-canonical. Basically, the code is the following now: 

`class Canonicalizer(stopwordManaging: StopwordManaging) {

  protected def isContentTag(tag: String) =
      tag.startsWith("NN") ||
      tag.startsWith("VB") ||
      tag.startsWith("JJ")

  def isCanonical(lemma: String, tag: String, ner: String): Boolean =
    isContentTag(tag) &&
    !stopwordManaging.containsStopword(lemma) &&
    !StopwordManager.STOP_NER.contains(ner)
}`

The results drop a little bit more. So the results now are following: 

`master: 69.27%
kwalcock-canonical: 66.67%
kwalcock-canonical-adj: 64.06%` 

results are updated in https://docs.google.com/spreadsheets/d/1CNb0swt1EBJOX7rD74S39GU0NCUZQvGM2Ky14ZxELLg/edit#gid=0

kwalcock:
Thanks, @ajaynagesh.  Is this with all ontolgies?

ajaynagesh:
It is with un_ontology only. Also I am evaluating with only un_ontology currently.

kwalcock:
Just so people know, the UN ontology has only examples rather than descriptions.  Examples were not being filtered at all previously.  That was reserved for descriptions.  In addition, the name of the leaf node has been added as well as names of all the parents.  I see a field in the ontology that I didn't notice before, keywords.  The program is ignoring those.

ajaynagesh:
I would vote for merging `kwalcock-canonical` as is .. given the results with `adj` bumps down the accuracy a little bit on our small test set ..

MihaiSurdeanu:
@kwalcock: Got it. I think it's Ok to ignore keywords. Can we run one more test: ignore the names of all the parents?

kwalcock:
Well, I wouldn't want to add to Ajay's workload.

The names would need to be commented out of these lines of code in TreeDomainOntology.scala for the test.  Later I can remove some of the unnecessarily calculations.
```
class OntologyLeafNode(nodeName: String, parent: OntologyBranchNode, polarity: Float, names: Seq[String], examples: Option[Seq[String]] = None, descriptions: Option[Seq[String]] = None) extends OntologyNode(nodeName, parent) with Namer {

  def name: String = fullName

  // Right now it doesn't matter where these come from, so they can be combined.
  val values: Array[String] = (/*names ++ */ examples.getOrElse(Seq.empty) ++ descriptions.getOrElse(Seq.empty)).toArray

```

ajaynagesh:
I could run this test, thanks @kwalcock. I will comment the lines, run and report the results soon ...

ajaynagesh:
@kwalcock  Do I need to comment the entire `OntologyLeafNode` class ? With your suggested changes I am getting the following error: 

screen shot 2018-10-02 at 8 05 13 pm

kwalcock:
Sorry, just the /*names ++ */ part.

ajaynagesh:
I did not read your code snippet correctly! Thanks!

ajaynagesh:
With the parents removed (only leaves retained), I get identical performance on the test set i.e **64.06%** 

I checked that there are some differences in the outputs produced, just as a sanity check to ensure I am not plotting the same run again.

bsharpataz:
Ok gentlemen.  I am going to try making a change to the grounding procedure, which will remove the text from any attachment trigger+quantifiers from the text used to ground.  What version of what branch/master should I start with?  And @ajaynagesh when I am done, I'll push it here and perhaps you can see how it affects the grounding precision?

kwalcock:
@ajaynagesh, I assume the most recent test without the node names was with the adjectives so that the conclusion might be to skip the node names and leave out adjectives, but possibly keep the consistency.  I was waiting for @MihaiSurdeanu to make a ruling before doing anything.  It shouldn't be long, @bsharpataz, and then it will be in master.

MihaiSurdeanu:
I am confused by Keith's latest comment. Can somebody please summarize the latest grounding results in this thread for the configurations tested?
Thanks!

kwalcock:
master: 69.27%
kwalcock-canonical: 66.67%
kwalcock-canonical-adj: 64.06%
With the parents removed (only leaves retained), I get identical performance on the test set i.e 64.06%

I just wonder if the last is kwalcock-canonical-noParents or kwalcock-canonical-adj-noParents.  Probably the latter.

bsharpataz:
you know, we could also try using the same intuition as what I did in the mapper --> add in (interpolate in) a weighted avg of the similarities between the extraction canonical name and the ancestors of the node (I used reciprocal rank for the weights)....
we can try that once it gets merged too....

PR #441, opened 2018-09-21
regex bug

bsharpataz:
let's see if travis is ok with this...?


kwalcock:
The caches on Travis seem to be messed up.  On some of my tests I had to delete the caches and then even unconfigure the caching.  I'll see what can be done here.  The regex isn't the problem.

kwalcock:
Is this just a Travis test, or should it really be merged?

bsharpataz:
Merged :)

PR #445, opened 2018-09-26
added relation mentions to webapp display

bsharpataz:
added relation mentions to webapp display

### DIFF
--- a/webapp/app/controllers/HomeController.scala
+++ b/webapp/app/controllers/HomeController.scala
@@ -66,9 +66,9 @@ class HomeController @Inject()(cc: ControllerComponents) extends AbstractControl
 
   // Entry method
   def parseText(text: String, cagRelevantOnly: Boolean) = Action {
-    val (doc, eidosMentions, groundedEntities, causalEvents) = processPlaySentence(ieSystem, text, cagRelevantOnly)
+    val (doc, eidosMentions, groundedEntities) = processPlaySentence(ieSystem, text, cagRelevantOnly)
     println(s"Sentence returned from processPlaySentence : ${doc.sentences.head.getSentenceText}")
-    val json = mkJson(text, doc, eidosMentions, groundedEntities, causalEvents)
+    val json = mkJson(text, doc, eidosMentions, groundedEntities)
     Ok(json)
   }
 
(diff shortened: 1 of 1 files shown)


kwalcock:
Do you have a simple sentence I can test with?

kwalcock:
@bsharpataz, can you check this code?  In the top one I see `e.arguments.values` and in the bottom `e.arguments.values.flatten`.  I don't think the top one will ever satisfy `if a.isInstanceOf[EventMention]` because e.arguments.values will be an Iterable[Seq[Mention]] and a will be Seq[Mention].  Is there a sentence that can be used to test this?  The code is probably never used in an important place.

```
    // collect triggers for event mentions
    val triggers = events.flatMap { e =>
      val argTriggers = for {
        a <- e.arguments.values
        if a.isInstanceOf[EventMention]
      } yield a.asInstanceOf[EventMention].trigger
      e.trigger +: argTriggers.toSeq
    }
    // collect event arguments as text bound mentions
    val entities = for {
      e <- events ++ relations
      a <- e.arguments.values.flatten
    } yield a match {
      case m: TextBoundMention => m
      case m: RelationMention => new TextBoundMention(m.labels, m.tokenInterval, m.sentence, m.document, m.keep, m.foundBy)
      case m: EventMention => m.trigger
    }
```

kwalcock:
In a local version I've gone and replaced things like this
```
    val topLevelTBM = mentions.flatMap {
      case m: TextBoundMention => Some(m)
      case _ => None
    }
```
with this
```
    val topLevelTBM = mentions.collect { case m: TextBoundMention => m }
```
Apparently that's the Scala way and it's probably more efficient as well.  I'll merge it after hearing about the issue above.

bsharpataz:
> @bsharpataz, can you check this code? In the top one I see `e.arguments.values` and in the bottom `e.arguments.values.flatten`. I don't think the top one will ever satisfy `if a.isInstanceOf[EventMention]` because e.arguments.values will be an Iterable[Seq[Mention]] and a will be Seq[Mention]. Is there a sentence that can be used to test this? The code is probably never used in an important place.
> 
> ```
>     // collect triggers for event mentions
>     val triggers = events.flatMap { e =>
>       val argTriggers = for {
>         a <- e.arguments.values
>         if a.isInstanceOf[EventMention]
>       } yield a.asInstanceOf[EventMention].trigger
>       e.trigger +: argTriggers.toSeq
>     }
>     // collect event arguments as text bound mentions
>     val entities = for {
>       e <- events ++ relations
>       a <- e.arguments.values.flatten
>     } yield a match {
>       case m: TextBoundMention => m
>       case m: RelationMention => new TextBoundMention(m.labels, m.tokenInterval, m.sentence, m.document, m.keep, m.foundBy)
>       case m: EventMention => m.trigger
>     }
> ```

yeah -- I think you're right.... and i also think you're right that it's a not-very-used part of the code, I think it only hits there if the arg of an EventMention is itself an EventMention, but YEP i but this wouldn't work :)  Care to throw a `.flatten` in there for me? and THANKS for the thorough reviewing!

Issue #444, opened 2018-09-24
Hedging in JSON-LD


bgyori:
PR #432 added the extraction of hedgings to mentions with a test sentence ```Rainfall is likely to cause flooding.``` Running this sentence in the shell, I found that it produced the expected Hedging attachment:
```
events: 
List(Causal, DirectedRelation, EntityLinker, Event) => Rainfall is likely to cause flooding 
	 ------------------------------ 
	 Rule => ported_syntax_1_verb-Causal 
	 Type => EventMention 
	 ------------------------------ 
	 trigger => cause 
	 cause (Concept, Entity) => Rainfall 
	  * Attachments: Quantification(likely,None) 
	 effect (Concept, Entity) => flooding 

	 Event Attachments: Hedging(likely) 
	 ------------------------------ 
```
However, I can't find this attachment in the JSON-LD output in any form. Are we sure it appears in the JSON-LD? Sorry if I'm just missing something!

kwalcock:
@bsharpataz, did you see this?  I don't remember it being on my list, but then there are lots of things I don't remember, most of them in fact.

bsharpataz:
closed with #446

PR #446, opened 2018-09-28
Converted hedging to a TriggeredAttachment

bsharpataz:
... to help it go through to the jsonld.  Format is now trigger is the
num of hedging words (as a string) and the quantifiers holds the list of
the hedging terms

@kwalcock if there is a cleaner solution please feel free to modify however you want!

@bgyori I verified, and with this branch hedging gets through to the jsonld.  for example, for the sentence: `Rainfall might cause increased flooding.` we get:
```...
{
    "@type" : "Extraction",
    "@id" : "_:Extraction_1",
    "type" : "relation",
    "subtype" : "causation",
    "labels" : [ "Causal", "DirectedRelation", "EntityLinker", "Event" ],
    "text" : "Rainfall might cause increased flooding",
    "rule" : "ported_syntax_1_verb-Causal",
    "canonicalName" : "rainfall cause increase flooding",
    "provenance" : [ {
      "@type" : "Provenance",
      "document" : {
        "@id" : "_:Document_1"
      },
      "documentCharInterval" : [ {
        "@type" : "Interval",
        "start" : 1,
        "end" : 39
      } ],
      "sentence" : {
        "@id" : "_:Sentence_1"
      },
      "positions" : [ {
        "@type" : "Interval",
        "start" : 1,
        "end" : 5
      } ]
    } ],
    "states" : [ {
      "@type" : "State",
      "type" : "HEDGE",
      "text" : "1",
      "modifiers" : [ {
        "@type" : "Modifier",
        "text" : "might"
      } ]
    } ],
...
```

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/apps/ExtractAndExport.scala
+++ b/src/main/scala/org/clulab/wm/eidos/apps/ExtractAndExport.scala
@@ -5,10 +5,10 @@ import java.io.PrintWriter
 import ai.lum.common.StringUtils._
 import com.typesafe.config.{Config, ConfigFactory}
 import org.clulab.utils.Serializer
-import org.clulab.odin.{EventMention, Mention, State}
+import org.clulab.odin.{Attachment, EventMention, Mention, State}
 import org.clulab.serialization.json.stringify
 import org.clulab.utils.Configured
-import org.clulab.wm.eidos.attachments.{Decrease, Increase, Quantification}
+import org.clulab.wm.eidos.attachments.{Decrease, Hedging, Increase, Quantification}
 import org.clulab.wm.eidos.groundings.EidosOntologyGrounder
 import org.clulab.wm.eidos.mentions.{EidosEventMention, EidosMention}
 import org.clulab.wm.eidos.{AnnotatedDocument, EidosSystem}
@@ -18,8 +18,6 @@ import org.clulab.wm.eidos.utils.FileUtils.{findFiles, printWriterFromFile}
 import org.clulab.wm.eidos.utils.GroundingUtils.{getBaseGrounding, getGroundingsString}
 
 import scala.collection.mutable.ArrayBuffer
-
-
 
 
 /**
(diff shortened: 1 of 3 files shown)


bsharpataz:
I think this closes #444

kwalcock:
I need to document this on the JSON-LD page.  I'm a little concerned about text: "1" and how that compares to other usages.  I can't tell from the documentation what we were thinking there.  Merge shortly.

MihaiSurdeanu:
@bsharpataz: can you please clarify the JSON format?

kwalcock:
I added documentation on the JSON-LD page.  It seems that by analogy to INC, DEC, QUANT, the first text in the helpful! example would be "might" rather than "1" and modifiers would be for other words as in "definitely might".  It's probably a side-effect of some implementation detail I haven't checked on, but for INC we have provenance.  Is it not available here?  I probably need a memory refresh.

bsharpataz:
> I added documentation on the JSON-LD page. It seems that by analogy to INC, DEC, QUANT, the first text in the helpful! example would be "might" rather than "1" and modifiers would be for other words as in "definitely might". It's probably a side-effect of some implementation detail I haven't checked on, but for INC we have provenance. Is it not available here? I probably need a memory refresh.

Yes -- by analogy that's what it would be, but I was thinking that in terms of what the downstream users need, they really just need a number to convey how hedged it is.  I guess it could also be just like the other (INC, DEC, etc), and the assemblers could count how many hedging attachments there are...?  Thoughts?

MihaiSurdeanu:
I think we should stick with a number, because that number is theoretically unbounded. It would be hard to come with verbose descriptions for all.

bsharpataz:
> I think we should stick with a number, because that number is theoretically unbounded. It would be hard to come with verbose descriptions for all.

I think we may go with the standard TriggeredAttachment like Inc/Dec and ask the assemblers to count how many Hedge attachments there are (or rather, look for the presence of a Hedge attachment).  This would port well to negation too (i.e., look for existence of a Negation)

bsharpataz:
@bgyori @MihaiSurdeanu @kwalcock 
updated main comment to reflect current design.  
TLDR you can look for presence of HEDGE on a Causal relation or count them, however you want!
I plant to do negation similarly (except for the counting, i.e. just presence/absence)

bgyori:
Sure, this looks good to me! Once this PR and negations are merged, I'll update INDRA to extract both.

kwalcock:
@bsharpataz, did you notice the failing tests?  It doesn't look like a Travis problem.

PR #448, opened 2018-09-29
Indicator mapper tool

bsharpataz:
Can be run as a stand-alone, but also gets run when CacheOntologies is run.  The mapping file will be stored wherever is specified in the config: `apps.ontologymapper.outfile` and the number stored in `apps.groundTopN` will be used to determine how many of the mappings are printed for each un_ont node (per indicator ontology).

fyi @adarshp


adarshp:
Thanks for the heads-up!

kwalcock:
Thanks, @bsharpataz.  I wonder if there is some reason in OntologyMapper not to do
```
val reader = new EidosSystem()
val proc = reader.proc
val config = reader.config
```
since those values are all exposed.  I think there is also a org.clulab.embeddings.word2vec.Word2Vec.dotProduct available.

kwalcock:
I'm a little concerned about how this code
```
  // Filter out non-content words or not
  def selectWords(ws: Seq[String], contentOnly: Boolean): Seq[String] = {
     def contentTag(s: String): Boolean = {
      s.startsWith("N") || s.startsWith("V") || s.startsWith("J")
    }
     if (contentOnly) {
      val doc = proc.mkDocument(ws.mkString(" "))
      proc.tagPartsOfSpeech(doc)
      for {
        (w, i) <- doc.sentences.head.words.zipWithIndex
        tag = doc.sentences.head.tags.get(i)
        if contentTag(tag) && w != "provision"
      } yield w
    }
    else {
      ws
    }
  }
```
compares to something unfortunately not merged yet (#439),
```
class Canonicalizer(stopwordManaging: StopwordManaging) {

  protected def isContentTag(tag: String) =
      tag.startsWith("NN") ||
      tag.startsWith("VB")

  def isCanonical(lemma: String, tag: String, ner: String): Boolean =
    isContentTag(tag) &&
    !stopwordManaging.containsStopword(lemma) &&
    !StopwordManager.STOP_NER.contains(ner)
}
```
However, I guess this is a rough draft, so I shouldn't worry too much about it.

PR #449, opened 2018-09-29
Do not time-parse sentences without DATE in entities. DCT is passed a…

EgoLaparra:
…s a string and parsed by the temporal parser

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,8 +34,8 @@ val minorVersionRegex = "\\d+\\.(\\d+).*".r
 libraryDependencies ++= {
   scalaVersion { sv =>
     sv match {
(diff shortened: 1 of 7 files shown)


kwalcock:
The config file is probably not right for Travis.  All change it.

kwalcock:
Is there any progress on reading the model from a stream or otherwise from the jar file?  If necessary, I can extract it into a real file for use with timenorm.  For the time being there is a hard-coded temporary fix that eventually needs to disappear.

EgoLaparra:
DL4J does not still support loading the model as stream and, as it need the path to file, I cannot think any solution besides yours.

When running a test, the project is compiled but not packaged, so it doesn't try to read from the jar and it doesn't produce any exception. Is it possible to configure the lifecycle of the "sbt run"  command to avoid the package step?

kwalcock:
The 52 files completed in 1870 seconds on my computer, or 31 minutes.  That looks good.

PR #450, opened 2018-10-02
Negation

bsharpataz:
Added negation from Reach.  Multiple negations are "handled such that if > 1, even numbers get reduced to 0 (i.e., no negation) and odd numbers get reduced to 1, and the first at that.  Thus, downstream users (@bgyori, @adarshp) can check for presence of a `NEGATION` state on an event.

Currently (and ever?) only applies to Relations.

Here is what negation looks like in the jsonld for the sentence "Rain doesn't cause increased puppies.":

```
{
    "@type" : "Extraction",
    "@id" : "_:Extraction_1",
    "type" : "relation",
    "subtype" : "causation",
    "labels" : [ "Causal", "DirectedRelation", "EntityLinker", "Event" ],
    "text" : "Rainfall doesn't cause increased puppies",
    "rule" : "ported_syntax_1_verb-Causal",
    "canonicalName" : "rainfall cause increase puppy",
    "provenance" : [ {
      "@type" : "Provenance",
      "document" : {
        "@id" : "_:Document_1"
      },
      "documentCharInterval" : [ {
        "@type" : "Interval",
        "start" : 1,
        "end" : 40
      } ],
      "sentence" : {
        "@id" : "_:Sentence_1"
      },
      "positions" : [ {
        "@type" : "Interval",
        "start" : 1,
        "end" : 6
      } ]
    } ],
    "states" : [ {
      "@type" : "State",
      "type" : "NEGATION",
      "text" : "n't"
    } ]
```

fwiw @kwalcock some of the changes will no longer be changes if #446 goes thru...

### DIFF
--- a/shell
+++ b/shell
@@ -1,2 +1,2 @@
 #!/bin/bash
-sbt 'run-main org.clulab.wm.eidos.apps.EidosShell'
+sbt 'runMain org.clulab.wm.eidos.apps.EidosShell'
--- a/shell.bat
+++ b/shell.bat
@@ -1,1 +1,1 @@
-sbt "run-main org.clulab.wm.eidos.apps.EidosShell"
+sbt "runMain org.clulab.wm.eidos.apps.EidosShell"
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -18,12 +18,5 @@ EidosSystem {
   timeNormModelPath = /org/clulab/wm/eidos/models/timenorm_model.hdf5
            cacheDir = ./cache/${EidosSystem.language}
              useW2V = false
-           useCache = true
-}
-apps {
-  inputDirectory = "watermelon/"
-  outputDirectory = "canteloupe/"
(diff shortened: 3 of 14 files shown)


kwalcock:
I like that variable djskdjs!

bsharpataz:
hahaha did I leave that in? let me remove it :)

bsharpataz:
@kwalcock where is it?

kwalcock:
You removed it, but not before git made a recording.

We really need Mention.withAttachments() so that things like this are not needed.  This is at least the second time it has appeared, so I'm going to file an issue for it.  I should also eventually do a pass and change vars to vals where the r is a typo.
```
  // If any hedging was found, store it as an attachment
  def withApplicableHedging(m: Mention, hedgedLemmas: Seq[String]): Mention = {
    var finalMention = m
    for (hl <- hedgedLemmas) {
      finalMention = finalMention.withAttachment(Hedging(hl, None))
    }
    finalMention
  }
```

PR #451, opened 2018-10-03
Minor things

bsharpataz:
- changed the order of expanding (now first on limited incoming, then liberal outgoing)
- rule mod
- removed a redundant keepMostComplete
- changes to 2 tests to accommodate changed behavior


kwalcock:
Thanks for a few new passing tests!

bsharpataz:
> Thanks for a few new passing tests!

@kwalcock in fact, one todo on my list is to check and see if any old failing/ignored tests now pass :)

PR #453, opened 2018-10-04
added dec from neg dep on nouns, duh.

bsharpataz:
shoulda done this way earlier...

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/english/grammars/explicitModifiers.yml
+++ b/src/main/resources/org/clulab/wm/eidos/english/grammars/explicitModifiers.yml
@@ -34,6 +34,16 @@ rules:
       trigger = [word=/(?i)^(limit|little)/] to no
       theme: Entity = <neg
 
+  - name: Decrease_explicit_neg_dep
+    priority: ${ rulepriority }
+    example: "... was no exchange of goods ..."
+    label: Decrease
+    action: ${ action }
+    pattern: |
+      trigger = [incoming=neg]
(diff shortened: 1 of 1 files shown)


bsharpataz:
though... this sentence brought up a limitation to the negation handler:
“ Trade worsened during the conflict as there was no exchange of goods from Sudan and the other states of South Sudan , ” he added .

image

image

@MihaiSurdeanu how general do you think it is that if the `neg` dep lands on "no" then it's not negating the event?

MihaiSurdeanu:
@bsharpataz: I don't know. I have *no* opinion if this decision *impacts* performance. (see the example? :))
Shouldn't we force the neg dependency to be headed by the predicate?

Also, why does this Correlation rule match here? Certainly a copulative verb is no indicator of correlation...

bsharpataz:
re: "Shouldn't we force the neg dependency to be headed by the predicate?" 
I'm looking in this PR for negs off of nouns to count as "DEC" of that noun.  In the negation handler, I think that negs coming off of anything on the event token interval count, I ported from Reach, so I guess that was super intentional?

re: correlation -- the rule actually looks for a word with incoming advcl_as: `trigger = [incoming=advcl_as]`, just so happens it was "was" here.  The trigger here is less triggery than most rules... I could make a special action that removes the trigger and turns it into a RelationMention if you want...

kwalcock:
Conclusion?  Go for it?

MihaiSurdeanu:
Ok then!

I think this is reasonable then. Please merge.

PR #455, opened 2018-10-04
added rule confidence summary

bsharpataz:
FYI @bgyori this is the file from sheet 2 of "your" google sheets file, has each rule and the annotation details, including precision, frequency, and level of curation.  Please let me know if this is difficult to ingest in any way!

@MihaiSurdeanu and @kwalcock please let me know if the placement is unideal or if this is not what you had in mind!


bgyori:
Awesome, thank you!

kwalcock:
The rest of the world would probably put it in /doc, but I don't mind being different.  Giving people a little more time to veto before merging...

kwalcock:
It can always be moved, so I'll go ahead.

PR #456, opened 2018-10-04
DCT attachment

EgoLaparra:


kwalcock:
@EgoLaparra, can you check the description at https://github.com/clulab/eidos/wiki/JSON-LD for accuracy?  I'm probably going to change your code slightly, but it will be in kwalcock-timeTime.

EgoLaparra:
Yes, it is fine. And thanks for look over the code. :-)

PR #457, opened 2018-10-04
need to track this file

bsharpataz:


kwalcock:
While Eidos is running or something running Eidos is running.  Can someone get it just as well from GitHub, like the other files in /doc, I presume.

bsharpataz:
ok, per @MihaiSurdeanu i think this is good.  we can always move it later! :) thanks @kwalcock !

PR #459, opened 2018-10-04
Add DCT to eidos and to JLD

kwalcock:
This is still in the final stages of testing, but people can look.


kwalcock:
useTimeNorm is usually not true, so I'm inclined to be lenient and merge this.  We don't have a solution implemented to deal with reading the model from a jar.  For this large run, I've just got a hardcoded patch.

src/main/scala/org/clulab/wm/eidos/apps/ExtractJsonLDFromFile.scala
```diff
@@ -20,7 +20,7 @@ object ExtractJsonLDFromFile extends App {
   val pw = FileUtils.printWriterFromFile(outputFile)
   // 3. Extract causal mentions from the text
   val annotatedDocument = Timer.time("Just extraction") {
-    reader.extractFromText(text)
+    reader.extractFromText(text, documentCreationTime = Some("2018-09-07"))
```
MihaiSurdeanu: What is this arbitrary date?

kwalcock: Thanks.  Leftover from testing.  Removing...

kwalcock:
Further changes in master then.

EgoLaparra:
I have tried to run the last changes and everything works as expected.

PR #460, opened 2018-10-05
Hedge to resources

bsharpataz:
EidosSystem now owns a HypothesisHandler, and hedging is done in EidosSystem's main pipeline method.  As a consequence, the lexical terms are in a resource file for language specific
versions... fyi @gcgbarbosa @marcovzla


kwalcock:
Heads up then, @gcgbarbosa @marcovzla.  It is merged and you have English in your Portuguese.

PR #462, opened 2018-10-07
Tests

bsharpataz:
Felt like working on some tests/rules to see if any of them pass now... made a buncha changes, added some rules, I guess we'll see in next month's annotation effort if this improves things...


kwalcock:
Thanks for all the comments with the tests.

PR #464, opened 2018-10-07
added version of mapping with no indicators for interventions

bsharpataz:
per convo with Uncharted and Kimetrica -- basically they were saying that interventions inherently shouldn't be mapped to indicators, however, we still have the complete mapping file 

@adarshp fyi


adarshp:
Got it, thanks!

PR #466, opened 2018-10-11
Neg handler

bsharpataz:
Converted negationhandler to a class, currently not super lexical so I didn't make a resource file, though one could certainly be added.  However, the class is initialized with the language from the eidos.conf so lang specific implementations can (and should!) be implemented...

FYI @gcgbarbosa !

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/EidosActions.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosActions.scala
@@ -55,13 +55,11 @@ class EidosActions(val taxonomy: Taxonomy) extends Actions with LazyLogging {
     // In Kenya , the shortened length of the main growing season , due in part to a delayed onset of seasonal rainfall , coupled with long dry spells and below-average rainfall is resulting in below-average production prospects in large parts of the eastern , central , and southern Rift Valley .
     val modifiedMentions = assemble2 ++ nonCausal
 
-    // Basic coreference, hedging, and negation
+    // Basic coreference
     val afterResolving = basicDeterminerCoref(modifiedMentions, state)
-//    val afterHedging = HypothesisHandler.detectHypotheses(afterResolving, state)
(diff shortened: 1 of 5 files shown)


kwalcock:
That's a great idea passing the language to the NegationHandler.  Depending on how that develops, the companion object may want to take that language and create an EnglishNegationHandler or PortugueseNegationHandler, each derived from an AgnosticNegationHandler to take care of common tasks.  There is hopefully only one match statement necessary.

kwalcock:
The test framework has not kept up with the additional attachments and neither have the tests.  At some point don't we need to go through all the tests and next to INC, DEC, and QUANT add TIMEX, HEDGE, NEGATION, and PROP?  Third parties and planning might be involved.  I'd vote for merging this and then scheduling additional rounds.

bsharpataz:
> The test framework has not kept up with the additional attachments and neither have the tests. At some point don't we need to go through all the tests and next to INC, DEC, and QUANT add TIMEX, HEDGE, NEGATION, and PROP? Third parties and planning might be involved. I'd vote for merging this and then scheduling additional rounds.

I agree that we should merge these and then schedule add'l rounds, but I'll add that we may want to wait to make sure that we're happy with the representation (i.e., let the dust settle?)
As long as we're doing extra rounds we should also finish going thru failing tests and see if we can make more of them pass... I only did like 5 or 6 files so there is more to do

bsharpataz:
> That's a great idea passing the language to the NegationHandler. Depending on how that develops, the companion object may want to take that language and create an EnglishNegationHandler or PortugueseNegationHandler, each derived from an AgnosticNegationHandler to take care of common tasks. There is hopefully only one match statement necessary.

@kwalcock can i leave for future PRs when we see how much of the code in the English one is re-usable for the Portuguese (i.e. what can be shared vs language specific?)

kwalcock:
Yes, therefore the "Depending on how that develops"...

PR #468, opened 2018-10-12
Pt rules

gcgbarbosa:
@kwalcock Is that ok if in TestResources.scala we don't run utf8 checks on eidos.conf, reference.conf and wmseed.conf? We changed it to check only the references/english folder for unicode.


kwalcock:
That's a fine idea.  Those few files usually aren't a problem and now neither is Portuguese.

kwalcock:
I just wonder about this change to EidosActions which doesn't look specific to Portuguese.  Does the first regular expression overmatch something?

    "^cc".r, =>  "^cc$".r,

gcgbarbosa:
I guess "^cc".r overmaches ccomp. @bsharpataz is that right?

bsharpataz:
yep -- that was a good fix! for english too!

kwalcock:
Thanks all.  I'll press the button then.

PR #469, opened 2018-10-12
remove words in attachments from canonical form simple

bsharpataz:
And finally, on PR number 3 the branch name makes sense..... :)

@ajaynagesh  can you please check to see if this affects the grounding performance? (Ideally it will help!)

@kwalcock can you please wait for merging until we see how it affects grounding?

Thanks!

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/mentions/EidosMention.scala
+++ b/src/main/scala/org/clulab/wm/eidos/mentions/EidosMention.scala
@@ -5,7 +5,7 @@ import java.util.IdentityHashMap
 import org.clulab.odin._
 import org.clulab.wm.eidos.groundings._
 import org.clulab.struct.Interval
-import org.clulab.wm.eidos.attachments.EidosAttachment
+import org.clulab.wm.eidos.attachments.{ContextAttachment, EidosAttachment, TriggeredAttachment}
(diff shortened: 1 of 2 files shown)


bsharpataz:
@ajaynagesh did you get the chance to check this yet? thanks!

ajaynagesh:
@bsharpataz : Looking at it. When I tried to run `ExtractAndExport` I got some errors in the time normalization component (I tried running from IDE). So I am running from the commend line now .. will keep you posted ..

kwalcock:
Are these errors because the model file cannot be accessed?  That needs to be fixed in master still.  Normally time should not be configured to be active.

ajaynagesh:
@kwalcock : was able to fix .. but deleting the .idea file and reloading in IntelliJ

bsharpataz:
ok, then, @MihaiSurdeanu and @kwalcock let's go ahead an merge it, since only one grounding was made worse and I think this is a good way to go to prevent the properties from affecting the groundings... we can still revisit this later.  The next thing I plan to do is work on getting property groundings in.  @kwalcock , I plan to make use of the fact that diff ontologies can act differently, but please feel free to suggest alternate ways of doing things when you see the changes!

kwalcock:
OK.  I just hope that the different ontologies are stored in close to the same way.

PR #470, opened 2018-10-12
Fix bugs in time serialization, mention copy

kwalcock:
Also speed copy of mention with multiple attachments

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/utils/MentionUtils.scala
+++ b/src/main/scala/org/clulab/wm/eidos/utils/MentionUtils.scala
@@ -1,0 +1,20 @@
+package org.clulab.wm.eidos.utils
+
+import org.clulab.odin._
+
+object MentionUtils {
+
+  def newCrossSentenceMention(m: CrossSentenceMention, attachments: Set[Attachment]) =
+      new CrossSentenceMention(m.labels, m.anchor, m.neighbor, m.arguments, m.document, m.keep, m.foundBy, attachments)
+
+  def withAttachments(mention: Mention, seqOfAttachments: Seq[Attachment]): Mention = {
+    val setOfAttachments = seqOfAttachments.toSet
+
+    mention match {
+      case m: TextBoundMention => m.copy(attachments = setOfAttachments)
+      case m: RelationMention => m.copy(attachments = setOfAttachments)
+      case m: EventMention => m.copy(attachments = setOfAttachments)
+      case m: CrossSentenceMention => newCrossSentenceMention(m, setOfAttachments)
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


src/main/scala/org/clulab/wm/eidos/utils/MentionUtils.scala
```diff
@@ -0,0 +1,20 @@
...
+object MentionUtils {
+
+  def newCrossSentenceMention(m: CrossSentenceMention, attachments: Set[Attachment]) =
+      new CrossSentenceMention(m.labels, m.anchor, m.neighbor, m.arguments, m.document, m.keep, m.foundBy, attachments)
+
+  def withAttachments(mention: Mention, seqOfAttachments: Seq[Attachment]): Mention = {
+    val setOfAttachments = seqOfAttachments.toSet
+
+    mention match {
+      case m: TextBoundMention => m.copy(attachments = setOfAttachments)
```
bsharpataz: @kwalcock  I think we might need this to be `m.copy(attachments = m.attachments ++ setOfAttachments)` or something similar (or added up in line 11).  I think as it it would overwrite any attachments that were previously there.  I don't think it is "critical" here because hedging is only done on event mentions and they don't usually (so far ever?) have attachments before hedging, but if we want it to be more general purpose then I think the change needs to be made...? There's a chance I'm wrong but I think this is the way this would work

bsharpataz:
@kwalcock left a comment on the code... not sure how github lets you know :)

src/main/scala/org/clulab/wm/eidos/utils/MentionUtils.scala
bsharpataz: and for the other lines too...

bsharpataz:
actually, idk... i thought that was the way it worked (and please check!) but i check out the branch and played and things _seem_ ok...? idk... anyway, I am PRing a couple more places where your general fix should be used to your branch... (hope that's the way we do things)

src/main/scala/org/clulab/wm/eidos/utils/MentionUtils.scala
kwalcock: Uh oh.  I think you're right and I remember thinking about the ++ but don't see it anywhere.  I'll change and test.  Thanks for the keen powers of observation.

kwalcock:
I've done it manually, one new attachment at a time, once before and need to track that down as well.

kwalcock:
For the record, see
https://github.com/clulab/processors/issues/280
https://github.com/clulab/processors/issues/277
If/when these are added to processors, the eidos code can be simplified.

kwalcock:
The bug(fix) made no change to the output of the 52 documents.  I'm spot checking the larger collection.

kwalcock:
No changes in over 5000 documents.  Maybe I got lucky.

bsharpataz:
I think the only place it was in use was for hedging, and those only apply
to event mentions (which don't get the other attachments, as the system
stands).  So, I think that it was fine, but risky for the future and not
safe for other uses (see PR I submitted to your branch)

kwalcock:
@bsharpataz you are tagged again.  Thanks for finding that error!

PR #471, opened 2018-10-13
more places

bsharpataz:


kwalcock:
@bsharpataz, it looks like the behavior of EidosActions.mergeAttachments has changed.  The former copyWithAttachments starting by first removing all the attachments that were present and then adding the new ones.  The removing part, val attachmentless, needs to stay (although I should write a withoutAttachment_s_ function to replace it).  The second part...  Too many words.  I'm just going to do it and tag you to inspect.  This has turned out to be harder than it looked.

bsharpataz:
ah, ok, but... if it's a set, why are we removing?

kwalcock:
I think the idea was to replace the existing attachments with better ones.

PR #472, opened 2018-10-13
Fix compilation problem

kwalcock:


bsharpataz:
Haha, we’re the BEST with names, _right?_ thanks @kwalcock !

PR #473, opened 2018-10-13
Fix mergeAttachments

kwalcock:


kwalcock:
These will be slightly altered for final merge.

PR #483, opened 2018-10-19
Enable multi-lingual testing

kwalcock:


kwalcock:
TestUtils now defines various kinds of tests
- Test (should not use EidosSystem, eventually abstract)
  - ContraptionTest (should not use EidosSystem, concrete)
  - ExtractionTest (uses EidosSystem, defaults to English for now)
    - EnglishTest (uses EidosSystem in English)
    - PortugueseTest (uses EidosSystem in Portuguese)

Default settings for use in tests is in englishTest.conf and portugueseTest.conf.  eidos.conf is for production, not for testing now.

In order to prevent overabundance of EidosSystem instances, the most recent one is reused if the configuration matches  build.sbt groups tests so that the configurations are more likely to match.  This is based on package names, so don't just put tests anywhere.  Some degree of parallelism is lost and I'm trying to improve that.

kwalcock:
Thanks, @gcgbarbosa.  That's probably not the last problem you'll find.

PR #484, opened 2018-10-23
Documented how to submit a `request` to the Eidos web service

brandomr:
@kwalcock here is some documentation around how I was able to interact with the web service. 

That said, I uncovered this functionality while trying to invoke Eidos via the INDRA (see the [`process_text`](https://github.com/sorgerlab/indra/blob/master/indra/sources/eidos/api.py#L22-L60)) API for Eidos. It has the option to use Eidos as a webservice, instead of as a JAR. The problem is that INDRA is looking for a `process_text` [endpoint](https://github.com/sorgerlab/indra/blob/master/indra/sources/eidos/api.py#L54-L55) from the Eidos webservice which as far as I can see does not exist. 

So, I assumed that it should point to the `parseText` endpoint. However, that endpoint returns JSON (including the HTML which is ultimately rendered by the webapp) whereas INDRA expects JSON-LD.  

Ideally @jgawrilo and I can wrap Eidos in a Docker container so that it can be used via INDRA, through the web service, with minimal assembly/build. 

Welcome your thoughts as it is totally possible I'm missing some details around the Eidos/INDRA web service integration!

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,8 @@ correlation, `same-as`, and `is-a` relations.
       - [Export to JSON](#to-serialize-to-json)
     - [Command Line Usage](#command-line-usage)
       - [Interactive shell](#running-an-interactive-shell)
-      - [Webapp](#running-the-webapp)
+    - [Webapp](#running-the-webapp)
+      - [Interacting with the web service](#interacting-with-the-web-service) 
   - [How to use Eidos output](#how-to-use-eidos-output)
     - [Visualizing Eidos output](#visualizing-eidos-output)
     - [Using Eidos output for modeling](#using-eidos-output-for-modeling)
@@ -186,7 +187,7 @@ for testing the output of Eidos. To run it, do
 ./shell
 ```
 
-#### Running the webapp
+### Running the webapp
 
 To run the webapp version of EidosShell locally, do:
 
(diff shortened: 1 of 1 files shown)


bgyori:
Just to clarify, I implemented a Python-based web service around Eidos as part of INDRA (`indra.sources.eidos.server`) which invokes the Python API to Eidos in `indra.sources.eidos.api` rather than Eidos directly, hence the use of `process_text` which is a valid API function in INDRA.

bgyori:
Yes, I think a native Eidos web service that exposes a `parseText` endpoint with text input and JSON-LD output would be preferable (as far as I understand, such an endpoint doesn't yet exist, right?). Then you could set up a Docker image without having to include Python and INDRA, and one could use Eidos reading via INDRA just by pointing to a local or a remote instance of the running service.

kwalcock:
Thanks, @brandomr, for the ideas and extra effort to make the PR.  It has given us some things to think about.

@bgyori, I was thinking about making an optional fat jar download that would require only Java so that the Scala API might be used more directly, without sbt, maven, docker, web servers, etc.  It might cut down on the documentation I have to write, most importantly.  Would a single eidos.jar be useful?  It might also be possible to streamline the web service or apps so that
java -jar eidos.jar org.clulab.etc.webapp
or
java -jar eidos.jar org.clulab.etc.processThis "Food insecurity causes increased migration."
produce useful output.

bgyori:
Yes, a fat JAR is also useful - I don't think it's necessarily better than the web service / Docker approach but it's complementary and there will be use cases where it's more convenient!

kwalcock:
I'll look into adding `parseText` when there is time.

bgyori:
One drawback of the JAR CLI approach, I think, is that each reading will reload the system and its resource files whereas a running service (local or remote) just does that one time when initializing.

kwalcock:
Yes, indeed.  Does having single jar make using the API easier from Python?  IIRC you are using IPython and not Jython and that there is a default constructor somewhere that allows you to create an EidosSystem.  Is the reason for the web interface so that multiple machines can be involved?  That sounds very useful.

bgyori:
Yes, here is the Python wrapper around an Eidos fat JAR that I implemented: https://github.com/sorgerlab/indra/blob/master/indra/sources/eidos/reader.py
The prerequisite is to build a JAR by running `sbt assembly` and then point the env variable EIDOSPATH to the JAR file.
This instantiates a singleton EidosSystem object and then calls its extractFromText method to run reading. The Python-Scala bridge is implemented using a package called pyjnius.

MihaiSurdeanu:
I agree that a fat jar with a simple `parseText` API would be nice to have.

PR #489, opened 2018-10-25
Pt neg+tets

gcgbarbosa:
@kwalcock Is there a problem if I leave the tests commented until the next version of processors is released?

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/attachments/NegationHandler.scala
+++ b/src/main/scala/org/clulab/wm/eidos/attachments/NegationHandler.scala
@@ -20,10 +20,58 @@ class NegationHandler(val language: String) {
     // Note that the approach can be different for different languages!
     language match {
       case "english" => mentions.map(detectNegationEnglish)
-      case "portuguese" => ???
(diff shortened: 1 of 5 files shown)


kwalcock:
I don't especially mind comments.  It might be easier to not use passingTest and instead mark them with waitingForProcessors or something.  That's easier to search and replace for and it also means that syntax checking is still active as well as refactoring support.  Things would be less likely to change out from under you.

kwalcock:
I'm not sure about `class NegationHandler(val language: String)` in the long run.  It might turn into something like
```
object NegationHandler {
  def apply(language: String) = 
    if (language == 'portuguese') new PortugueseNegationHandler else new EnglishNegationHandler
```
fairly soon.  Do you have an idea how many classes need to be customized in similar ways?  I wonder how much needs to be coordinated with Odin if that strategy is used.

kwalcock:
In case you don't know, the convention is that the branch gets deleted from the server after it is merged.  I'll wait, though, in case there isn't a local branch and because this one will probably be used shortly when processors is updated.

gcgbarbosa:
Thanks for letting me know.

Issue #486, opened 2018-10-24
DL4J error when running with fat jar


bgyori:
When running extractFromDirectory via an assembled fat jar, I'm getting this error:
```
18:12:13.419 [main] WARN  o.d.n.m.keras.utils.KerasModelUtils - Could not read keras backend used (no backend field found) 

18:12:13.448 [main] WARN  o.d.n.m.k.l.e.KerasEmbedding - Masking in keras and DL4J work differently. We do not completely support mask_zero flag on Embedding layers. Zero Masking for the Embedding layer only works with unidirectional LSTM for now. If you want to have this behaviour for your imported model in DL4J, apply masking as a pre-processing step to your input.See https://deeplearning4j.org/usingrnns#masking for more on this.
Exception in thread "main" java.lang.ExceptionInInitializerError
	at org.deeplearning4j.nn.conf.dropout.Dropout.initializeHelper(Dropout.java:107)
	at org.deeplearning4j.nn.conf.dropout.Dropout.<init>(Dropout.java:100)
	at org.deeplearning4j.nn.conf.dropout.Dropout.<init>(Dropout.java:80)
	at org.deeplearning4j.nn.conf.layers.Layer$Builder.dropOut(Layer.java:257)
	at org.deeplearning4j.nn.modelimport.keras.layers.embeddings.KerasEmbedding.<init>(KerasEmbedding.java:125)
	at org.deeplearning4j.nn.modelimport.keras.utils.KerasLayerUtils.getKerasLayerFromConfig(KerasLayerUtils.java:258)
	at org.deeplearning4j.nn.modelimport.keras.KerasModel.prepareLayers(KerasModel.java:220)
	at org.deeplearning4j.nn.modelimport.keras.KerasModel.<init>(KerasModel.java:166)
	at org.deeplearning4j.nn.modelimport.keras.KerasModel.<init>(KerasModel.java:98)
	at org.deeplearning4j.nn.modelimport.keras.utils.KerasModelBuilder.buildModel(KerasModelBuilder.java:305)
	at org.deeplearning4j.nn.modelimport.keras.KerasModelImport.importKerasModelAndWeights(KerasModelImport.java:127)
	at org.clulab.timenorm.TemporalCharbasedParser.<init>(TemporalCharbasedParser.scala:56)
	at org.clulab.wm.eidos.EidosSystem$LoadableAttributes$.apply(EidosSystem.scala:164)
[...]

PR #490, opened 2018-10-26
Do not discard nd4j services

EgoLaparra:
This solves the problem with the assembled fat jar.


bgyori:
Thanks, I tested the build from this branch and it works so this indeed fixes #486!

PR #493, opened 2018-10-29
PT hedging+transparent+stop

gcgbarbosa:
Fixed a problem with the negation, translated filtering files and updated rules.


kwalcock:
I just wonder about `(depNegations).toSet` which must be an artifact left over from previous code.

gcgbarbosa:
Thanks for pointing it out. I have cleaned it out.

kwalcock:
OK.  I don't want this waiting here to slow you down, so I'll merge.

kwalcock:
I see that this is a different branch than the previous pt-neg+tets.  It looks like next time I should maybe resume deleting the branch after merging.  Would you go ahead and delete remote branches pt-* if you're done with them and/or have local copies?  Thanks, @gcgbarbosa.

gcgbarbosa:
Thanks @kwalcock. I deleted them, except for pt-tests because I am using it to synchronise the version of eidos I have in my machine and the version on the servers if i need to run something there. Idk if this is ok or if there is a better way to do that.

kwalcock:
That's great.  I have a branch to do exactly the same.

gcgbarbosa:
Awesome! Thanks.

PR #496, opened 2018-10-30
Web service endpoint for web application

brandomr:
This PR includes service enabling the web application with the creation of a `/process_text` endpoint. After the web application is run, a user may `POST` a request with JSON to the `/process_text` endpoint. The user will be returned an Eidos processed document in JSON-LD format.

I included additional docs/examples in the README as well as a test. I corrected the webapp tests so that they function when invoked with `sbt webapp/test`.

I attempted to follow your design paradigms as best as I could, and intend this mostly to open discussion around this capability and hopefully to see it included in Eidos if you think it is useful and valid. 

The use case for this is to be able to provide a ready made Docker container which initializes the web application upon start so that users (e.g. data scientists) who may only work in Python/R/etc. and are unfamiliar with Java/Scala could easily deploy Eidos. As INDRA and Delphi are Python this reduces friction for someone who wants to run INDRA + Delphi + Eidos.

To manually test this PR you should run (after cloning this fork) :
```
cd eidos
git checkout webservice
sbt webapp/run
```
Then you can submit a request with:
```
curl \
  --header "Content-type: application/json" \
  --request POST \
  --data '{"text": "Drought increases regional insecurity."}' \
  http://localhost:9000/process_text
```

To run the automated tests:
```
cd eidos
git checkout webservice
sbt webapp/test
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ page.
 - Running Eidos
   - Using the Sample Apps
     - [Webapp](#webapp)
+    - [Web Service](#web-service)
     - [Interactive Shell](#interactive-shell)
     - [ExtractAndExport](#extractandexport)
     - [Other Apps](#other-apps)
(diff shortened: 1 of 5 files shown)


MihaiSurdeanu:
@kwalcock: can you please review this PR? Thanks!

brandomr:
@kwalcock thank you for taking a look! I added a type conversion which I believe resolves this warning.

kwalcock:
I wonder what interesting visualizations might become of this.  Thanks, @brandomr.

PR #497, opened 2018-10-30
Update the geoNorm code

kwalcock:
Some changes had to be made and then one thing led to another...  Please look this over carefully.  The original code would not produce enough output for me to run regression tests.  So, this new code may contain mistakes.  It is not working completely, either.  See the TODOs and recent discussion.


src/main/scala/org/clulab/wm/eidos/context/GeoPhraseID.scala
```diff
@@ -0,0 +1,5 @@
+package org.clulab.wm.eidos.context
+
+case class GeoPhraseID(phraseID: String, PhraseGeoID: Option[Int], StartOffset_locs: Int, EndOffset_locs: Int)
```
EgoLaparra: @vikas95, the names of this class and its variables are a bit confusing.  Could you change them after reviewing and merging Keith's changes.

```suggestion
case class GeoLocation(text: String, geonameID: Option[Int], startOffset: Int, endOffset: Int)
```

kwalcock:
I just updated the JSON-LD page and think that

`geoids | [GeoidPhrase] | and a list of geographic identifier phrases.`

should change to something like

`geolocs|[GeoLocation]|and a list of geographic location records.`

However, there is probably some standard nomenclature I am unaware of.

EgoLaparra:
I think the second option describes better the content of that element. @vikas95, what do you think?

src/main/scala/org/clulab/wm/eidos/context/GeoPhraseID.scala
vikas95: > @vikas95, the names of this class and its variables are a bit confusing. Could you change them after reviewing and merging Keith's changes.
> 
> Suggested change Beta 
> [Give us feedback](https://www.research.net/r/SuggestedChanges)
> case class GeoPhraseID(phraseID: String, PhraseGeoID: Option[Int], StartOffset_locs: Int, EndOffset_locs: Int)
> case class GeoLocation(text: String, geonameID: Option[Int], startOffset: Int, endOffset: Int)

Yes, I think I have to change few more variable names, I'll work on these.

kwalcock:
Just nagging.  I'm hoping that this can be figured out today.  Please let me know if I can help by doing something other than keeping my hands off.  Thanks.

vikas95:
> I think the second option describes better the content of that element. @vikas95, what do you think?

Yes, the names mentioned in the second option look better.

kwalcock:
I updated the JSON-LD texts and the matching documentation on the Wiki.  The code will hopefully be merged into geoNorm and then into master.

kwalcock:
@vikas95, I'm merging my code into branch geoNorm then.  There will still be some changes from me plus I think at least some name changes coming from you.  Can we treat geoNorm as the current dev branch and each of us branch and merge into that before the whole thing goes into master?  So
```
kwalcock-geoNorm ->
                    > geoNorm -> master
 vikas95-geoNorm ->
```

PR #506, opened 2018-11-06
Sorted to expand closest args first

bsharpataz:
... had to refactor to make it work...

should help with sentences where both args are on the same side of the trigger and the further away arg swallows the closer, as with the sentence:
```Areas of greatest concern include Wau , northwestern Jonglei , and central Unity , though given the severity of food insecurity across South Sudan and unpredictable nature of conflict , the possibility of Famine ( IPC Phase 5 ) should be considered in many states of the country .```

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/actions/ExpansionHandler.scala
+++ b/src/main/scala/org/clulab/wm/eidos/actions/ExpansionHandler.scala
@@ -326,11 +326,6 @@ class ExpansionHandler(val language: String) extends LazyLogging {
       m
   }
 
-  def addOverlappingAttachmentsTextBounds(ms: Seq[Mention], state: State): Seq[Mention] = {
-    for {
-      m <- ms
-    } yield addOverlappingAttachmentsTextBounds(m, state)
-  }
   def addOverlappingAttachmentsTextBounds(m: Mention, state: State): Mention = {
     m match {
       case tb: TextBoundMention =>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by kwalcock** [comment]
I'm pretty sure this change is a good idea, but at some point I'm going to try to review anything with O(x) > O(n), like sorting and distinct.  I don't think that Scala keeps track of the sorted by values (distToTrigger(trigger, _)) and makes possibly expensive calculations multiple times.  (These are pretty simple ones, though.)

The RuntimeException is scary.
```
    for {
      m <- ms
    } yield addOverlappingAttachmentsTextBounds(m, state)
```
 =>
```
ms.map(addOverlappingAttachmentsTextBounds(_, state))
```
or something.

I'm not sure about the math.abs.  In one case 2nd case it isn't necessary.  I'm not sure about the first with the start/end/abs and comparison of intervals.  If it works I can't complain.

Can m.arguments be mappend onto an immutable HashMap?  It can wait until a different pass if the code sticks around long enough and if it's even possible.

**Comment by bsharpataz** [comment]
re: addOverlapping -- > I can do you one better, removed in a push I'm about to do! (it was kinda leftover from previous code)

re: RuntimeException -- huh? where is it?

re: math.abs -- you're right in that it should be fine without it......... I was trying to be safe, but perhaps I should assume and catch issues that fail my expectations?

re: m.arguments -- what line?

**Comment by kwalcock** [comment]
You tricked me!  Distracting me with unused code :-(  you were.  The m.arguments is between

```
    val newArgs = scala.collection.mutable.HashMap[String, Seq[Mention]]()
    for ((argType, argMentions) <- m.arguments) {
...
      newArgs.put(argType, attached)
```
It seems like it could be m.arguments.map{ ...; (argType -> attached) }.toMap.  I'm not sure which is more efficient, but one is prettier.

kwalcock:
However, I think @bsharpataz should ignore tiny things like this.  Do you need Mihai to look at it?  I never know if multiple reviewers is an and or an or and consequently wait to long for the and.

bsharpataz:
I don't think I need him to, and he's kinda busy at the moment, so whatever
you think is best

PR #507, opened 2018-11-09
Review use of null, var, and fix sorting

kwalcock:
This does include changes to the cached ontologies, so they need to be redone.  Null is no longer used as a sentinel value there.  Regression testing showed unstable output.  PropertiesOntologyGrounder.groundOntology needed to sort items coming from a set.


kwalcock:
I'm tagging @bsharpataz to keep me honest and sane.

Issue #315, opened 2018-05-17
Namespace for JSON-LD


kwalcock:
This is being moved here to separated it from other JSON-LD issues.  More will appear soon.

kwalcock:
So, I think this means that we can use the URL (below) on the left in the JSON-LD and it will map to the one on the right which will take the curious to our specification.  If our wiki page moves or changes format, the mapping can be changed to match.  @dgarijo, can you make a rewrite rule for .htaccess to do this mapping?  If so, we have about 15 of them just now and they will change as we update the format.  I'll make a list.

https://w3id.org/wm/cag/Sentence => https://github.com/clulab/eidos/wiki/JSON-LD#user-content-Sentence

kwalcock:
Cool.  The present list then continues with the same pattern for these items: Corpus, Document, Word, Dependency, Entity, DirectedRelation, UndirectedRelation, Groundings, Grounding, Provenance, Interval, State, Modifier, and Trigger.  If there is a way to match them generically, it could save a lot of work.  Something like ^cag/([^/]+)$ https://github.com/clulab/eidos/wiki/JSON-LD#user-content-$1 would be in that direction.  I didn't check the details, though.

dgarijo:
Done: e.g., https://w3id.org/wm/cag/Grounding

kwalcock:
Great.  Thanks for sorting out the details.

dgarijo:
no problem

Issue #278, opened 2018-04-10
Extension of JSON-LD format to accommodate BBN data and our own future work


kwalcock:
Disclaimer: There are many ways to do this and picking one will likely involve unknown forces of subjectivity, intuition, and conjecture that won't match mine.  In other words, this may be totally off base.

I'd like to divide the participants in our (binary) DirectedRelation and (binary) UndirectedRelation into essential and non-essential participants and not mix them.  This is so that the focus is clear and simple and not everyone is forced to deal with the non-essential.  This comes at the expense of needing different code to process the two types.

In the case of DirectedRelation, we have to somehow specify the two (sets of) participants and the direction.  Without these there just isn't a DirectedRelation.  These should be placed in unambiguous locations and not require extra reasoning to extract.  We have something called source and destination to fulfill these requirements.

The situation is then complicated by the fact that there may be different kinds of DirectedRelation.  We've been focussed on Causal, but there are more in our own pipeline and other groups will have theirs.  Right now we have some indication of the particular relationship in the labels field and less directly, a rule field.  The former contains a list for us and it is only by convention that someone would know that the first item is most specific.  It doesn't seem likely that all groups will have a list or a rule field.  I suggest a new field called relationName that specifies in a single string fit for human consumption the kind of DirectedRelation.

Each kind relation may have special names for the participants that will help a human understand what is meant by the generic source and destination.  These might be used to label nodes even when the relationName is not recognized.  I'd call these sourceName and destinationName.

It is possible to store this information in many different ways, such as
participants: [ { role: "source", name: "cause", value: "lack of food" }, { role: "destination", name: "effect", value: "hunger" } ]
and I will suggest that for the non-essential participants that have more open class roles.  For the essential participants I think this complicates extraction of information such as the name of the destination.  Finding the answer requires a search through a list.

The non-essential participants, which may be thought of as modifying the main relationship in any of a myriad of ways (and drawn as such) need a field name.  Modifier and argument are already in use once, attachment is used in the program, but not JSON-LD.  A brainstorm of other names is pasted below.  I'd like to require that these non-essential participants be other instances of Entity, DirectedRelation, or UndirectedRelation and be included by reference.  They can be in charge of their own rules, triggers, labels, provenance, etc.  Other fields can be "role" for a formal, potentially standardized explanation of what the non-essential is doing there, and then "name" for a human friendly version, and lastly, "value".  

Here is a simplified example:

She gave him a book yesterday at the library.  (This is an increase event.)

{
[...]

PR #508, opened 2018-11-09
Fix the jsonld context values

kwalcock:


kwalcock:
This somehow got away from me several months ago after there were temporarily two jsonld formats and we standardized on this one which happened to have the old base value.

kwalcock:
I hope my memory is accurate.  This was the subject of issues #278 way at the bottom and all of #315.  At one time it was a good idea and I think it just got lost in the format change.  Perhaps there were other issues as well, and therefore the double check before merge.

PR #509, opened 2018-11-10
Make sure everything closes automatically, even in the case of exceptions

kwalcock:
This is probably excessive, but I just processed an excessive number of files and experienced quite a few unexplained problems.  sbt would quit, Eidos would crash with heap corruption, files that probably should have been processed did not appear in the output.  Excessive carefulness may be justified (and this will extend to nulls, vars, and mutable collections).  Prior to processing I did find some files not being closed and although it was not a significant problem, I'd like to grep for close() and be able to tell the from old (countable on a couple fingers) from the new.  Java has a standardized try with resources and one would normally never see a close().  This is making up for not using the convention from the start.  I wasn't satisfied with any other implementation, so made my own.  Replacements are welcome.  I would like to add some implicit conversions so that one could write (new Closeable()).autoClose { } instead of autoClose(new Closeable()) { } but maybe next time.


bsharpataz:
Fwiw, @marcovzla was _just_ showing me something similar he wrote elsewhere (I think), IDK if this bit of info is helpful, but thought I’d “bring you together” (through github @s ??) just in case! You guys rock!

kwalcock:
Marco has a good implementation in org.clulab.utils.Serializer which I consulted.  For some time we couldn't use it because the whole of processors wasn't released with it.  I also wanted something independent of the Serializer.  The particular closer isn't too worrisome to me.  Spreading its use throughout the code is more of an issue.  I hope people don't mind their work being revised or standards being instituted.

kwalcock:
In my experience mutables are seldom a problem, especially if they are temporary variables.  Their immutable counterparts are worth Scala bonus points, though, and I like having one less thing to worry about.  Seldom != never and it's sometimes easier just to remove all doubt (IMO).

kwalcock:
Merge passes regression test as well

PR #510, opened 2018-11-12
Touch up geolocation and timex

kwalcock:


kwalcock:
@vikas95, @EgoLaparra what's the reason for using both B-Time and I-Time in one case and a simple LOC in the other case rather than B-LOC and I-LOC?

kwalcock:
@EgoLaparra, your code was changed some as well, so you are being tagged.  A lot of regression testing was done, so I don't *think* that the output has changed, but I can be wrong, obviously.

vikas95:
> @vikas95, @EgoLaparra what's the reason for using both B-Time and I-Time in one case and a simple LOC in the other case rather than B-LOC and I-LOC?

@kwalcock - thanks for pointing this, Ideally it should be B-LOC and I-LOC but currently I was updating norms with just "LOC" and therefore "B-LOC" and "I-LOC" are not visible. I will try to fix this.

kwalcock:
There is at least one place where the output has been changed.  If the original text includes a _ character, it is no longer converted to a space for the geolocation.  This resulted in one discrepency in the regression test.

vikas95:
> There is at least one place where the output has been changed. If the original text includes a _ character, it is no longer converted to a space for the geolocation. This resulted in one discrepency in the regression test.

@kwalcock - By original text, do you mean a location phrase has "_" in the raw input text ?? 
I had selected "_" as a delimiter between two tokens of a location phrase because I was assuming that there is no location phrase with "_" in it. Can you share an example ??

kwalcock:
Better not to assume, especially if it isn't necessary.

```
      "geolocs" : [ {
        "@type" : "GeoLocation",
        "@id" : "_:GeoLocation_223",
        "startOffset" : 88421,
        "endOffset" : 88540,
        "text" : "http://blogs.worldbank.org/voices/givingvoice-poor-adding-human-touch-poverty-data-southsudan?CID=POV_TT_Poverty_EN_EXT"
      } ]
```

vikas95:
> Better not to assume, especially if it isn't necessary.
> 
> ```
>       "geolocs" : [ {
>         "@type" : "GeoLocation",
>         "@id" : "_:GeoLocation_223",
>         "startOffset" : 88421,
>         "endOffset" : 88540,
>         "text" : "http://blogs.worldbank.org/voices/givingvoice-poor-adding-human-touch-poverty-data-southsudan?CID=POV_TT_Poverty_EN_EXT"
>       } ]
> ```

Thanks for finding this example, it is weird that weblink is tagged as a location. Is this the only outlier case or we have other examples too ??

kwalcock:
Not yet.  It's fixed anyway, though.

vikas95:
> Not yet. It's fixed anyway, though.

Thanks, "@kwalcock is the best" - which is also a very common quote here on GitHub  :-D

kwalcock:
Here goes.  Cached ontologies will need to be rebuilt after these changes.

PR #514, opened 2018-11-19
added document filter

bsharpataz:
and some comments for good measure...

currently only filtering by length and after tokenization, before parsing, but that can change.

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
@@ -49,6 +49,9 @@ class EidosSystem(val config: Config = ConfigFactory.load("eidos")) extends Stop
     }
   }
 
+  // Prunes sentences form the Documents to reduce noise/allow reasonable processing time
+  val documentFilter = FilterByLength(proc, cutoff = 100)
+
   val debug = true // Allow external control with var if needed
 
   println("Loading W2V...")
@@ -249,15 +252,21 @@ class EidosSystem(val config: Config = ConfigFactory.load("eidos")) extends Stop
 
   // Annotate the text using a Processor and then populate lexicon labels
   def annotate(text: String, keepText: Boolean = true, documentCreationTime: Option[String] = None, filename: Option[String]= None): Document = {
-    val oldDoc = proc.annotate(text, true) // Formerly keepText, must now be true
-    val doc = EidosDocument(oldDoc, keepText)
+    // Syntactic pre-processing
+    val tokenized = proc.mkDocument(text, keepText = true)  // Formerly keepText, must now be true
+    val filtered = documentFilter.filter(tokenized)         // Filter noise from document
+    val annotated = proc.annotate(filtered)
+    val doc = EidosDocument(annotated, keepText)
+    // Add the tags from the lexicons we load
     doc.sentences.foreach(addLexiconNER)
+    // Time and Location
     doc.parseDCT(loadableAttributes.timenorm, documentCreationTime)
     doc.parseTime(loadableAttributes.timenorm)
     doc.parseGeoNorm(loadableAttributes.geonorm)
-
+    // Document ID
     doc.id = filename
     doc
+
   }
 
   protected def addLexiconNER(s: Sentence) = {
--- a/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
@@ -50,7 +50,7 @@ class EidosSystem(val config: Config = ConfigFactory.load("eidos")) extends Stop
   }
 
   // Prunes sentences form the Documents to reduce noise/allow reasonable processing time
-  val documentFilter = FilterByLength(proc, cutoff = 100)
+  val documentFilter = FilterByLength(proc, cutoff = 300)
 
   val debug = true // Allow external control with var if needed
 
--- a/src/main/scala/org/clulab/wm/eidos/utils/DocumentFilter.scala
+++ b/src/main/scala/org/clulab/wm/eidos/utils/DocumentFilter.scala
@@ -1,0 +1,32 @@
+package org.clulab.wm.eidos.utils
+
+import org.clulab.processors.corenlp.CoreNLPDocument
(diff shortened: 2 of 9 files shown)


kwalcock:
I'm worried about this.  
```
def getSentenceText:String =  getSentenceFragmentText(0, words.length)

def getSentenceFragmentText(start:Int, end:Int):String = {
...
      val numberOfSpaces = math.max(1, startOffsets(i) - endOffsets(i - 1))
```
forces a space into words like `it's`.  When it gets reparsed from the text, what happens?  What happens to provenance?  I wish that parsed sentences could be reused or that the simple, raw text was available.

bsharpataz:
@kwalcock - I agree with that, but I don't know a work-around with the methods in processors.  I will say that i had to adjust spacing on a bunch of unit tests, bc indeed it affects the spaces..... @MihaiSurdeanu ?

MihaiSurdeanu:
Note that newer processors preserve the input text. That is, each token has a "raw" form (which is exactly the corresponding input text), as well as startOffsets and endOffsets, which are character offsets in the input text, from which the necessary number of spaces can be inferred:

https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/processors/Sentence.scala#L15

kwalcock:
Eidos does have access to the original text, so one could just slice it from startOffsets[0] to endOffsets[last].  We're lucky that
```  def getSentenceFragmentText(start:Int, end:Int):String = {
...
      text.append(raw(i))
```
rather than `text.append(words(i))`.

kwalcock:
However, the constructor for (a processor) Document takes an array of sentences.  Can we just take the sentences of the original document, filter them like you do, and use them to make a new document with some missing sentences?  Maybe the sentences don't need to be processed twice then.

bsharpataz:
@kwalcock sorry -- I'm confused, are we concluding that so I should take the Array[Sentence] that survives the prune, then map each of these to a new string using the Sentence.raw tokens along with: https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/processors/Processor.scala#L21
?

bsharpataz:
> However, the constructor for (a processor) Document takes an array of sentences. Can we just take the sentences of the original document, filter them like you do, and use them to make a new document with some missing sentences? Maybe the sentences don't need to be processed twice then.

no-- that's where this all began, recall the Issue I started on processors.... I would love for that functionality to exist though, I think this is the 2nd time I've tried to do this in the past year or so

kwalcock:
An experiment will set me straight.  Results forthcoming...

kwalcock:
It might look like this, which you probably already tried, but just in case:
```scala
  def annotate(text: String, keepText: Boolean = true, documentCreationTime: Option[String] = None, filename: Option[String]= None): Document = {
    val oldDoc = proc.annotate(text, true) // Formerly keepText, must now be true
    val newSentences = oldDoc.sentences // .filter(sentence => sentence.words.size < 100)
    val newDoc = Document(newSentences)
    newDoc.id = oldDoc.id
    newDoc.coreferenceChains = oldDoc.coreferenceChains // do these need to be filtered?
    newDoc.discourseTree = oldDoc.discourseTree // do these need to be filtered?
    newDoc.text = oldDoc.text
    val doc = EidosDocument(newDoc, keepText)
```

bsharpataz:
isn't the problem `val oldDoc = proc.annotate(text, true)` ?
I'm filtering bc the sentences take too long to process, right?

kwalcock:
Duh, yes indeed.  What was I thinking?  This is why you are tagged on so many PRs!

kwalcock:
I *believe* to understand what you were battling now.  The ShallowNLPProcessor doesn't return a plain Document, but a CoreNLPDocument.  Only the plain Document object understands how to recreate a plain Document with an edited list of sentences.  ShallowNLPProcessor has a hack in basicSanityCheck that checks for a CoreNLPDocument for dealing with annotations and this is why a plain Document cannot substitute.  I think the proper solution would be to add a method to class Document (and CoreNLPDocument) that is ~copyWithSentences so that they do the right thing for us.  In the meantime, I suggest countering the one hack temporarily with another such as
```scala
    val newDoc: Document = proc.mkDocument(text, keepText = true)
    val newerSentences = newDoc.sentences // .filter(sentence => sentence.words.size < 100)
    val newerDoc = Document(newDoc.id, newerSentences, newDoc.coreferenceChains, newDoc.discourseTree, newDoc.text)
    val newestDoc = // This is a hack for lack of constructor for CoreNLPDocument
        if (newDoc.isInstanceOf[CoreNLPDocument])
          ShallowNLPProcessor.cluDocToCoreDoc(newerDoc, true)
        else
          newerDoc
    proc.annotate(newestDoc)
    val doc = EidosDocument(newestDoc, keepText)
```
This should preserve the original text (my major concern), offsets, and provenance.  I *think* it just means that some bad sentences will not be included in the output as if they didn't exist.  (I'm going to test on those files soon, in part to see if the limit should be 100, 200, or what.)  All of the tests pass unchanged.  Your filter seems like an idea to keep (and some code above might be moved internal to it).  I can (probably) do a PR to processors with updated code that can be used from eidos after processors gets published to maven.  Is this hack worth it?

kwalcock:
The problem sentences had 17009 and 16942 words in them.  In these documents there were a bunch of other "sentences" with up to 245 words.  I bumped the limit up to 300 so that only the longest ones were filtered and the remaining didn't take too much time.  Bismarck wrote really long real sentences.  I don't know where the best cutoff is, but this lets everything but the problem cases (of these two documents) through.  I'll run one more collection and then I'm satisfied.

bsharpataz:
thanks @kwalcock !

kwalcock:
For the record, at a limit of 300 words, documents from the collection of 52 with these numbers lost sentences: 2, 5, 10, 24, 25, 26, 27, 30, 31, 42, and 49.  They are being checked for false alarms.

MihaiSurdeanu:
Thanks!

kwalcock:
None of the "sentences" with more than 300 words were real sentences in this set of texts.  At most there was a sentence hanging off the end of a table that was skipped, but lowering the number wouldn't help that.  I wonder if this limit should eventually be accessible as a setting.  There will be an issue filed on processors to remind me to add a copy method for documents.  I'm happy otherwise.

PR #516, opened 2018-11-20
Fix the missing diagram by producing expected json

kwalcock:


kwalcock:
You may press the button.

marcovzla:
I see this fixes the error when there is no time or location. But does it actually renders something when it does? Maybe I don't understand brat syntax but I suspect it may not.

kwalcock:
I tested with time and geolocations turned on and it shows at least the TimeExpression (as in https://github.com/clulab/eidos/blob/master/README.md).  I have yet to see a location in the diagram.  The code
```scala
        Json.arr(
          s"X$x",
          "x",
```
does look suspicious.

marcovzla:
I'm pretty sure brat only handles `entities`, `events`, `triggers`, `attributes`, and `relations`, and any other key should be ignored

marcovzla:
I suspect that `timexs` and `geoexps` are being ignored, and what gets rendered comes from `entities` or `events`

kwalcock:
Perhaps it has been updated.

image

image

EgoLaparra:
I think the brat part for Locations was not implemented. So, they shouldn't show up.

kwalcock:
There is this code in main.js:
```javascript
        {
            "type"   : "GeoidPhrases",
            "labels" : ["GeoidPhrases", "GEOLOC"],
            // Blue is a nice colour for a person?
            //"bgColor": "thistle",
            "bgColor": geoLocationColor,
            // Use a slightly darker version of the bgColor for the border
            "borderColor": "darken"
         },
```

kwalcock:
I'll go ahead and check it in though, so at least the time shows on master.  Thanks for your help. everyone.

PR #517, opened 2018-11-20
Make the  mapOntologies method accessible

bsharpataz:
@bgyori  was asking for access to this method to be called chez lui.  I think the re-factoring allows this, but Ben, please check before you approve the PR...?


kwalcock:
@bsharpataz, rather than
```scala
      val lines = source.getLines().toArray
```
can you try
```scala
      val lines = source.getLines()
      val result: Iterator[ConceptEmbedding] = for {
      ...
      }
      result.toSeq()
```
The debugger thinks the problem is indeed with delayed evaluation caused by the iterator (which even survives the for loop), but I think that toSeq() (or vector or array) afterwards will prevent the entire file from needing to be read at once.  (I haven't managed to get the environment set up to call this myself.)

bsharpataz:
I can try later this afternoon, yes :)

kwalcock:
See also #520 that you can take credit for.

kwalcock:
@bsharpataz any luck on using result.toSeq rather than getLines.toArray?

bgyori:
In the comment above I attempted to fix the issue of getting a string with \n between every character. But I think I must have done something wrong because now I'm getting at "head of empty list" exception and know too little about Scala to solve it. Sorry...

bsharpataz:
Sorry, i totally forgot to do this! Will do now

bsharpataz:
> @bsharpataz any luck on using result.toSeq rather than getLines.toArray?

I spoke with @marcovzla and he told me that Seq is a trait, so it will try to map it to the closest type of Seq to what it is, giving me a Stream.  So, I can choose List, Array, etc, Array seems fine.  But I will try to wait to the end to not load the whole line at once

bsharpataz:
> > @bsharpataz any luck on using result.toSeq rather than getLines.toArray?
> 
> I spoke with @marcovzla and he told me that Seq is a trait, so it will try to map it to the closest type of Seq to what it is, giving me a Stream. So, I can choose List, Array, etc, Array seems fine. But I will try to wait to the end to not load the whole line at once

so -- nope.  Seems the autoclose doesn't like me converting after the scope... you're welcome to play

kwalcock:
Becky is right, of course.  `"this\nis\na\ntest".lines.map { _.toUpperCase }.toSeq` doesn't work, but `"this\nis\na\ntest".lines.map { _.toUpperCase }.toArray` does.  It's probably better than `"this\nis\na\ntest".lines.toArray.map { _.toUpperCase }` because this one will need room for two arrays, the lines and the uppercase lines.  So, I'm hoping that the toArray can be moved to as late as possible within the block.  (Not that it's a super big deal.)

bgyori:
Looks like I misplaced my conf file but once I reinstated it, it works - thanks for the tip @bsharpataz ! I pushed another commit to rename the IDs used for each ontology in the mapping file. This is good to go as far as I'm concerned.

kwalcock:
Sorry, @bsharpataz, but there were merge conflicts with the refactored version in master.  I probably resolved them wrong, but you will be able to fix it in an instant if so.  Search for `// Version from master` and that might be code that should replace what's there now from this branch.

bsharpataz:
sorry @kwalcock I don't know what you mean, sorry to be dense!

kwalcock:
Can you check OntologyMapper.scala in this branch?  The parts that are likely merged wrong are commented.

bsharpataz:
ok, yeah, now I see them, ok I'll fix and push... forthcoming

bsharpataz:
please check me @kwalcock !

kwalcock:
You're checked.  It was just unclear to me whether Ben's version was the latest and greatest or not completely.  This compiles and runs fine.  Thanks.

PR #522, opened 2018-11-22
use new indicators

bsharpataz:
- Moved Ajay's file to the ontology folder (fyi @ajaynagesh )
- added adarsh's set (from Mitre) (fyi @adarshp )
- added to EidosSystem and CacheOntologies
- refactored the MapIndicators method to be WAY WAY cleaner and like, sensible.  Should be way way more trivial to add new ones now.  More refactoring could prob be done, but it's a start...
- I went ahead and added the new mapping files bc I think they should be tracked
- add a python script to make ~ontology from flat list


kwalcock:
You paid a high price for making a variable name longer than all the rest!  In the eidos.conf file should ${EidosSystem.language} continue to be used for all ontologies?  (That design may have to be revisited soon.)  Was useCache not working in EidosSystem?  Maybe it was too implicit or I guess redundantly calculated.  Your version is better.  I'm going to run this once and inspect the output before signing off.  We're all thankful for you!

bsharpataz:
Aw thanks Keith! Right back atcha!  useCache prob would have worked just
fine, I occasionally add sanity checks like that without thinking, here I
wanted to be super sure I was passing/using the right thing, but this may
be me waving a palm branch to make the plane land again.... feel more than
free to change it back to useCache !!!

kwalcock:
My comment above about ${EidosSystem.language} will not be relevant until those new files have been added in Portuguese.

kwalcock:
This made it through the 10 and 52, so being merged it is.

PR #525, opened 2018-11-26
Add program to process serialized documents

kwalcock:
This includes is a simple program to address this request:

I’d like to add some new functionality to Eidos. In particular, it should have the ability to parse documents that have been preprocessed by processors (that is, POS tagging and parsing). This output is a JSON format that serializes the document with this information. So, Eidos may see a directory of .txt files (as it does today) or a directory of pre-processed files produced by processors.


kwalcock:
Uh oh.  There's always a way, but did I break it or does ExtractAndExport need to load in the serialized documents as well or is there something else?  So many input formats and so many output ones...

bsharpataz:
none of the above -- just wondering if we can keep the number of apps down, if the functionality can be baked into ExtractAndExport is all

PR #529, opened 2018-11-29
Eidos web app/web service Docker container

brandomr:
### What's new

This PR adds a directory to the top level of the project called `Docker`. Within this directory is a `Dockerfile` which builds the Docker container. Instructions for building the container and running it are included in the readme (`Docker/README.md`).

### Configuration

This container automatically runs the Eidos web application on port 9000. It defaults to using word2vec and Google's pre-trained vectors. In the future this could be configured by a user at the time the container is run, but would require adding GloVe vectors to the container image (thereby increasing its size). 

### Testing this PR

After pulling this PR code you can try:

```
cd Docker
docker build -f Dockerfile . -t eidos-webservice
```

Building the container will take some time. Once it completes you can try:

```
docker run -id -p 9000:9000 eidos-webservice
```

Then, you should be able to navigate to the web application in your browser at `localhost:9000`.

### Outstanding Issues

1. When the container is first initialized via `docker run [etc]` the web application runs, but the vectors are not loaded until the first call to the application is made. Therefore, the first API request to the `process_text` endpoint  is slow. Subsequent requests run normally.
2. The container must be run in interactive mode (with the `-i` flag specified) at run time since the web application requires stdin, which is normally surpressed by Docker. See this [StackOverflow post](https://stackoverflow.com/questions/46637070/akka-http-shuts-down-prematurely-in-a-docker-container) for more detail.
3. A built container could be pushed to a registry (e.g. Dockerhub). This can be done automatically with [Travis CI](https://docs.travis-ci.com/user/docker/).

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -1,0 +1,44 @@
+FROM ubuntu:16.04
+
+# Install base packages
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    software-properties-common \ 
+    dialog \
+    git \
+    graphviz \
+    libgraphviz-dev \
+    vim
+
+# Install Java
+RUN add-apt-repository -y ppa:webupd8team/java
+RUN echo "oracle-java8-installer shared/accepted-oracle-license-v1-1 select true" | debconf-set-selections
+RUN apt-get update && apt-get install -y \
+    oracle-java8-installer 
+RUN export JAVA_HOME="/usr/lib/jvm/java-8-oracle/"
+RUN export PATH="$JAVA_HOME/bin:$PATH"    
+
+# Install Scala
+WORKDIR /
+RUN apt-get remove scala-library scala
+RUN wget http://scala-lang.org/files/archive/scala-2.12.7.deb
+RUN dpkg -i scala-2.12.7.deb
+RUN apt-get update && apt-get install scala
+RUN apt-get install apt-transport-https -y
+RUN echo "deb https://dl.bintray.com/sbt/debian /" |  tee -a /etc/apt/sources.list.d/sbt.list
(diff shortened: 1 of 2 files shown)


kwalcock:
Many thanks!

It seems like the part most likely to fail is the sed'ding of the configuration file.  We tend to change that file without considering potential side effects.  There is one pull request in the pipeline that I think will cause problems, so I'll probably wait briefly and merge that one before this one.

Less about your PR than about Docker -- Is it just standard procedure to install all those base packages?  Are they used for anything?  Can they (e.g., git) be uninstalled after being used just the once?

kwalcock:
@brandomr, thanks for mentioning the vectors file.  It would be great if the Docker distribution used cached versions of the ontologies and the vectors, both for size and speed considerations.  That complicates things quite a bit, though, not that it can't be done.  I wonder if this will be run in an environment where the startup time or memory use is a constraint.  It shouldn't prevent these files from being merged, however.

brandomr:
@kwalcock that would be really cool if possible, I'm not even sure where to start to incorporate that. One use case that I discussed with @bgyori last week would be running Eidos as a web service which could be leveraged by the INDRA REST API his team has [currently running](http://www.indra.bio/rest_api/docs/). Startup time is probably less important than memory use, though I've had success running Eidos on relatively small machines (I tested on an [AWS EC2 t3.large](https://aws.amazon.com/ec2/instance-types/t3/) without problems).

brandomr:
@kwalcock just wanted to bump this before the holidays! I spoke with @jgawrilo about trying to used cached vectors and he thought he may have seen something like that before so could take a look. I think generally though, this container code is a good start so that we can auto-build a container (Travis) and push it to a registry (Dockerhub).

brandomr:
Fantastic, thanks @kwalcock!

PR #530, opened 2018-11-30
Updates for newest geonorm model

bethard:
Vikas updated the geonorm model recently, but since geonorm tests aren't run by default, no one noticed that some tests were failing with the new model (in all cases, the new model was finding locations that the old model wasn't). This pull request first fixes those tests, then updates the code to load a DeepLearning4J (.dl4j.zip) version of the model, instead of loading the Keras (.hdf5) model and converting it to DeepLearning4J every time the geonorm parser is loaded. This also gets rid of the need for extracting the .hdf5 to a temporary directory; we simply load the .dl4j.zip model as a resource stream.

A few other changes that were needed to get this all working:
* An update to the grammar for attaching locations
* A fix to a bug in EidosActions.mergeAttachments, which was discarding location attachments
* A fix to GraphTester.toString to display attachments so that it's possible to debug test failures due to attachments

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,7 +438,7 @@ Eidos and declared in `built.sbt` as a library dependency.  Change the value of
 `useTimeNorm` from `false` to `true` to use the time functions.
 
 The newest model file used for geolocations is too large for GitHub and requires additional installation.
-Download [geonorm_model.hdf5](https://drive.google.com/file/d/16vIRhfEHjEMxnGTemnaUhkQT1eVbUm6j/view?usp=sharing)
+Download [geonorm_model.dl4j.zip](https://drive.google.com/open?id=10d8QSatXHQCfW5G-ON0kC698zo7q5BzT)
 and place it in the directory `src/main/resources/org/clulab/wm/eidos/models/`.  Then change the value of
 `useGeoNorm` from `false` to `true` to use the functions.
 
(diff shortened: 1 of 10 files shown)


kwalcock:
That's a great idea storing the models in Maven! +1

kwalcock:
Maybe this was discussed at the hackathon or elsewhere, but if not, can @bsharpataz check that change 
2ae2a26 in EidosActions.scala because I don't know what it happening there?

bethard:
Yes, @bsharpataz should check that, but in short, the change is making sure that `flattenedContextAttachments` gets included even when `filteredAttachments` is empty. That is, the previous code discarded all context attachments whenever `filteredAttachments` was empty. But that's wrong; all context attachments should always be included.

kwalcock:
Just a note: sometimes zipWithIndex is avoided because it creates an entire new collection with time and space consequences.  This particular code is not concerned with efficiency, so it's a fine change.  Maybe next time, though :-)

kwalcock:
As discussed at the meeting, this PR will eventually include a true value for configuration of both time and location functionality, assuming it doesn't cause problems with Travis or unforeseen consequences.

kwalcock:
Travis told and it passed in around 12.5 minutes, so this is going in.

kwalcock:
Time may need some test updates.  It is also crashing sbt on my computer.  It's being saved for another PR.

bethard:
Thanks for finishing this up, and yes, the time stuff should be enabled in a separate pull request after Egoitz gets the new model running.

PR #531, opened 2018-11-30
Replaces many uses of asInstanceOf with better Scala alternatives

bethard:


kwalcock:
I really like the (filter, map) => collect ones!  Not so sure of the others, but will check carefully after the hackathon quiets down.  Thanks!

bethard:
We should prefer pattern matching to `asInstanceOf` wherever possible, since the latter only "does the minimal thing to make bytecode pass": https://github.com/scala/scala-dev/issues/456

kwalcock:
But I'm still going to check ;-)  In some cases the asInstanceOf is known to be safe and used for optimization, but I'll also check how much the IDEs complain or not.

bethard:
I wasn't suggesting you shouldn't check. ;-)

There are quite a few more `asInstanceOf` uses in Eidos, but the other ones didn't look as easy to get rid of. Things like where the code assumes that a `Document` is always an `EidosDocument`. Those probably imply that the Eidos type system needs to be improved so that the right types are propagated, but that's obviously well beyond what I'm willing to do at this point. ;-)

bethard:
Looks good, and the doc improvement is helpful.

PR #534, opened 2018-12-12
Compactize the patterns

kwalcock:
The compact version likes Arrays rather than Seq, so in haste I changed several of them.  If need be and provided time, they can be converted.

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -26,41 +26,8 @@ EidosSystem {
         geoLoc2IdPath = /org/clulab/wm/eidos/english/context/geo_dict_with_population_SOUTH_SUDAN.txt
            ontologies = ["un", "props"]
              cacheDir = ./cache/${EidosSystem.language}
-               useW2V = true
+               useW2V = false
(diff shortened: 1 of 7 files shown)


kwalcock:
While testing I was reminded that a node like month is pretty much useless because none of the examples makes it through the filter (I think).  season isn't much better.  Is this a problem?

bsharpataz:
> While testing I was reminded that a node like month is pretty much useless because none of the examples makes it through the filter (I think). season isn't much better. Is this a problem?

probably :)  but fwiw we're about to align the ontologies with BBN so I think I'll put that off til then

bsharpataz:
@kwalcock I "fixed" the one test file (by marking the tests as failing -- we decided to remove the trigger that they relied on...).  Can you please help diagnose/fix the JSONLD failing tests? (or tell me if I need to do it bc you're busy).  Thanks!!

bsharpataz:
@kwalcock on the `hackathon_int` branch I marked the tests that are failing as failingTest, so I am going to merge this in

PR #535, opened 2018-12-14
2nd Definition Rule

scottagt:
Rule to recognize definitions like:
"Epidemiology is the study and analysis of the distribution (who, when, and where) and determinants of health and disease conditions in defined populations." Could benefit from further scoping after learning more about the rules.


bsharpataz:
@jpfairbanks @scottagt can we (@marcovzla you’re part of “we”) start a convo about eidos for aske?  We can likely save you some time and/or help you get started...  i have a refactor planned for the holiday break that will hopefully make your way much much much easier...

scottagt:
Hi @bsharpataz, sorry, I accidentally made that push request, and as soon as I realized it, I thought it might start a conversation. I didn't realize that when making a merge request it defaults in the drop-down to the original source repo and not the forked repo. 😅 

It's been nice using the current Eidos system and we've already gotten started using the rule manual and have been writing cursory rules; we will likely go back and do some re-scoping of rules we're writing (in our own forked repo) to better capture what we're focusing on as we become better rule writers too.

re: refactor - Great, I think we're going to continue our current efforts (I'm not sure any performer has time to wait given the timeline) and look forward to checking that out after the holiday break. Feel free to write me about it too. And then I think we would be interested later on next year, after the break and the refactoring, to chat more about Eidos in ASKE as you've stated, that would be great. 

From my understanding each performer has a slightly different focus and for us IE is less a focus and more so in another area. Also, if you're not already aware or didn't meet at the recent meeting, James is the PI of our project and I'm a CoPI so he may have other or differing thoughts. CC: @jpfairbanks

bsharpataz:
@scottagt @jpfairbanks -- yes, let's talk!  I agree that Odin would be a great tool for you to use (it's extremely flexible and works over multiple representations of text), and the grammars in eidos are Odin grammars.  

After the refactor, I think that eidos would (if all goes well) be a really good starting point.  Until then, however, there are problematic assumptions about the content of the grammars (e.g., that the relations/events have `cause` and `effect` arguments).  This will get in your way A LOT while you try to write rules for definitions, etc.  

The refactor will be designed to (a) remove these assumptions (allowing for better customization of *what* you want to extract), (b) make the various components toggleable (e.g., entity finding, expansion, and grounding).  And your input would be very much appreciated!

However, depending on what you want, **you may want to start from "scratch"** -- essentially Odin with a webapp.  In our case, for the ASKE grant, we're planning on doing this precisely because of the time constraints you mentioned.  

We can help you (and I think it will help!):
1) we can set you up with a stub repo (NLP preprocessing pipeline, odin, and a webapp)
 - this will let you move fast and flexibly 
2) we can have the call to explain/help get you started/etc (next week is good for us, depending on the day) 
3) we can help with early grammar development as much as we have time (but we're all for collaboration fwiw)
 - we can even share rules we develop (for variables, etc) that really may help you
 - OR/ALSO you can fork the repo we start (essentially a stub like mentioned above) for ASKE (it will all be open source anyway!), and you can make use of any rules we write etc.
- ideally, what we develop will be complementary too!

We can talk more about this all on the call :)
(ps -- we can also talk about the other questions/points above, they seemed easier to discuss "live" than in this already too long message) :)

PR #536, opened 2018-12-21
Hackathon int

bsharpataz:
Merging in the ontology pattern capabilities, added patterns for the intervention nodes.


bsharpataz:
@MihaiSurdeanu you’re right, but why should we? Grounding isn’t even default true.  Idk how many sentences, but I can check, didnt you say Processors uses 150? And for the backoff, yes, every node is eligible as a backoff, but with this PR only the 4 categories and generic “humanitarian aide” are in the un ontology.

MihaiSurdeanu:
- we'd like to have geonorms enabled for Travis tests
- just curious on the impact of the 150 threshold. Not that important.
- ok on the grounding semantics !


________________________________

bsharpataz:
ok -- reset the config values -- please let me know if other changes are needed or if this can be merged.  thanks!

PR #537, opened 2018-12-31
Now grounding to interventions when the mention is grounded to a specific branch

bsharpataz:
I tried to do this in a cleanerish-ish way so that in the future we can have other "plugins"... I am not in love with my overloaded methods in OntologyGrounder (specifically `groundable()` and `groundOntology()` -- so if you have suggestions about how to do this better that would be great).

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -8,7 +8,7 @@ EidosSystem {
       entityRulesPath = /org/clulab/wm/eidos/${EidosSystem.language}/grammars/entities/grammar/entities.yml
        avoidRulesPath = /org/clulab/wm/eidos/${EidosSystem.language}/grammars/avoidLocal.yml
          taxonomyPath = /org/clulab/wm/eidos/${EidosSystem.language}/grammars/taxonomy.yml
-  //      wordToVecPath = /org/clulab/wm/eidos/${EidosSystem.language}/w2v/vectors.txt
+//        wordToVecPath = /org/clulab/wm/eidos/${EidosSystem.language}/w2v/vectors.txt
         wordToVecPath = /org/clulab/wm/eidos/${EidosSystem.language}/w2v/glove.840B.300d.vectors.txt
         stopWordsPath = /org/clulab/wm/eidos/${EidosSystem.language}/filtering/stops.txt
       transparentPath = /org/clulab/wm/eidos/${EidosSystem.language}/filtering/transparent.txt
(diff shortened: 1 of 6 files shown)


bsharpataz:
@kwalcock and @MihaiSurdeanu 
I refactored with Marco's help and cleaned the accessing in EidosSystem -- please let me know if it's ok, I like it better this way bc it avoids the casting issue etc.

MihaiSurdeanu:
I'm sure you saw it: there is a compilation error now.

bsharpataz:
Well poop. Ok, i’ll fix that asap

kwalcock:
I appreciate the fewer switches on class/type.

kwalcock:
At one time I had concluded that the original
```
mitre12OntologyPath = /org/clulab/wm/eidos/english/ontologies/mitre12_indicators.yml
```
was better than
```
mitre12OntologyPath = /org/clulab/wm/eidos/${EidosSystem.language}/ontologies/mitre12_indicators.yml
```
because there is no Portuguese translation of the file.  If someone sets the language to Portuguese, then there *might* be a file not found.  Maybe that's better than accidentally using the hard coded English, though.

kwalcock:
It would be great if this code, which has probably been there for quite some time could be changed:
```
val canonicalName = mention.canonicalName
val canonicalNameParts = canonicalName.split(" +")
```

Something goes to great lengths to build a string only to have it ripped apart again.  Could the parts method be added to EidosMention to perform the calculation up to
```
canonicalMentions.sortWith(lessThan).map(canonicalFormSimple)
```
and then have another local method add the mkString(" ") part?  Clients can call either as needed.

This is definitely not critical, but maybe as long as the code is being scrutinized...

bsharpataz:
> It would be great if this code, which has probably been there for quite some time could be changed:
> 
> ```
> val canonicalName = mention.canonicalName
> val canonicalNameParts = canonicalName.split(" +")
> ```
> Something goes to great lengths to build a string only to have it ripped apart again. Could the parts method be added to EidosMention to perform the calculation up to
> 
> ```
> canonicalMentions.sortWith(lessThan).map(canonicalFormSimple)
> ```
> and then have another local method add the mkString(" ") part? Clients can call either as needed.
> 
> This is definitely not critical, but maybe as long as the code is being scrutinized...
> 
> I have noticed that the regular expression handling for " +" is much slower than a split on ' ' FWIW.

Hahaha, would you believe I actually had started that here too and then decided I should wait -- ask and you shall receive!   Please check the changes to EidosMention :)
BTW I have high hopes for a MAJOR refactor of Eidos VERY SOON to make it more usable in other projects....

bsharpataz:
> At one time I had concluded that the original
> 
> ```
> mitre12OntologyPath = /org/clulab/wm/eidos/english/ontologies/mitre12_indicators.yml
> ```
> was better than
> 
> ```
> mitre12OntologyPath = /org/clulab/wm/eidos/${EidosSystem.language}/ontologies/mitre12_indicators.yml
> ```
> because there is no Portuguese translation of the file. If someone sets the language to Portuguese, then there _might_ be a file not found. Maybe that's better than accidentally using the hard coded English, though.

I do think that a crash is better than a silent wrongness

bsharpataz:
> I appreciate the fewer switches on class/type.

me too!! yay for @marcovzla !

MihaiSurdeanu:
Nice! And +1 on the refactoring!

bsharpataz:
@kwalcock fyi I'd personally vote for merging this before you start the run.

kwalcock:
@bsharpataz, I think that the latest change may be causing a stack overflow.  There may be some infinite recursion there that is not tested on Travis where we don't do grounding.

bsharpataz:
yep! @kwalcock and @MihaiSurdeanu you're right! I discvered that myself when i was working on the sanity check.  Trying to fix it, I'll work on it tonight -- I think i almost have it pinned down

PR #539, opened 2019-01-04
Update timer to keep in sync with 17k version

kwalcock:


kwalcock:
Just so I can diff the two branches and note only added apps...

PR #544, opened 2019-01-18
Kwalcock tidy j object

kwalcock:


kwalcock:
This addresses some issues recorded in an email thread, some of which is pasted here:

Just a side note about this to the two of you: Keith is right that some metadata related to documents is not created when reading arbitrary text with the webapp; this is to be expected. Another thing I found was that in JSON-LD produced when using the Scala API directly, empty entries under "states" or "modifiers" either were not in the JSON-LD at all (i.e. the "states" or "modifiers" key is not there) or were empty lists as in "modifiers": [], whereas in JSON-LD from the webapp, we encounter null values in some of these positions, e.g. "modifiers": null. This is not a big deal and can be handled easily downstream, just required some extra attention when we first started processing output via the webapp.

On Thu, Jan 17, 2019 at 12:00 AM Keith Alcock <keith@keithalcock.com> wrote:

The recently produced eidos output and the webapp would differ in at least the metadata handling (e.g., missing title without metadata) and the webapp version could be missing a filename somewhere, for example.  I hope it is something on that order.  Perhaps Ben can let me know if there is some gotcha that should be documented.

On Wed, Jan 16, 2019 at 9:26 PM Mihai Surdeanu <surdeanu@gmail.com> wrote:
That should not happen, right Keith?

PR #549, opened 2019-01-21
refactor: moving grounding and filtering out

bsharpataz:
Moved a bunch of the filtering components to StopwordManger, also moved the ontology loading to OntologyHandler, refactored to (a) simplify the types and (b) make it easier to add new ontologies whenever we want
Also added a new ontology (the catalyst for all of this -- icasa variables).

Moved a couple of related things out of LoadableAttributes bc of a nullpointer exception I couldn't fix, but that was the extent of the intentionality behind that, so feel free to rework...

more refactoring to come...

fyi @MihaiSurdeanu

### DIFF
--- a/src/main/python/icasa_to_yaml.py
+++ b/src/main/python/icasa_to_yaml.py
@@ -1,0 +1,36 @@
+import sys, re
+from mk_yaml_ontology import ont_node, dump_yaml
+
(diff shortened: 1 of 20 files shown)


kwalcock:
@bsharpataz, is the next PR, of the lexiconner_optional branch, continued development of this branch then?

PR #551, opened 2019-01-21
refactor pt 2: Lexiconner and entityfinder optional

bsharpataz:
This would ideally be PRd after the dssat_vars branch (i made it from that branch, so the "diffs" will greatly reduce once that is merged....!)

I made the LexiconNER and EntityFinder optional (toggled in the config) and moved the params of those and the geonorm model to the companion objects in a `fromConfig` method

Updates to the configs accordingly

fyi @MihaiSurdeanu

### DIFF
--- a/src/main/python/icasa_to_yaml.py
+++ b/src/main/python/icasa_to_yaml.py
@@ -1,0 +1,36 @@
+import sys, re
+from mk_yaml_ontology import ont_node, dump_yaml
+
+def replace_token(pattern, replacement, s):
+    replacement = f' {replacement} '
+    s = re.sub(f' ?{pattern} ', replacement, s)
+    s = re.sub(f' {pattern} ?', replacement, s)
+    return s
+
+def mk_ont_node(line_string):
+    fields = line_string.split("\t")
+    assert(len(fields) >= 4)
+    var_name = fields[0].strip()
+    description = fields[3].strip()
+    description = replace_token("C", "carbon", description)
+    description = replace_token("CO2", "carbon dioxide", description)
+    description = replace_token("CH2O", "formaldehyde", description)
+    description = replace_token("N", "nitrogen", description)
+    description = replace_token("NH3", "ammonia", description)
+    description = replace_token("NH4", "ammonium", description)
+    description = replace_token("NO3", "nitrate", description)
+    description = replace_token("P", "phosphorus", description)
+    return ont_node(var_name, [description], None, add_name = False)    # the name isn't in a format we can use
+
+
+def main():
+    flat_file = sys.argv[1]
+    ont_file = sys.argv[2]
+    ont_name = sys.argv[3]
+    with open(flat_file, "r") as f:
+        _ = f.readline()    # read header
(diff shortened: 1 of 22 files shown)


MihaiSurdeanu:
This seems a lot cleaner. Of course, tests fail now...

One question: in the interest of modularity, Eidos should work when the `ontologies` variable is empty, by doing no grounding. What happens today?

bsharpataz:
@MihaiSurdeanu -- the tests failed bc i accidentally committed changing useW2V to true -- fixing now...
Re: no ontologies -- idk, i'll check.  Right now grounding is toggled through turning w2v on or off.  I think that w2v is currently only being used in grounding, so one of the immediate next plans is to move all of the grounding magic (including w2v) to another place.
In fact, since apparently I will never have time to meet with @kwalcock between meetings (!) the (not particularly ordered) current list for future refactors is:
- refactor EidosActions (remove assumptions -- will likely be several steps)
- simplify/unite the code to filter mentions out which don’t have content, can we use the `keep` attribute? (i'd like to combine the StopwordManager and the Canonicalizer into one thing at this step...)
- have a single “grounder” (with all ontologies etc) and move the w2v to the grounder
- move the domain params and all to some other "Handler"… but think more generally? IDK yet about this.  It's super domain dependent so I'd like to brainstorm a way to make it at least a _little_ more domain general...

Suggestions welcome, and we can move this to an issue as well...

kwalcock:
There are configuration files in the test directory that could be used to ensure that Travis doesn't fail if the main configuration file has been changed, but I agree with you @bsharpataz that it's better to fail quickly, in this case from the unexpected configuration.

kwalcock:
He likes it!  Hey Mikey!  Or vice versa.  Here are some notes I was making, but they don't mean that any code needs to change.

* Some (unimportant) things that were "reloaded" before aren't now.  There's a slight difference between `ieSystem.stopwordManager` and `ieSystem.loadableAttributes.stopwordManager`.  Inconsequential, but heads up.
* I like fromConfig and maybe EidosSystem needs one as well.  This could take care of (some of) the reloading.  It might just be an overloaded apply().
* logger.info(s"Processing yml ontology ${ontologyPath}..")
* Glad that I'm not the only one removing blank lines :-)
* The match/case statements in StopwordManager concern me a little.  Something might not be in the right place.  Maybe grep for CAG_EDGES.  I'm not sure something like this is any better because of the cast.  Probably a TODO for later.
```scala
  def releventEdge(m: Mention, state: State): Boolean = {
    val isCagLabel = EidosSystem.CAG_EDGES.contains(m.label)

    if (isCagLabel && m.isInstanceOf[EventMention])
      argumentsHaveContent(m.asInstanceOf[EventMention], state)
    else
      isCagLabel
  }
```
* It seems like EidosSystem.stopwordManager is there so other things don't have to use eidosSystem.loadableAttributes.  But there are still lots of uses of the latter, especially with adjectiveGrounder.  That may be temporary.

Thanks!  I hope you knew that today was a holiday.

MihaiSurdeanu:
+1,000!

PR #553, opened 2019-01-23
More corr

bsharpataz:
per convo with @bgyori , we weren't capturing "associated with" patterns:

> So I just noticed that there are a lot of occurrences of “associated with” phrases that could be extracted as Correlations but “association” doesn’t appear as a trigger word. There are even some test sentences like
> ```Significant decline in poverty will be associated with a decrease in family size and increase in non-farm income .```
> but currently I’m not getting a correlation from these
> I was wondering if adding this as a trigger and writing some extraction rules would be a good idea or if you avoided this on purpose

not on purpose, so here it is! :)

image

thanks @bgyori !

Also -- @MihaiSurdeanu -- fun fact: when we only show "CAG Relevant" extractions we lose the `Dec(family size)` Mention bc both `family` and `size` are transparent/filtered (I think we likely added these individually on purpose, but as an FYI in case we want to rethink any of those decisions)


MihaiSurdeanu:
@bsharpataz: let's discuss the `Dec(family size)` issue at the project meeting.

PR #554, opened 2019-01-24
Separate the AdjectiveGrounder from EidosSystem

kwalcock:


kwalcock:
@bsharpataz, this should be along the lines of what we discussed this morning, applied so far only to the AdjectiveGrounder.  There were lots of subjective decisions, so don't hesitate to suggest making the other choice.  We did discuss having 0, 1, or more of various components.  The switch betweeen 0 and 1 for this one has to be done in code.  It might be nice to enable or disable this through configuration and that might end up being appropriate for other components.  At some point it could look like
```
adjectiveGrounder {
  enabled = true
  quantifierKBPath = /org/clulab/wm/eidos/${EidosSystem.language}/quantifierKB/gradable_adj_fullmodel.kb
}
```

kwalcock:
@bsharpataz is full of great ideas!  I didn't realize that the DomainParams was so closely related even though inclusion in the same directory might have tipped me off.  It looks like it is fairly independent of the AdjectiveGrounder, though, so I'm thinking that it should have its own "stanza" but you may know better.  The numbers in the two files may be coordinated, for example, so that the files should be paired in the same stanza.

kwalcock:
I would like to wait on the `enabled` idea to see to what extent it applies to the other components, including the ones with useThis or useThat.  The strategy may imply that (Eidos)AdjectiveGrounder.fromConfig needs to return a NullAdjectiveGrounder supporting the same interface if `enabled = false`.  Alternatively, the adjectiveGrounder can manage an enabled flag of its own that can be queried by clients.  For that matter, fromConfig could return an Option[] and None if the component is disabled.  That sounds like something for the next round.

bsharpataz:
> @bsharpataz is full of great ideas! I didn't realize that the DomainParams was so closely related even though inclusion in the same directory might have tipped me off. It looks like it is fairly independent of the AdjectiveGrounder, though, so I'm thinking that it should have its own "stanza" but you may know better. The numbers in the two files may be coordinated, for example, so that the files should be paired in the same stanza.

They are super-related I promise (the domain params file is a theoretical accompaniment to the adj model, supplying the background knowledge about domain-relevant entities).  In reality it's just a toy to demonstrate a potential capability, but regardless, it's meaningless without the adj grounder and was designed to showcase the adj grounding.

kwalcock:
@bgyori, is it a good time to check this in?  It may require a change in your Python code.  If something was using adjective grounding, it would change from
```
val corpus = new JLDCorpus(annotatedDocuments, reader.loadableAttributes.adjectiveGrounder)
val mentionsJSONLD = corpus.serialize()
```
to
```
val corpus = new JLDCorpus(annotatedDocuments)
val adjectiveGrounder = EidosAdjectiveGrounder.fromConfig(EidosSystem.defaultConfig.getConfig("adjectiveGrounder"))
val mentionsJSONLD = corpus.serialize(adjectiveGrounder)
```
If it isn't using adjective grounding, then client code can remain unchanged like this:
```
val corpus = new JLDCorpus(annotatedDocuments)
val mentionsJSONLD = corpus.serialize()
```

bgyori:
Got it, I can make the necessary change whenever you merge this!

kwalcock:
Here goes then...

PR #558, opened 2019-02-06
Removing obsolete FAO and WDI ontologies

adarshp:
This PR removes the obsolete FAO and WDI ontologies - the variables in them are now subsumed by the MITRE12 ontology.


bgyori:
I agree with @bsharpataz in that the inclusion of these mappings can be turned on/off using the config. One advantage of the WDI and FAO mappings is that they correspond to data entries in those sources directly so it might make sense to keep them around just in case.

kwalcock:
Is there consensus on this?

BeckySharp:
Ok, I agree with @bgyori , @adarshp i think i’ll close for now but we can always reopen?

adarshp:
@BeckySharp sure! I have reverted the removal of the files.

PR #561, opened 2019-02-24
Remove duplicate provision of

### LINKED ISSUE #560: Provision of duplicated in intervention ontology
In the internvetion ontology, many entries have "provision of" duplicated in their name and examples, e.g. "provision of provision of non-formal education". For instance there are several examples in this section:
https://github.com/clulab/eidos/blob/master/src/main/resources/org/clulab/wm/eidos/english/ontologies/interventions.yml#L157-L182
I just want to confirm this is not on purpose. I'll PR a fix if it's confirmed that this is a problem.

bgyori:
Fixes #560


bsharpataz:
👍 ben!

PR #566, opened 2019-03-10
Reground

BeckySharp:
Hey -- I think that @bgyori  was able to use this properly, it extends OntologyHandler to be able to reground

I have another branch i'm working on for the refactor, but would like to merge this first

fyi @MihaiSurdeanu

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/apps/OntologyMapper.scala
+++ b/src/main/scala/org/clulab/wm/eidos/apps/OntologyMapper.scala
@@ -200,14 +200,23 @@ object OntologyMapper {
     * @param topN the number of similarity scores to return, when set to 0, return them all (default = 0)
     * @return String version of the mapping, akin to a file, newlines separate the "rows"
     */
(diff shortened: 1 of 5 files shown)


kwalcock:
Will nodePatternsMatch be in the next version?  Can it be changed to something like
```
  patterns.map(_.exists(_.findFirstIn(s).nonEmpty)).getOrElse(false)
```

BeckySharp:
> Will nodePatternsMatch be in the next version? Can it be changed to something like
> 
> ```
>   patterns.map(_.exists(_.findFirstIn(s).nonEmpty)).getOrElse(false)
> ```

It can be anything your heart desires, especially if you slip me the juju like the above :)
primarily, i was focusing on moving w2v out of eidos system as much as possible to keep the grounding more compartmentalized, and trying to simplify the W2V stuff (i.e. not have a real and fake one, but rather operate in Options).  The deeper I go, the more I think that PR number 3 in this series will be a reimagining of what it means to be a "grounder" but it may have been a 4am pipe dream :)

PR #569, opened 2019-04-01
Refactoring to decouple w2v from EidosSystem

BeckySharp:
Fairly major changes, @kwalcock please check over, and also @bgyori if you can re-check to make sure the `reground` api still works that would be swell!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -89,6 +89,35 @@ publishTo := {
 // let’s remove any repositories for optional dependencies in our artifact
 pomIncludeRepository := { _ => false }
 
+
+// These values in scmInfo replace the <scm/> section previously recorded in
+// pomExtra so that default values aren't used which then double up in the
+// XML and cause a validation error.  This problem was first noted with
+// sbt.version=1.1.6
+// addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.8")
+// addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "2.3")
+// This produced
+// <scm>
+//     <url>https://github.com/clulab/eidos</url>
+//     <connection>scm:git:https://github.com/clulab/eidos.git</connection>
+//     <developerConnection>scm:git:git@github.com:clulab/eidos.git</developerConnection>
+// </scm>
+// that must be automatically generated and a duplicate
+// <scm>
+//     <url>https://github.com/clulab/eidos</url>
+//     <connection>https://github.com/clulab/eidos</connection>
+// </scm>
+// Judging from this, the scmInfo is collected automatically, perhaps by
+// addSbtPlugin("com.typesafe.sbt" % "sbt-git" % "0.9.3")
+// However, the developerConnection is undesired, so this is used:
+scmInfo := Some(
+  ScmInfo(
+    url("https://github.com/clulab/eidos"),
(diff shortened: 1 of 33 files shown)


BeckySharp:
> This is fine. But isn't there an EidosOntologyGrounder class now? Where is that?

I think that there are lots of Grounder classes of various flavors, so more refactoring is still needed.  To better answer your question, did you want me to send you pointers to the various ones?

BeckySharp:
@MihaiSurdeanu 
trait for Grounders: https://github.com/clulab/eidos/blob/reground/src/main/scala/org/clulab/wm/eidos/groundings/OntologyGrounder.scala#L25

Sorry -- IDK what you mean re: implementation with w2v -- the w2v is now "owned" by OntologyHandler (https://github.com/clulab/eidos/blob/reground/src/main/scala/org/clulab/wm/eidos/groundings/OntologyHandler.scala#L11) which has several Grounders (i.e., one for each active ontology).  
Specifically, the ontologyHandler gives EidosSystem a MultiOntologyGrounder (which, as I understand it is just a wrapper for a Seq[OntologyGrounder] ??): 
- https://github.com/clulab/eidos/blob/reground/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala#L111

- https://github.com/clulab/eidos/blob/reground/src/main/scala/org/clulab/wm/eidos/groundings/OntologyHandler.scala#L17

In terms of functionality, the grounding is still "enabled" by setting `useW2V=true` in the config, but that's now in a diff section of the config (bc EidosSystem doesn't know/care if it's on or off): https://github.com/clulab/eidos/blob/reground/src/main/resources/eidos.conf#L25

The actual grounding happens as the Mentions are converted to EidosMention  -- which is a very mysterious portion of the code to me, but it certainly works... :)  But every time I try to trace through it I tell myself I'll sign up for a CS course or 2

MihaiSurdeanu:
Why do we need a MultiOntologyGrounder here?
https://github.com/clulab/eidos/blob/reground/src/main/scala/org/clulab/wm/eidos/groundings/OntologyHandler.scala#L17

Maybe @kwalcock knows?

BeckySharp:
also -- here's the grounding line: https://github.com/clulab/eidos/blob/reground/src/main/scala/org/clulab/wm/eidos/mentions/EidosMention.scala#L164

kwalcock:
@MihaiSurdeanu, the OntologyHandler is rather new to me and I'll look more closely.  Perhaps the return value can be made more generic.

@BeckySharp, I think the grounding line makes grounding part of the EidosMention constructor.  They are inherently grounded rather than having a var and an Option so that it can be done later or not at all or independently with some external method.   There are so many ways to do it that the best way likely changes with the circumstance.  Perhaps it's time to reevaluate.

bgyori:
I'm testing this branch now and looks like some adaptations are needed on our end which I'm working on, I'll post again when ready!

bgyori:
Okay, this works well!

kwalcock:
I *think* that #570 is the more urgent PR.  This one doesn't necessary need to be merged before the other, but it may well be easier to do all at once.  I was hoping to use that strategy with #567 as well.

kwalcock:
So maybe I should just merge these all so that you can simply use master.

BeckySharp:
However you want to do it is fine with me!

PR #571, opened 2019-04-01
Avoid creating extra Seq when adding label to mention

kwalcock:


kwalcock:
This won't save as much CPU time as it cost to write it.

PR #573, opened 2019-04-04
Refactor actions

BeckySharp:
This is a start to refactoring the EidosActions (aimed at making eidos more modular and so domain and task independent). This is by no means "the solution" but just (I think) a step in the right direction.

**NOTE 1**: I think this _does_ allow for using eidos to extract *non-causal* events by disabling the coref (I think that corefHandler is the only place where explicit refs to "cause" and "effect" args are made)

**NOTE 2**: made from master, so after next round of PRs may need conflicts resolved

Highlights:
 - `fromConfig` constructor for EidosActions that makes expansion and coref not only optional but swappable (note the `useCoref` and `useExpansion` keys in config)
 - CorefHandler which should make coref approaches swappable (note the `corefType` key in config)
 - Slight changes to config again (sorry!) -- the placement of the `stopWordsPath` and `transparentPath` in `EidosSystem` block of config
- I _may_ have added some grammar tweaks to make 2 more tests pass, even though I completely agree that this was the wrong place to do it (i.e., I should have made it a separate PR...) :)

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -6,7 +6,7 @@ EidosSystem {
         stopWordsPath = /org/clulab/wm/eidos/${EidosSystem.language}/filtering/stops.txt
       transparentPath = /org/clulab/wm/eidos/${EidosSystem.language}/filtering/transparent.txt
           hedgingPath = /org/clulab/wm/eidos/${EidosSystem.language}/confidence/hedging.txt
-    timeNormModelPath = /org/clulab/wm/eidos/models/timenorm_model.hdf
+    timeNormModelPath = /org/clulab/wm/eidos/models/timenorm_model.hdf5
           useTimeNorm = false
            useGeoNorm = true
  keepStatefulConcepts = true
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -3,6 +3,8 @@ EidosSystem {
              language = english
       masterRulesPath = /org/clulab/wm/eidos/${EidosSystem.language}/grammars/master.yml
          taxonomyPath = /org/clulab/wm/eidos/${EidosSystem.language}/grammars/taxonomy.yml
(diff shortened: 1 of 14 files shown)


kwalcock:
Having to resolve conflicts will teach me not to slack off ;-)  Still testing the previous conflict resolution...

kwalcock:
@BeckySharp, are all the changes other than those to the rules thought to preserve behavior?  If so, I'll compare before and after just as a sanity check.  I think the code is looking better and better.

BeckySharp:
> @BeckySharp, are all the changes other than those to the rules thought to preserve behavior? If so, I'll compare before and after just as a sanity check. I think the code is looking better and better.

yes -- they are, and if I had thought about it like that, I would have done the rule change commit on a diff PR for sure (lesson learned, thanks!).  Those rule changes are on a diff commit **I think** (https://github.com/clulab/eidos/commit/d915dba62bd1afee16c2bfcf136ccec41e883d34), though malheureusement not the _last_ commit.  Are you able to easily tease it apart?  Or do you want me to remove those changes chez moi and make a diff PR?

BeckySharp:
also, thanks! :)

kwalcock:
I can tease it out just enough for a test and I only noticed because I've done that several times myself and finally decided it might not be the best idea.  It looks like the cat walked across your keyboard.  I think you meant to type helaas.

kwalcock:
The rule change resulted in minor changes to the output of 7/52 files.

kwalcock:
I have no idea what is happening with all these configuration files.  I'm probably going to want to send this merge back to you, especially if I can figure out how to back out of the mess.

BeckySharp:
I can work on the configs -- give me a min

kwalcock:
Thank you!!!  Oops all you want.  I'll run a few more tests.

BeckySharp:
@kwalcock mind if I merge?  I'd like to work on something and would like to use this as a starting point

kwalcock:
Sorry, someone was distracting me.

PR #574, opened 2019-04-16
Read in the JSONLD files

kwalcock:
Notes are being composed

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/EidosActions.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosActions.scala
@@ -410,7 +410,7 @@ class EidosActions(val taxonomy: Taxonomy, val expansionHandler: ExpansionHandle
       trigger = m.asInstanceOf[EventMention].trigger
       theme = tieBreaker(m.arguments("theme")).asInstanceOf[TextBoundMention]
       times = m.document.asInstanceOf[EidosDocument].times
-      time: Option[TimeInterval] = if (times.isDefined) times.get(m.sentence).find(_.span._1 == trigger.startOffset) else None
+      time: Option[TimeInterval] = if (times.isDefined) times.get(m.sentence).find(_.span.start == trigger.startOffset) else None
     } yield time match {
       case None => theme
       case Some(t) => theme.withAttachment(new Time(t))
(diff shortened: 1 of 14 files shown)


kwalcock:
Notes

In order to read in the JSONLD, some changes had to be made to the output.  As updated on the Wiki page (https://github.com/clulab/eidos/wiki/JSON-LD), the Provenance of the modifiers and states needed to be recorded.  Also, the text of the Words is no longer the raw text (such as °) but rather the processed text (such as degree).  We have no way at all to recover the raw text from the processed text and processing the raw text requires a run through processors, which we are trying to avoid.  The text of the Sentence also now matches the processed version.  As mentioned on the Wiki (https://github.com/clulab/eidos/wiki/Anatomy-of-Mentions), Sentence.raw is just a copy of Sentence.words now after the file has been read back in.

Attachments used to be associated with a mention, which was used to provide provenance.  Since we haven't been storing these mentions in the JSONLD and don't want to have to unnecessarily recreate them when reading it back in, Attachments no longer use the mention, but only its extracted provenance.  Provenance was added to Hedging and Negation.  The trigger, on the other hand, needs its Mention in part because it is Odin code.  This mention is created based on information from the EventMention that it triggered.

In order to get a perfectly matching round-trip, the order of the mentions was changed.  They should appear in fairly strict document order (left to right, top to bottom).  They are also numbered better so that their names should match the order of appearance.

Some time-related tuples were formalized so as not to need _._1 or _._2 and case classes were cleaned up.  The location of Option was moved around so that the deserializer could better work with times.

Some text constants were moved to companion objects so that the constants could be better shared with the deserialization code.

One bug was fixed where the DocumentFilter was overwriting the document text.

kwalcock:
Sorry that these changes will be so difficult to follow.  Perhaps it was time to squash the multiple commits into a single one.

The code has been tested at various times against the internal tests and the collections of 10, 52, and 500 documents.  It seems to be holding up well.

kwalcock:
Stop the press.  I've concluded that my statements related to the raw values of words is incorrect.  I'm starting to fix them now.

kwalcock:
@MihaiSurdeanu, @BeckySharp, any reservations (such as clean up your sloppy code!) before I assume silence implies satisfaction?  It's going to be difficult to follow what happened in the individual commits.  A before and after diff might work.  Almost all changes were to the Deserializer class, which will sit there largely unused.  Changes to other classes might be worth some attention, however.

MihaiSurdeanu:
Silence is satistfaction in my case!

BeckySharp:
I'm good with it I'm sure!

PR #575, opened 2019-05-09
Test for crlf in resources

kwalcock:
The standard has been to store text files in GitHub with LF line endings.  Usually this happens transparently, especially if AutoCRLF is on.  It usually doesn't matter anyway because almost all modern programs handle the different line endings appropriately.  It's only a problem if Keith (me) needs to publish to maven.  Then the files on the local computer have to be LF.  I had been keeping AutoCRLF on, mostly so that if there were any files with CRLF in the repo already, they would be automatically converted back to LF when I check them in.  Unfortunately, that means that files on my computer are CRLF and I have to remember to re-clone with different settings.  That doesn't always happen, so instead I'm adding this check to make sure that at least the resources are LF at GitHub.  That way I can skip the eol conversions and re-cloning and publish with less chance of mistake, at least that's the working hypothesis.  This should not really affect anyone else.  If some files turn out to be problematic, they can be skipped in the test.


kwalcock:
cf. https://github.com/clulab/reach/issues/627

PR #576, opened 2019-05-15
Api 2

BeckySharp:
it's a doozy, @kwalcock !
Major points:
- lexiconner has become a finder
- expanders are more modular for argument vs mention expanding
- moved some stuff to config files
- ...
todo still:
- postprocessing step
-  to Eidos mention, grounding, canonicalization, expansion as postprocessingStep
- see if we can combine some of the filtering/mention/attachment merging code...?

thanks!

also, sorry for the extraneous comment in all commits, i forgot to delete something I was playing around with...


kwalcock:
I'm not sure what time zone you are in, but you should probably be asleep.  Can you check the conflicts and the compilation problem?  If you are indeed asleep, I'll start with the compilation.

BeckySharp:
@kwalcock i think i resolved those properly....

BeckySharp:
sorry -- a couple of old versions of files were pushed eons ago, and i forgot, so when i decided not to push changes to those unused files, the old (not compiling) versions were still there... they are removed.

kwalcock:
If I wan't to use this to match the previous output, do I need to change `entityFinders = ["gazetteer", "rulebased"]` to just `entityFinders = ["rulebased"]`?  Even after I do that, there are a lot of changes in the entity-ness of words.  The old version will have `"entity" : "B-Quantifier",` and the new `"entity" : "O",`.  Is this expected?

kwalcock:
Was something like `validIncoming.map(toRegex).toSet` purposefully chosen above `validIncoming.map(_.r).toSet`?  That looks like me trying to avoid an abbreviation.

BeckySharp:
> If I wan't to use this to match the previous output, do I need to change `entityFinders = ["gazetteer", "rulebased"]` to just `entityFinders = ["rulebased"]`? Even after I do that, there are a lot of changes in the entity-ness of words. The old version will have `"entity" : "B-Quantifier",` and the new `"entity" : "O",`. Is this expected?

no -- the gazetteer entity finder replaces the LexiconNER, so it is needed to replicate the previous results.

BeckySharp:
> Was something like `validIncoming.map(toRegex).toSet` purposefully chosen above `validIncoming.map(_.r).toSet`? That looks like me trying to avoid an abbreviation.

nope -- idk, guess i didn't think about it, want me to change? do you want to?

BeckySharp:
and also -- the order matters for the entity finders (gazetteer before rulebased), which prob deserves a comment in the config...

kwalcock:
Thanks.  I can make the one trivial change.  In general, master finds more extractions and I'm trying to figure out why.  The gazetteer changes some of the rule names, but that can be dealt with.

kwalcock:
The problem with B-Quantifier was my fault for having turned off the gazetteer.

kwalcock:
With the conceptExpander maybe in use now, the number of extractions matches more closely, but there are some differences I'm trying to track down.  Does this look familiar?

| master | api_2 |
| --- | --- |
| high figures | figures |
| The situation | situation |
| this alarming situation | situation |
| little rainfall | rainfall |
| significant potential for increased cereal production | potential |
| increased cereal production | cereal production |
| the highest annual rainfall | annual rainfall |
| Accurate data on crop area and production for South Sudan | accurate data |
| considerable uncertainty in estimates | uncertainty |

BeckySharp:
looks like it's not expanding into the increase/decrease.  we used to have
this behavior, so it's not a deal breaker, but i didn't expect it to
change, so i should look at it.  can you please send me a sentence or 2 so
i can debug?

kwalcock:
I forbid you to work on it before Monday ;-( but I'll try to add sentences to the table above.

kwalcock:
If you are testing, you might want to include #578.

BeckySharp:
thanks! will do!

BeckySharp:
@kwalcock I think I addressed it -- i was using the wrong type of expanded for the non-event-participating Concepts.  PLease check to see if fixed.
Thanks!

kwalcock:
Thanks!  I got distracted, but will come back to this.

kwalcock:
It's looking very good so far.  There are some curious instances of the rules being switched around, though.  One jsonld file will have in the master branch
```
"rule" : "simple-np++property-lexiconner",
```
and in the api2 branch
```
"rule" : "gazetteer++simple-np",
```
Does it ring a bell at all?

BeckySharp:
yes -- I would expect changes like that, as I changed the pipeline around,
and the rules are a paper trail of whose hands were involved in the
extraction.  However, the extractions themselves I would not expect to
change.
Thanks!

kwalcock:
Thanks.  It's going well so far.  Changes are limited to rules.

BeckySharp:
awesome!

kwalcock:
That was one the 10 documents.  There are discrepancies on the 52.  Details in a few hours.

kwalcock:
@BeckySharp, did I say a few *dozen* hours?  It looks like in the api2 version, the removal from `eidosSystem.annotateDoc` of the line
```
doc.sentences.foreach(addLexiconNER)
```
causes some discrepancies.  In the master version there is a DATE entity sitting around in a sentence after 'proc.annotate'.  That DATE gets changed to a B-Property after the aforementioned `addLexiconNER`.  That doesn't happen in api2.  Then later in `EidosDocument.parseTime` there is a check for that DATE which means some time stuff is not added for master, but is for api2.  In the example sentence below, the time stuff is incorrect.  The check for DATE is very strange in that it doesn't care where about the location in the sentence.

Sentence: In Rwanda, children in households that experienced positive production shocks at the time of birth had better HAZ (P < 0.01) (42).

The word "time" gets a DATE entity and then some time norms, B-Time and I-Time, are added starting at 0.01.

kwalcock:
@EgoLaparra, if you are still around, can you comment on the issue above?

kwalcock:
It looks like the code involving the DATE has been removed in Egoitz's updated version.  So, I should probably just ignore the very slight change in output.  However, it would be nice to get confirmation.

BeckySharp:
Re: DATE, idk, but good call on the property stuff.  I vote we merge for
now, bc the Property info will need to be extracted differently anyway in
the next month or so, so reproducing the master Property tags is not
necessary.

EgoLaparra:
@kwalcock, you are right. We used the DATE entities given by stanford temporal parser to filter the sentences to be processed by our parser. We have changed this to make it independent of stanford's entities.

kwalcock:
OK, there was just one more sentence among the 52 documents where it made a difference and for the record, it was this one:
```
Over 9,000 children were registered as unaccompanied, separated or missing at the time of writing.
```
The word `time` gets a DATE and in master and this is later turned into B-Property while in api_2 it remains and is processed for time.

There are a couple of cases where the order of output changed, but I filed an issue against myself for that and will fix it when things are quieter.

BeckySharp:
awesome!

PR #578, opened 2019-05-16
Get the conceptExpander from eidosConf rather than config

kwalcock:


kwalcock:
It looked like the conceptExpander wasn't being used.  It seems to be located not in the top level config but in the EidosSystem within that, which is stored in eidosConf.  At least with this change that useful warning about "NOTICE: You're keeping stateful Concepts but didn't load an expander." went away.  It may be causing other problems, though.

BeckySharp:
thanks!!!!!!!!!!!!!!!!!!!!!!!!!!

PR #579, opened 2019-05-21
Time parser with flair embeddings using tensorflow

EgoLaparra:


kwalcock:
I'm seeing cases of
```
"lemma" : "2,000", "norm" : "2000.0"
"lemma" : "mm", "norm" : "I-Time"
"lemma" : "per", "norm" : "I-Time"
"lemma" : "year", "norm" : "I-Time"
``
that I don't remember before.  Is it OK to have I-Time without a B-Time?

**Comment by kwalcock** [comment]
For the tokenized sentence
```
"Rainfall within the country ranges from 500 mm to 2,000 mm per year , with the highest levels of rainfall in the southern part of the country .
```
there is a timex
```
      "timexes" : [ {
        "@type" : "TimeExpression",
        "@id" : "_:TimeExpression_12",
        "startOffset" : 1545,
        "endOffset" : 1551,
        "text" : "000 mm",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_12",
          "duration" : 0
        } ]
      }, {
```
Is it going to mess up anything to have offsets that don't match word boundaries?

**Comment by EgoLaparra** [comment]
No, there should not be any I-Time without B-Time. Let me debug that example. Is there any process that writes in "norm" after the temporal parser?

**Comment by kwalcock** [comment]
Not that I'm aware of.  I think the problem is in
```
if (wordStart == timeExpressions(matchIndex)._1._1) "B-Time"
```
Because the start of the time expression does not match the start of any word, the B-Time will not be used.  The word starts with "2,000" and the time expression with "000".

When possible, it is sometimes good to change things like ._1._1 to something more descriptive.

**Comment by EgoLaparra** [comment]
Yes, the problem comes from the fact the our parser annotates character by character, so the time expression could have boundaries that don't match the word boundaries.  I'm not sure what is the best way to annotate these cases. We could add the B-Time label to the token that wraps the start of the time expression:
```
"lemma" : "2,000", "norm" : "B-Time"
"lemma" : "mm", "norm" : "I-Time"
"lemma" : "per", "norm" : "I-Time"
"lemma" : "year", "norm" : "I-Time"
```

But the time expression would still have the same startOffset:
```
      "timexes" : [ {
        "@type" : "TimeExpression",
        "@id" : "_:TimeExpression_13",
        "startOffset" : 1545,
        "endOffset" : 1560,
        "text" : "000 mm per year",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_13",
          "duration" : 31536000
        } ]
      }, 
```

**Comment by kwalcock** [comment]
Probably.  Does @BeckySharp have an opinion?  It might be helpful to use org.clulab.struct.Interval and overlaps.  If or some reason the time interval winds up starting with a space, it would fall outside of all words.  Perhaps the first time has to start between the end of the previous word and the end of the current word.  That would use up all the space between the words, between the previous end and the current start, just to be careful.

**Comment by kwalcock** [comment]
This isn't a programming error at all, but I find it interesting that in
```
        "@type" : "TimeExpression",
        "@id" : "_:TimeExpression_26",
        "startOffset" : 3131,
        "endOffset" : 3154,
        "text" : "four months of the year",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_26",
          "duration" : 31536000
        } ]
```
the duration is for the entire year, not just the four months.  The related
```
        "@type" : "TimeExpression",
        "@id" : "_:TimeExpression_24",
        "startOffset" : 3125,
        "endOffset" : 3142,
        "text" : "first four months",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_24",
          "duration" : 0
        } ]
```
gets found, but the duration is not helpful.

I do wonder if we need something other then 0 to indicate an unknown duration, if that's what's happening.

**Comment by kwalcock** [comment]
Here's an example that *might* indicate a problem.  The master version knows how long July is
```
        "@type" : "TimeExpression",
        "@id" : "_:TimeExpression_1",
        "startOffset" : 12,
        "endOffset" : 16,
        "text" : "July",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_1",
          "duration" : 2678400
        } ]
```
but the new version doesn't:
```
        "@type" : "TimeExpression",
        "@id" : "_:TimeExpression_1",
        "startOffset" : 12,
        "endOffset" : 16,
        "text" : "July",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_1",
          "duration" : 0
        } ]
```
Is this expected?

**Comment by kwalcock** [comment]
This seems to be working on Windows with no issues at all.  That's great.  On the collection of 52 documents, runtime without timenorm is about 5 minuts and runtime with is 66 minutes, or about 13 times as long.  Is that about what you expect?  It does a very good job keeping my processor(s) busy at nearly 100%.

**Comment by kwalcock** [comment]
@EgoLaparra, will you have time to make the B-Time changes?  I think that @BeckySharp would agree to start a B-Time even in the middle of the first word.

**Comment by EgoLaparra** [comment]
- Yes, I will do it. 
- I've done a small change in our parser related with the translation of words ("two") into numbers ("2"), so I have to deploy the parser again.
- I am also doing some benchmark because 66 minutes is the double that I expected for the 52 documents.

Hopefully, I'll have everything ready during the weekend.

**Comment by EgoLaparra** [comment]
By the way, @kwalcock, does it take 66 minutes including geonorm?

**Comment by kwalcock** [comment]
No, just with time.

**Comment by BeckySharp** [comment]
> @EgoLaparra, will you have time to make the B-Time changes? I think that @BeckySharp would agree to start a B-Time even in the middle of the first word.

If you mean that the whole word should be `B-Time` even if it starts n-characters into the word, then yes i for sure agree :)  The Norm field is at the token level so we should use it at that level

**Comment by EgoLaparra** [comment]
Done with my changes. @kwalcock, timenorm should be faster now.

**Comment by kwalcock** [comment]
Thanks!  I'll look in the morning.

**Comment by kwalcock** [comment]
Thanks for the names replacing the numbers and for dealing with the merge conflicts.  I notice the nice abbreviations for the months.  Do you ever run into the days of the week being abbreviated?  Is the main way to affect performance to adjust the context_window_size?

**Comment by EgoLaparra** [comment]
Days abbreviations are not very frequent in our training set, so I'm not sure how well the parser will handle them.

The biggest bottleneck were a set of regular expressions that took a lot of time in some documents with several consecutive newline characters. But the context_window_size and the batch_size have also some impact. Maybe it would be a good idea if they could be set in "eidos.conf", right?

**Comment by kwalcock** [comment]
One fairly easy change might be to transform
```
(?i)\b\w*spring\w*\b
(?i)\b\w*summer\w*\b
(?i)\b\w*fall\w*\b
(?i)\b\w*autumn\w*\b
(?i)\b\w*winter\w*\b
```
into
```
(?i)\b\w*(?:spring|summer|fall|autumn|winter)\w*\b
```
I wonder if there is a tool that might automatically do this.

**Comment by kwalcock** [comment]
I'm reading this carefully, if only to make sure that others (and I) can understand it.  There is a notable case
```
    val sentenceMatches = sentences.map(sentence => regexs.flatMap(_.findAllMatchIn(sentenceText(sentence))).sortBy(_.start))
```
in which `sentenceText(sentence)` is called for every regex in every sentence.  The sentenceText should probably be extracted once per sentence and just passed to findAllMatchIn.  Do you mind if I make some modifications in such places (i.e., that I muck with your code) with a PR for your approval?

**Comment by EgoLaparra** [comment]
Go ahead!

**Comment by BeckySharp** [comment]
Eidos.conf sounds like a good idea! (And reference.conf too)

**Comment by kwalcock** [comment]
I wonder how often either of the two parameters would ever be changed.  I'm not averse to having a few named constants (not magic numbers) left in the code if they actually stay constant.

**Comment by kwalcock** [comment]
The new version seems to be about twice as fast on the 52 documents:
none - 5 minutes
geo - 7 minutes
geo & time - 33 minutes

**Comment by MihaiSurdeanu** [comment]
Nice!

**Comment by EgoLaparra** [comment]
That's what I was expecting. 👍

**Comment by kwalcock** [comment]
Here's an example of how the norms and timexes can be out of sync.  For the text "the same time last year", there is

* timexes
    * the same time
    * time last
    * year
* norms
  * the B-Time
  * same I-Time
  * time I-Time
  * last I-Time
  * year B-Time

From the norms, one would only expect two timexes, neither of which are the timexes listed.  I wonder if this will confuse something downstream.  Won't it also confuse the rules in temporalAttachment.yml?  Just in case...

**Comment by EgoLaparra** [comment]
A possible solution would be to merge the intervals of the overlapping time expressions. For that case the output would be something like:
```
"timexes" : [ {
        "@type" : "TimeExpression",
        "@id" : "_:TimeExpression_1",
        "startOffset" : 0,
        "endOffset" : 18,
        "text" : "the same time last",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_1",
          "duration" : 0
        },{
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_2",
          "duration" : 0
        } ]
      }, {
        "@type" : "TimeExpression",
        "@id" : "_:TimeExpression_2",
        "startOffset" : 19,
        "endOffset" : 23,
        "text" : "year",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_3",
          "duration" : 0
        } ]
 } ]
```
_TimeInterval_1_ and _TimeInterval_2_ would come from "the same time" and "time last" respectively. The annotation would be still wrong but at least we would solve the out of sync problem.

**Comment by BeckySharp** [comment]
and we can't sync it to have a single interval for "the same time last
year"?

How frequently to adjacent time expressions truly specify different
intervals?

**Comment by kwalcock** [comment]
@EgoLaparra, maybe I broke Jenkins.  It looks like the two most recent tests are not completing, starting with "Merge pull request #582 from clulab/kwalcock-updateFlairTime" and "Add comment to force retesting".  At Jenkins I only see up to "eidos » egoitz-flairTimeTF #8" and it won't go beyond that.  Any ideas?

**Comment by EgoLaparra** [comment]
I think you are looking just the branches, have you checked the "Pull requests" tab?
http://jenkins.cs.arizona.edu:8090/job/eidos/view/change-requests/

**Comment by kwalcock** [comment]
Thanks!

**Comment by kwalcock** [comment]
That was a very strange problem on Jenkins that I don't see on the local computer.  Here goes anyway.

**Comment by EgoLaparra** [comment]
Do you get different outputs running tests in your local and in jenkins?

**Comment by kwalcock** [comment]
Yes, for some reason this line was a problems on the Jenkins configuration but not locally:
```
val timeIntervel = TimEx(Interval(0, 4), List.empty[TimeStep], "hello there")
```
Changed to
```
val timeIntervel = TimEx(Interval(0, 4), Seq.empty[TimeStep], "hello there")
````
it worked.  Perhaps something was out of sync.

Have you tried to change the $TERM on the Jenkins account?  My default on river is TERM=xterm-256color and it may be the reason for things like
```
[0m[[0m[0minfo[0m] [0m[0mLoading settings from plugins.sbt ...[0m
```

EgoLaparra:
What sbt version do you have in your computer?

There is a plugin that adds ansi color support, but I haven't tried it yet:
https://wiki.jenkins.io/display/JENKINS/AnsiColor+Plugin

PR #582, opened 2019-05-27
Try to make time more efficient

kwalcock:


kwalcock:
As usual, one thing led to another.  This gets the same answer as your original version, but it stores some intermediate results in a SentenceBundle for lack of a better name.  I still do not like updateNorms.  It does too much searching and probably gets the wrong answer if there are overlapping text intervals.  That indexWhere will only find the first match, which I'm not sure is the longest one.  More needs to be stuffed into the norm than it can hold, so maybe some rules/conventions need to be made.  That might change the output, so I'm skipping it for now.

EgoLaparra:
Thanks for making everything so readable! For updateNorms we could use a filtered version of multiTextIntervalAndTimeIntervals with the overlapping time expressions filtered out. What do you think?

kwalcock:
I'm away from the computer for several hours.

kwalcock:
Something like that might be good.  I don't think it should hold back your PR.  Maybe others would have an opinion about it that could be discussed at a weekly meeting.  I'm just going to run a few more tests before merging the big one.

PR #590, opened 2019-05-30
Configure wm_metadata access through different repo

kwalcock:


kwalcock:
This adds one line to build.sbt
```
    "com.github.kwalcock" % "Ontologies" % "master-SNAPSHOT",
```
to fetch a jar containing /org/clulab/wm/eidos/english/ontologies/wm_metadata.yml that can be used in eidos more or less as if it was really in the local /src/main/resources directory.

The location contains a forked version of the WorldModelers/Ontologies project with three small changes:
* a fairly simple build.sbt has been added
* the README.md file was updated with a jitPack icon
* the repo was configured for use with jitPack

The last one might be the sticking point.  Someone with sufficient privileges needs to arrange this.

There are options other than "master-SNAPSHOT", and one might want to use a tag or release instead of whatever happens to be in master.

BeckySharp:
@kwalcock is there a way to do this without the fork?  as in if we configure the real repo with the jitpack bells and whistles and the build.sbt, could we use it directly?  would we want to do that anyway? or does that stuff not belong in a shared repo...? (ping @MihaiSurdeanu too)

BeckySharp:
btw -- YAY!

kwalcock:
Yes.  I didn't explain that well.  Can you do to the real repo what I did on the fork?  I could try to PR the fork, actually, if that's the thing to do.  I'm probably not allowed to hook it to jitPack, though.

kwalcock:
I PR'd the fork to WorldModelers/Ontologies last night in case it is helpful.  A tool like http://www.yamllint.com/ might likewise be useful.  The original wm.yml appears to be valid even while wm_metadata.yml is not.  Does this imply a problem with strip_metadata.py?

There is a test TestDomainOntology, but it would not have caught the problem, because the file names are hard-coded there and the new one was not included.  It probably does this so that eidos doesn't have to be loaded, which couldn't happen if there's a problem with the ontologies that eidos uses.  There's a PR in the works with this test added.

FYI @vanh17

BeckySharp:
awesome, thanks! and yes I like the idea of a linter for the yaml!
Especially since others are potentially using this ontology.  Can we set it
up as a github PR-hook-thing? (kinda like we do with travis chez nous?)

kwalcock:
Well, probably, but I'd do at least two other things first: see if build.sbt and the connection to jitPack work at all on the Ontologies repo, and then modify strip_metadata.py to at least first do a formal parse of wm.yml to make sure it's a valid file.  For bonus points it could also check the resulting file.  The eidos unit test for the ontologies just uses org.yaml.snakeyaml.Yaml which will complain on a malformed file and then it also double checks for duplicates, etc.  An internal test would be more feasible than an online (page or service) for linting.

BeckySharp:
Fair enough. I am trying to get ready for the demos, but i’ll try to work
some on this soooooooon

kwalcock:
This now fetches the wm ontology from WorldModelers/Ontologies which has been attached to jitPack.

kwalcock:
I don't think that eidos is configured to use this ontology right now, but it is being tested, and so apparently fetched properly.

BeckySharp:
Should we also point eidos.conf to point to this ontology?

kwalcock:
Yes, it can be added back in now.

PR #591, opened 2019-05-30
Add test for new wm_metadata.yml

kwalcock:


kwalcock:
Only merge if the ontology has been fixed so that this passes.  It uses the local copy.

kwalcock:
Fan implemented a library that detects the factuality of a predicate using DyNet. The resulting model files are too large and have to be stored separately. Currently Fan is storing them as regular files, but I would like to package them in a jar so we can manage dependencies easier. The problem is that DyNet does not seem to allow the loading of a file from a resource in the classpath. For reference, the method Fan uses (which was copied from yours) is here:

https://github.com/clulab/factuality/blob/master/src/main/scala/org/clulab/factuality/Factuality.scala#L439

This relies on ModelLoader, which seems to accept only a file name:

https://github.com/clulab/factuality/blob/master/src/main/scala/org/clulab/factuality/Factuality.scala#L461 

Can you think of a fix? Can we modify ModelLoader in DyNet? Alternatively, we could write code to pull the content of the “.rnn” file from the jar and store it as a regular file, but that is super ugly.

kwalcock:
Right, so DyNet being written in C++ doesn't know what to do with a
Java .jar file at all.  DyNet gets the file contents via an ifstream
(input file stream) from the operating system:
std::ifstream datastream(dataname, std::ios_base::in | std::ios_base::binary);

It looks like there *might* be some C++ libraries that can process a
zip file as an ifstream.  In Scala/Java we'd have to figure out which
.jar file the resource was stored in and pass that file name, almost
as usual, plus the name of the resource, so
ZipModelLoader(findJarFor("something.rnn"), "something.rnn").  C++
would need a new class like ZipFileLoader, and that doesn't seem too
difficult.  The DyNet project has already been forked and all.
Getting access to ZipFileLoader up through the Java and Scala layers
would hopefully only require some copy, paste, and rename operations.
It seems reasonable if the C++ library can be found, and it's 64-bit,
and it's cross-platform, etc.

How big is this model?  There has not yet been any reaction to the new
issue about large models: https://github.com/clab/dynet/issues/1558

PR #593, opened 2019-05-31
Test the migration events

kwalcock:
Here's a first draft.  Merge if desired to your branch.


BeckySharp:
well good grief @kwalcock -- that was fast!

kwalcock:
Do these need to be jsonld'd before the hackathon so that something can make use of them?  We discussed adding some kind of grouping function to combine the Odin event mention(s) into maybe a special kind of Eidos event.  I'm not sure that applies to the present design, though.  I would at least have to do some special things to export all those arguments and give them names, etc. to fit into the design described on https://github.com/clulab/eidos/wiki/Anatomy-of-Mentions.  I would probably be looking for an Event with a label of HumanMigration and then raiding and renaming the arguments, etc.

MihaiSurdeanu:
Yes, we need the json output for these events. Your assumptions are correct.
Thanks!


________________________________

kwalcock:
OK, then I'm starting an issue #594 for details of the format.

kwalcock:
These latest changes provide provisional jsonld support so that you can get see some output.  It certainly doesn't read it back in yet.

PR #597, opened 2019-06-06
Replace some bullets with asterisks

kwalcock:


kwalcock:
I'm just manually adding people (@zupon, @MihaiSurdeanu) so that they can merge it into their branch or know it has been done, in case there isn't other notification.

PR #600, opened 2019-06-10
location attachments

maxaalexeeva:
Location attachments for migration events


maxaalexeeva:
Fyi, @BeckySharp. I don't think this broke anything. The code will still need to be cleaned up, but it seems to be attaching geoloc to loc mentions alright.

BeckySharp:
nice!

PR #607, opened 2019-06-16
rules to capture the causal events found in the migration docs

BeckySharp:


kwalcock:
@BeckySharp, can you change the capitalization on TestMIgrationCause.scala.  Is that an I, l, or |?  Thanks.  Windows is really bad at fixing the case of filenames.

BeckySharp:
haha, sure, my pleasure, fix forthcoming

kwalcock:
Thanks!

PR #610, opened 2019-06-24
turns geonorm into an Eidos Finder

bethard:
This gets rid of the use of `.norms` for geo-locations, simplifies EidosDocument, and makes some grammar rules and custom actions unnecessary.

I think these changes may also allow us to remove from EidosDocument:
```
var geolocs: Option[Array[Seq[GeoPhraseID]]] = None
```
But because that would also require some serialization changes, I have not tried to do that in this pull request.


MihaiSurdeanu:
@bethard: there is a build error here.

bethard:
@MihaiSurdeanu Looks more like a Jenkins error to me, since the log shows that it never finishes checking the repository out. Not sure how to debug Jenkins. I don't seem to have access to river.

MihaiSurdeanu:
@kwalcock, @EgoLaparra: it seems Jenkins runs out of space:

Screen Shot 2019-06-24 at 9 47 53 PM

This happens on the /partition because of an over enthusiastic use of the /var directory by the Jenkins docker:



Any ideas how to fix this?

kwalcock:
It looks like the admin interface to Jenkins includes the ability to delete
things.  Some things could be deleted manually now in order to reclaim
space.  Perhaps there is a Configure option that helps this happen
automatically.  It was working until recently, so we might just have to
learn how it works and what maintenance it requires.

https://stackoverflow.com/questions/7995079/delete-a-jenkins-build-via-gui

bethard:
Maybe someone with admin privileges (e.g., @MihaiSurdeanu) can enable the "Discard Old Build plugin" https://github.com/jenkinsci/discard-old-build-plugin

EgoLaparra:
I think that, for _Multibranch Pipelines_, this can be configured in _Orphaned Item Strategy_:

https://jenkins.io/doc/book/pipeline-as-code/#orphaned-item-strategy

kwalcock:
There is at least enough room now to retest and it passed.

bethard:
So does this approach look good? Should I go ahead and do the same thing for timenorm?

bethard:
Ok. I will create a separate pull request for that, so @kwalcock can merge this one whenever he has time.

kwalcock:
Thanks!  Separating geo and time seems like a good idea.

kwalcock:
I like it, the design, staying out of the norms.  A few insignificant changes have been made.  That's really clever how you assigned to eidosDoc.geolocs before the sentenceBuffers were supplied with the phrases.  That it works surprised me.

```
      val sentenceBuffers = eidosDoc.sentences.map(_ => mutable.Buffer.empty[GeoPhraseID])
      eidosDoc.geolocs = Some(sentenceBuffers.map(_.toSeq))
```

bethard:
I was wondering if we could get rid of `EidosDocument.geolocs` entirely, but since that would require a bunch of changes to serialization, I decided not to try in this pull request.

MihaiSurdeanu:
Yes, this sounds like a good plan. Maybe @kwalcock can tackle this?

kwalcock:
As maybe implied above, does this mean that in the jsonld the Sentence.geolocs should be removed (https://github.com/clulab/eidos/wiki/JSON-LD#user-content-sentence)? 

Alternatively and maybe more likely, it might just mean that EidosDocument.geolocs be kept more or less internal to the jsonld code.  It would not appear in EidosDocument but be reconstituted for serialization without changes to the output.

In the first case, when the jsonld changes, then in the state table (https://github.com/clulab/eidos/wiki/JSON-LD#user-content-statetype) instead of

|State type|Value type|Notes|
|---|---|---|
|LocationExp|{@id(GeoLocation)}|This is an ID referring to a geographic location.|

where a reference is made to a GeoLocation (https://github.com/clulab/eidos/wiki/JSON-LD#user-content-geolocation), the actual GeoLocation is used?

TimeNorm will probably need a similar change and it might be good to do both at once.

I also wonder whether it would be best to batch it up with other changes in the output for the migration events so that consumers don't have to sync up so often with our changes.

bethard:
I was assuming JSON-LD would not change, you would just retrieve the geolocs directly, in the same way you retrieve any other `TextBoundMention`, rather than keeping an extra `.geolocs` attribute on `EidosDocument`.

kwalcock:
Thanks.  "JSON-LD doesn't change" pretty much answers it.  When the serialization code gets ahold of the locations, I think they are already converted from TextBoundMentions to Location attachments.  This probably requires me to go back through all mentions and retrieve their attachments and group them based on their sentences, similarly to what you have done.  If for some reason they aren't there, because they were never used in an attachment of a surviving Mention, they wouldn't appear in the list with each sentence.  Is that correct, @BeckySharp?  I wonder if that would be a problem.  Debugging might be made more difficult.  A couple of classes use the locations in EidosDocument.  I wouldn't be surprised to see a couple of changes in the output.

BeckySharp:
yeah, I think that sounds about right... I know that @maxaalexeeva and @zupon are using them currently for their migration stuff, but I also think they have determined a likely alternate route (which may well be cleaner).  IDK what other classes use EidosDoc locations...?

kwalcock:
Thanks!  It didn't occur to me to look at that branch.  I'll check it out just in case.

maxaalexeeva:
We use geolocs and Timex in the linkingMigrations branch. That hasn't been merged into migration_schema-zupon yet, but it will be soon.

kwalcock:
It seems like it would be better to wait on removing EidosDocument.geolocs until at least the migration is ready so that not so many things are changing at the same time, creating moving targets, etc.  If more parts of Eidos than just the jsonld is wanting to use the locations, then the code might belong somewhere else that has not yet been identified.  I'll file an issue.  How's that?

bethard:
Yes, definitely file an issue, and we can wait on this until it makes sense.

BeckySharp:
Sound like a good plan, thanks all!

Issue #611, opened 2019-06-25
Dockerfile fails to install 'oracle-java8-installer' package


adamocarolli:
The docker build fails to complete with the following error:

```
Package oracle-java8-installer is not available, but is referred to by another package.
 This may mean that the package is missing, has been obsoleted, or is only available from another source
 E: Package 'oracle-java8-installer' has no installation candidate
```

The likely cause is that the PPA used `webupd8team/java` has been discontinued permanently due to a breaking update in the Oracle JDK license on April 16th, 2019 (See http://www.webupd8.org/2014/03/how-to-install-oracle-java-8-in-debian.html)

kwalcock:
Thanks for filing the issue *and* the PR to fix it!

PR #612, opened 2019-06-25
Install OpenJDK instead of OracleJDK in docker build to avoid oracle …

adamocarolli:
### Changes
The consequential change in this PR is that instead of installing the OracleJDK the Docker container will now be installed with an OpenJDK package. This switch is necessitated by a change to Oracle JDK's license on April 16th, 2019 (See https://www.oracle.com/technetwork/java/javase/overview/oracle-jdk-faqs.html) that breaks the previously used webupd8 package (See http://www.webupd8.org/2014/03/how-to-install-oracle-java-8-in-debian.html)

### Testing this PR
After pulling this PR you can try:
```
cd Docker
docker build -f Dockerfile . -t eidos-webservice
```
Building the container will take some time. Once it completes you can try:
```
docker run -id -p 9000:9000 eidos-webservice
```
Then, you should be able to navigate to the web application in your browser at localhost:9000.

### Issue Reference

#611

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -11,16 +11,17 @@ RUN apt-get update && apt-get install -y \
     vim
 
 # Install Java
-RUN add-apt-repository -y ppa:webupd8team/java
-RUN echo "oracle-java8-installer shared/accepted-oracle-license-v1-1 select true" | debconf-set-selections
-RUN apt-get update && apt-get install -y \
-    oracle-java8-installer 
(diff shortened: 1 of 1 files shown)


kwalcock:
@brandomr, I think you invented this feature.  Is this change in line with your needs?

brandomr:
@kwalcock I heard from @adarshp about this issue originally, and @adamocarolli had a chance to resolve it--the Docker build fails without this swap so I think it's a 👍!

kwalcock:
Thanks everyone behind the scenes!

PR #616, opened 2019-07-02
Transition to a primary ontology which is no longer un but instead wm

kwalcock:
wm is fetched dynamically from github


kwalcock:
I didn't handle INTERVENTION_PLUGIN_TRIGGER, so timeout.

kwalcock:
Better now.

kwalcock:
Hey @BeckySharp, could you give this a quick look mostly to make sure I understood the task correctly?

PR #617, opened 2019-07-03
turns timenorm into an Eidos finder

bethard:
This gets rid of the use of `.norms` for times, simplifies `EidosDocument`, and simplifies some grammar rules and custom actions.

`EidosDocument` is almost empty now. We should be able to get rid of it completely after clulab/processors#335 is complete and we can use that to store the document creation time.

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -4,9 +4,7 @@ EidosSystem {
       masterRulesPath = /org/clulab/wm/eidos/${EidosSystem.language}/grammars/master.yml
          taxonomyPath = /org/clulab/wm/eidos/${EidosSystem.language}/grammars/taxonomy.yml
           hedgingPath = /org/clulab/wm/eidos/${EidosSystem.language}/confidence/hedging.txt
-        timeRegexPath = /org/clulab/wm/eidos/${EidosSystem.language}/context/timenorm-regexes.txt
-        entityFinders = ["gazetteer", "rulebased", "geonorm"]
-          useTimeNorm = true
+        entityFinders = ["gazetteer", "rulebased", "geonorm", "timenorm"]
  keepStatefulConcepts = true
 }
 
@@ -64,6 +62,10 @@ geonorm {
   geoNormModelPath = /org/clulab/geonorm/model/geonorm_model.pb
    geoWord2IdxPath = /org/clulab/wm/eidos/${EidosSystem.language}/context/word2idx_file.txt
      geoLoc2IdPath = /org/clulab/wm/eidos/${EidosSystem.language}/context/geo_dict_with_population_SOUTH_SUDAN.txt
+}
+
+timenorm {
+  timeRegexPath = /org/clulab/wm/eidos/${EidosSystem.language}/context/timenorm-regexes.txt
 }
 
 adjectiveGrounder {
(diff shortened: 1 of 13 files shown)


kwalcock:
There is at least one gotcha in there.  Processor's Interval insists that an empty interval start at 0 for some reason, and some file luckily provoked the crash.  I'll submit a PR with proposed changes for your approval.
```
require(start < end || (start == 0 && end == 0), "invalid range")
```

kwalcock:
In `val wordInterval = Interval(wordIndices.head, wordIndices.last + 1)` of TimeNormFinder.scala, wordIndices can be empty and eidos will crash.  Would you like to look into it?  This attached file in particular is a problem.  I'm running `org.clulab.wm.eidos.apps.ExtractFromDirectory txtdir jsonlddir` with the attached file in the txtdir.

[21_i8343en.txt](https://github.com/clulab/eidos/files/3370363/21_i8343en.txt)

kwalcock:
I'm confused by
```
// expand to word boundaries
start = sentence.startOffsets.map(_ - sentenceStart).iterator.dropWhile(_ < start).next
end = sentence.endOffsets.map(_ - sentenceStart).iterator.dropWhile(_ < end).next
```
which seems to me would move both start and end to the right, which doesn't seem to match `expand`.  Since there is no .prev, it might need to be something like
```
start = sentence.endOffsets.map(_ - sentenceStart).iterator.dropWhile(_ < start).next
```
but I'm not sure.

bethard:
Working on these now.

kwalcock:
Thank you!  Can you look at the line that starts like
```
case i => m.start - i
```
It was looking strange yesterday and the feeling hasn't gone away yet.

bethard:
Are you thinking it should be
```
m.start - i + separator.length
```
to avoid the whitespace?

bethard:
Oh, I see the concern. It should actually be `start + i`, right? Maybe `start + i + separator.length`?

kwalcock:
I went through most of those choices too and vote for the last one.  I'm not sure whether the NN needs to see those characters or not, so didn't want to just change it.  Pesky details!

Yesterday I saw multiple intervals that went from 431 to 429, for example, backwards, and I hope one of these changes makes them go away.  If you are already editing, you might add something like
```
          if (start >= end)
            // Account for quirk in Interval which requires empties to start at 0.
            Interval(0, 0)
          else
            // yield the context interval
            Interval(start, end)
```

bethard:
Do you know how to replicate those intervals going from 431 to 429, etc.? Those are errors in Scala code somewhere (not in the neural network itself, since it doesn't produce intervals), so they need to be fixed. I'd be hesitant to add the `if (start >= end)` workaround because it would just silence an underlying bug.

I'll push my fixes for the other issues you identified shortly.

kwalcock:
I have a short list of files that caused problems.  It started with `(start == end)` because that seemed legitimate.  Some more testing might show that > is no longer needed.  Thanks.

kwalcock:
I finally noticed the pushes and am looking at this again.  Sorry for the delay.

kwalcock:
@EgoLaparra, last call soon.  Do you want to check off on this?

EgoLaparra:
I don't think I can contribute much to these changes, so if you approve the pull request just merge it.

PR #618, opened 2019-07-05
Uses Lucene to index GeoNames for location normalization

bethard:
Geographical locations are now normalized by looking up the location text in an index of the GeoNames database (https://www.geonames.org/export/). The approach tries exact matching, string-edit distance matching, and character n-gram matching. On the SemEval 2019 Task 12 development data, this approach achieves 82.11% accuracy.

The changes also split out the GeoLocationExtractor from the GeoLocationNormalizer, so that locations found by something other than GeoLocationExtractor can be normalized separately, if needed.

Possible issues:
* The index is large, about 1.5G, and it must be on the file-system, since Lucene cannot efficiently search an index in a .jar file. Currently, the code downloads the index from clulab.cs.arizona.edu if it is not present locally. The index could conceivably be hosted on Maven Central, but that's a bit misleading, since it can't be used from a .jar, and we would still have to extract it from the .jar file to disk somewhere.
* The code is using Lucene 6 at @marcovzla's request, though I really think we should be using Lucene 8, given that Lucene is already on its third 8.x release. I haven't tested it thoroughly, but it seems like only one API has changed between 6 and 8 for the code in this pull request.


MihaiSurdeanu:
@kwalcock: what do you think?
I think wget-ing a Lucene index from some public location is not too bad. ElasticSearch in the cloud might be the ultimate solution. But it's probably overkill for this.

kwalcock:
Actually, I'm relieved to see that it doesn't actually use wget, because absent extenuating circumstances I would normally draw the line at third-party libraries and not let things depend on third-party executables.  It might be good to reuse the cache directory for the geoNamesIndexPath and perhaps I should modify the misnamed CacheOntologies, which also handles the w2v compression, to download the file so that all the preparations can be made in one fell swoop to complement the on-demand download.  I will have to look up to what extent Java can easily unzip a file.  In FatDynet it's going the other way, but maybe the functionality is there and the `"net.lingala.zip4j" % "zip4j" % "2.0.3",` wouldn't be necessary.  It looks like exciting work and I'll check it out in detail soon, make sure it runs on my computer well, verify that it survives a reasonable size corpus, etc.

bethard:
Yes, I wondered about that cache directory under ontologies. I didn't put the Lucene index there because I wasn't sure if I might mess up the directory structure or something, but an Eidos-level cache directory would be a perfect place to put the Lucene index directory.

Unzipping all the files in a zip file is possible with just the Java standard library, though it requires more code:
https://stackoverflow.com/questions/10633595/java-zip-how-to-unzip-folder
I'm fine with switching to that, as long as someone other than me gets to maintain that code. ;-)

kwalcock:
You drive a hard bargain!  Am I making the changes or are both of us? ;-)

bethard:
I don't know how `ontologies.cacheDir` is currently used to confidently modify that.

I can put in a Java version of the unzip stuff though.

kwalcock:
Sounds good!

bethard:
Done. And thanks to Java NIO, it's much less ugly than the StackOverflow answers online would have me believe.

kwalcock:
Nice!  Thanks for the utility functions as well.  I'll change the cache directory and other things in not too long.

bethard:
Great. Just pushed one last update, as I realized my URL-reading code could also be greatly simplified using Java NIO.

kwalcock:
Is this index multi-lingual?  Should it be in an English subdirectory because there would be a different one for different languages?  The cache directory is split up into language subdirectories and I wonder whether it should go into one or stay a level above those distinctions.

bethard:
The index is actually multilingual, since GeoNames is, and I'm not filtering anything. I haven't tested any of the non-English names in there though.

kwalcock:
@bethard, can you check out the error below that appears when processing the file attached?  It looks like there are about half a dozen files that might be able to cause this branch to crash.  The text "/doi:10" is part of the document, so I wonder whether the queries to lucene need to be escaped.  I can get more details and examples, but figure you might see right away what's going wrong.  The line numbers might not be right because I have some edits to PR to you.  I'm again using ExtractFromDirectory with this single file in the directory.  Thanks!

[24_pnas.201409769.txt](https://github.com/clulab/eidos/files/3387648/24_pnas.201409769.txt)

Exception in thread "main" org.apache.lucene.queryparser.classic.ParseException: Cannot parse 'www.pnas.org/lookup/suppl/doi:10': Lexical error at line 1, column 33.  Encountered: <EOF> after : "/doi:10"
	at org.apache.lucene.queryparser.classic.QueryParserBase.parse(QueryParserBase.java:116)
	at org.clulab.wm.eidos.context.GeoNamesSearcher.apply(GeoNamesIndex.scala:121)
	at org.clulab.wm.eidos.context.GeoLocationNormalizer.scoredEntries(GeoNorm.scala:280)
	at org.clulab.wm.eidos.context.GeoLocationNormalizer.apply(GeoNorm.scala:304)
	at org.clulab.wm.eidos.context.GeoNormFinder.$anonfun$extract$7(GeoNorm.scala:101)
	at scala.collection.TraversableLike$WithFilter.$anonfun$map$2(TraversableLike.scala:739)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:59)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:52)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
	at scala.collection.TraversableLike$WithFilter.map(TraversableLike.scala:738)
	at org.clulab.wm.eidos.context.GeoNormFinder.$anonfun$extract$5(GeoNorm.scala:96)
	at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:241)
	at scala.collection.Iterator.foreach(Iterator.scala:929)
	at scala.collection.Iterator.foreach$(Iterator.scala:929)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1417)
	at scala.collection.IterableLike.foreach(IterableLike.scala:71)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:70)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
	at scala.collection.TraversableLike.flatMap(TraversableLike.scala:241)
	at scala.collection.TraversableLike.flatMap$(TraversableLike.scala:238)
	at scala.collection.AbstractTraversable.flatMap(Traversable.scala:104)
	at org.clulab.wm.eidos.context.GeoNormFinder.extract(GeoNorm.scala:93)
	at org.clulab.wm.eidos.EidosSystem.$anonfun$extractFrom$1(EidosSystem.scala:221)
	at org.clulab.wm.eidos.EidosSystem.$anonfun$extractFrom$1$adapted(EidosSystem.scala:220)
	at scala.collection.immutable.List.foreach(List.scala:389)
	at org.clulab.wm.eidos.EidosSystem.extractFrom(EidosSystem.scala:220)
	at org.clulab.wm.eidos.EidosSystem.extractFromDoc(EidosSystem.scala:186)
	at org.clulab.wm.eidos.EidosSystem.extractFromText(EidosSystem.scala:174)
	at org.clulab.wm.eidos.apps.ExtractFromDirectory$.$anonfun$new$2(ExtractFromDirectory.scala:27)
	at org.clulab.wm.eidos.apps.ExtractFromDirectory$.$anonfun$new$2$adapted(ExtractFromDirectory.scala:23)
	at org.clulab.wm.eidos.utils.Closer$.liftedTree1$1(Closer.scala:20)
	at org.clulab.wm.eidos.utils.Closer$.autoClose(Closer.scala:19)
	at org.clulab.wm.eidos.utils.Closer$AutoCloser.autoClose(Closer.scala:60)
	at org.clulab.wm.eidos.apps.ExtractFromDirectory$.$anonfun$new$1(ExtractFromDirectory.scala:23)
	at org.clulab.wm.eidos.apps.ExtractFromDirectory$.$anonfun$new$1$adapted(ExtractFromDirectory.scala:20)
	at scala.collection.parallel.mutable.ParArray$ParArrayIterator.foreach(ParArray.scala:142)
	at scala.collection.parallel.ParIterableLike$Foreach.leaf(ParIterableLike.scala:970)
	at scala.collection.parallel.Task.$anonfun$tryLeaf$1(Tasks.scala:49)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
	at scala.util.control.Breaks$$anon$1.catchBreak(Breaks.scala:63)
	at scala.collection.parallel.Task.tryLeaf(Tasks.scala:52)
	at scala.collection.parallel.Task.tryLeaf$(Tasks.scala:46)
	at scala.collection.parallel.ParIterableLike$Foreach.tryLeaf(ParIterableLike.scala:967)
	at scala.collection.parallel.AdaptiveWorkStealingTasks$WrappedTask.compute(Tasks.scala:149)
	at scala.collection.parallel.AdaptiveWorkStealingTasks$WrappedTask.compute$(Tasks.scala:145)
	at scala.collection.parallel.AdaptiveWorkStealingForkJoinTasks$WrappedTask.compute(Tasks.scala:440)
	at java.util.concurrent.RecursiveAction.exec(RecursiveAction.java:189)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
Caused by: org.apache.lucene.queryparser.classic.TokenMgrError: Lexical error at line 1, column 33.  Encountered: <EOF> after : "/doi:10"
	at org.apache.lucene.queryparser.classic.QueryParserTokenManager.getNextToken(QueryParserTokenManager.java:1119)
	at org.apache.lucene.queryparser.classic.QueryParser.getToken(QueryParser.java:854)
	at org.apache.lucene.queryparser.classic.QueryParser.jj_3R_3(QueryParser.java:668)
	at org.apache.lucene.queryparser.classic.QueryParser.jj_3_2(QueryParser.java:697)
	at org.apache.lucene.queryparser.classic.QueryParser.jj_2_2(QueryParser.java:653)
	at org.apache.lucene.queryparser.classic.QueryParser.Query(QueryParser.java:281)
	at org.apache.lucene.queryparser.classic.QueryParser.TopLevelQuery(QueryParser.java:215)
	at org.apache.lucene.queryparser.classic.QueryParserBase.parse(QueryParserBase.java:111)
	... 50 more

bethard:
Interesting. The code is already escaping all the characters that are special to Lucene queries:
https://github.com/clulab/eidos/blob/geonames-index/src/main/scala/org/clulab/wm/eidos/context/GeoNamesIndex.scala#L110
A debugger also shows that the query string being sent to Lucene is:
```
"www.pnas.org\/lookup\/suppl\/doi\:10"
```
So it looks like everything is being escaped. Let me dig a little more and see if I can figure out what's going wrong.

src/main/scala/org/clulab/wm/eidos/context/GeoNorm.scala
```diff
@@ -26,36 +26,58 @@ object GeoNormFinder {
...
+      Files.deleteIfExists(zipPath)
     }
 
+    def isCached: Boolean = {
+      val cached = Files.exists(segmentsPath)
+
+      if (cached)
+        logger.info(s"No GeoNames index at $geoNamesIndexPath.")
+      else
+        logger.info(s"GeoNames index found at $geoNamesIndexPath.")
```
bethard: I believe these are backwards. (Only noticed it after watching the log while testing other GeoNames fixes.)

kwalcock: Thanks!

bethard:
I believe I resolved the escaping problem. As many as three queries are issued to Lucene, and I had only been escaping special characters in the first two.

kwalcock:
Thanks!  I will pull and test some more.

kwalcock:
It worked!  Will resolve conflicts soon and merge.

PR #622, opened 2019-07-12
Show how DocumentAttachment for DCT might work

kwalcock:


kwalcock:
This shows how a DocumentAttachment, specifically for DCT, might work after updating processors, which is why it won't run on Jenkins yet.  Steve will probably want to make some changes and I didn't check jsonld output yet.  However, it seems to work and the question is mostly whether the design in processors is good enough for Eidos and other clients.
I didn't try to combine the EidosAttachment for Mentions and the DocumentAttachment, but they could share the same class as far as I can tell.

bethard:
Yes, something like this should work, though the approach would need to change a bit now that timenorm is a Finder. Basically, the DCT string needs to be injected into the Document mnetions by EidosSystem before the Finders run. (The TimeNormFinder can't inject it, because only EidosSystem sees the DCT string.)

kwalcock:
Thanks.  I will update now that master's EidosDocument has changed so much.

kwalcock:
In order to get that DCT "injected" into the document without having anywhere to temporarily store it, I've made the Option[String] value an extra argument to the extract methods.  That's not optimal because all the other finders just ignore it, but it's simple and straightforward.  I'm closing this PR because it has been subsumed by another one that combines multiple branches.

```
def extract(doc: Document, initialState: State = new State(), dctString: Option[String] = None): Seq[Mention]
```

PR #623, opened 2019-07-12
Coordinate different users of cache directory

kwalcock:
Rearrange the config files so that EidosSystem (in eidos.conf) controls the base cacheDir and then things like ontologies and geonorm can (optionally) put their caches inside it.  Flesh out the caching in GeoNormFinder into a CacheManager which can be reused by CacheOntologies to update the cache.


kwalcock:
@BeckySharp is tagged as a frequent user of the caches so that she is aware of the proposed changes.

kwalcock:
Yes.  The name of CacheOntologies will probably change soon to be more general.

PR #625, opened 2019-07-16
Improve CountSpec testing

kwalcock:
It uses the actual types like CountUnit rather than common denominator strings
The apply method uses default values to shorten the specification
The text representation of CountAttachment includes both "value" and value.d for comparison


kwalcock:
The "Improve CountSpec testing" changes should help.  I'm not sure how the previous changes were merged, so there's a PR for everything.

maxaalexeeva:
@kwalcock, thanks for working on this! I have just merged and adjusted the rest of the tests to the new format. Many of the group tests seem to work. We will go through them more thoroughly in the next few days to see if the testing works as intended.

PR #626, opened 2019-07-17
python script for pdf to txt conversion

EgoLaparra:


kwalcock:
Thanks, @EgoLaparra.  I'll be trying it out fairly soon.

kwalcock:
At some time I will make the work under Windows, which it doesn't, but it does seem to run fine under Ubuntu.  Not sure about Mac.

EgoLaparra:
The script includes a SIGALARM signal to stop the conversion of a document when it takes too long. If that part is removed the script should work in Windows.

kwalcock:
I did to that, but then there were still strange errors.  Some other people need to see the script already, so I had to check it in.  I can circle back for it.

```
Converting in\1_IPC_Alert_9_SouthSudan_Sept2017.pdf - Fail: 'charmap' codec can't encode character '\uf0b7' in position 466: character maps to <undefined>
Converting in\2_IFPRI.pdf - Fail: 'charmap' codec can't encode character '\u2019' in position 844: character maps to <undefined>
```

EgoLaparra:
Can you try changing the encoding of the output file? 
`with open(filedirout, 'w', encoding='utf-8') as outfile:`

kwalcock:
I had done that as well before the task was interrupted.  I'm just running on Linux instead for now.

kwalcock:
@EgoLaparra, you were correct about the `, encoding="utf-8"`!  The problem was that I additionally added `print(text)` and that caused a similar encoding problem with stdout which needed to be corrected instead with `set PYTHONIOENCODING=UTF-8` or by simply removing it.  I just now added conditionals around the alarms so that Windows will simply skip them.  Pull request momentarily...

PR #627, opened 2019-07-18
Document source of html2text script

kwalcock:


kwalcock:
I didn't include the actual source code because of potential license issues.

PR #628, opened 2019-07-19
Uses geonorm from Maven Central.

bethard:
The geonorm library has now been moved to a separate repository, so revise the Eidos code to just pull that library from Maven Central. Renamed `GeoNorm.scala` to `GeoNormFinder.scala` since that's the only thing it still contains.


kwalcock:
```
crossScalaVersions := Seq("2.11.11", "2.12.4")
"org.clulab"    %% "geonorm"                  % "0.9.5",
```
implies that there is a geonorm_2.11, but I can't find one (http://central.maven.org/maven2/org/clulab/).  We haven't built, tested, or published Eidos for 2.11 recently, but I don't think we've given up on it yet.  Can one be made, @bethard or @EgoLaparra?

bethard:
I have now released geonorm v0.9.6 built on 2.11, 2.12, and 2.13. They are [released on OSS Sonatype](https://oss.sonatype.org/service/local/repositories/releases/content/org/clulab/) but there is [currently some slowness in the Maven Central sync that Sonatype is trying to track down](https://issues.sonatype.org/projects/OSSRH/issues/OSSRH-52679?filter=addedrecently). Once that is resolved, you should be able to just update to 0.9.6 and use any of those versions of Scala.

PR #629, opened 2019-07-19
Account for multiple periods in filename

kwalcock:


kwalcock:
@brandomr, you'll want to add this change if the files you use are similarly named to what I received recently.

PR #636, opened 2019-07-31
Remove EidosDocument class (but not file)

kwalcock:


kwalcock:
This is a first draft of code with EidosDocument removed.  There is at least one problem that I think needs to be fixed.  In EidosSystem we're accepting an Option[String] for dctString that might come in from metadata.  In Document, we're using a DctDocumentAttachment which holds a DCT.  The DCT seems correct, especially since I don't want to have to run a string there through the time norm parser after reading an already processed document from file.  Right now the conversion from String to DCT happens when the TimeNormFinder finds/extracts.  In order to store the string until then and to convey it to the TimeNormFinder, a FinderArguments is used.  That's not tragic, but it seems better to have EidosSystem convert the string to a DCT and stick it directly into the Document before the TimeNormFinder runs.  This would remove the need to FinderArguments and it might also help with the problem that "invalid" dctStrings can come from metadata and cause an exception far into the extraction process.  (They may only be invalid because the parser does not understand them.)  For EidosSystem to do this conversion, TimeNormFinder.find would need to be torn apart to provide a function that processes only the DCT.  It seems like a useful function anyways.  Comments, particularly from @bethard?

kwalcock:
Tests won't pass until I release an updated processors with the DocumentAttachment included.

bethard:
You could just make the `parser: TemporalNeuralParser` instance variable public. Then you can create a DCT in `EidosSystem` pretty much the same way the code does it now in `TimeNormFinder`:
```
val Array(dct: TimExInterval) = parser.parse(dctString)
// use a SimpleInterval (character span = None) so it doesn't interfere with character span of TimeExpressions
DCT(SimpleInterval(dct.start, dct.end), dctString)
```

kwalcock:
Right, it's just a couple of lines of code.  I'd put them in a separate method of TimeNormFinder and have EidosSystem call it if a TimeNormFinder has been configured.  We sometimes also run into actual LocalDateTime and it will be good if something else can poke them into the document as DCTs for TimeNormFinder to read without needing to parse text.  I'll go ahead with some changes.

kwalcock:
That's my suggestion.  @bethard may want to check TimeNormFinder.

src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
```diff
@@ -149,26 +143,38 @@ class EidosSystem(val config: Config = EidosSystem.defaultConfig) {
...
-      documentCreationTime: Option[String] = None,
+      dctStringOpt: Option[String] = None,
       filename: Option[String] = None): AnnotatedDocument = {
+    // It is assumed and not verified that the document _has_ already been annotated.
+    require(doc.text.isDefined)
+
+    doc.id = filename
+    if (dctStringOpt.isDefined && timeNormFinderOpt.isDefined) {
+      val dctString = dctStringOpt.get
+      val dctOpt = timeNormFinderOpt.get.parseDctString(dctString)
```
bethard: I would write the above three lines as:
```
for (dctString <- dctStringOpt; timeNormFinder <- timeNormFinderOpt) {
  val dctOpt = timeNormFinder.parseDctString(dctString)
```

src/main/scala/org/clulab/wm/eidos/context/TimeNormFinder.scala
```diff
@@ -30,149 +43,187 @@ object TimeNormFinder {
...
 }
 
 class TimeNormFinder(parser: TemporalNeuralParser, timeRegexes: Seq[Regex]) extends Finder {
-
   private val CONTEXT_WINDOW_SIZE = 20
   private val BATCH_SIZE = 40
 
   def parseBatch(text: String, spans: Array[(Int, Int)],
-                 textCreationTime: TimExInterval = UnknownInterval()): Array[Array[TimeExpression]] = {
+    textCreationTime: TimExInterval = UnknownInterval()): Array[Array[TimeExpression]] = {
```
bethard: Was this de-denting intentional?

src/main/scala/org/clulab/wm/eidos/context/TimeNormFinder.scala
```diff
@@ -30,149 +43,187 @@ object TimeNormFinder {
...
+      val simpleInterval = SimpleInterval(timeExInterval.start, timeExInterval.end)
 
-        // merge overlapping contexts
-        val mergedIntervals = singleIntervals.sorted.foldLeft(Seq.empty[Interval]) {
-          case (Seq(), context) => Seq(context)
-          case (init :+ last, context) if last.end > context.start => init :+ Interval(last.start, context.end)
-          case (contexts, context) => contexts :+ context
-        }
+      DCT(simpleInterval, dctString)
+    }
```
bethard: Yes, that's a leap of faith. There's no guarantee it returns an `Interval`. For example, if the DCT string is just "weeks", it will return a `RepeatingInterval`. I would write this instead as something like:
```
parser.parse(dctString).collectFirst { case i: TimExInterval => DCT(SimpleInterval(i.start, i.end), dctString) }
```
Depending on how careful you want to be, you may also want to consider a `Try` around that `.parse`, since in some weird situations the machine learning model produces an invalid composition of time operators. I think @EgoLaparra has an example of where this happens, since he had to add a `Try` in `parseBatch`.

src/main/scala/org/clulab/wm/eidos/context/TimeNormFinder.scala
```diff
@@ -30,149 +43,187 @@ object TimeNormFinder {
...
+        (sentenceIndex, (interval.start, interval.end))
+      }
+    }
+    val (contextSentenceIndexes: Array[Int], contextSpans: Array[(Int, Int)]) = sentenceContexts.flatten.unzip
+
+    // run the timenorm parser over batches of contexts to find and normalize time expressions
+    val dctOpt = DctDocumentAttachment.getDct(doc)
+    val contextTimeExpressions: Array[Array[TimeExpression]] = dctOpt match {
+      case Some(dct: DCT) =>
+        val simpleInterval = dct.interval.asInstanceOf[SimpleInterval]
```
bethard: This cast should not be needed.

src/main/scala/org/clulab/wm/eidos/extraction/Finder.scala
```diff
@@ -7,7 +7,8 @@ import org.clulab.processors.Document
 import org.clulab.wm.eidos.context.{GeoNormFinder, TimeNormFinder}
 
 trait Finder {
-  def extract(doc: Document, initialState: State = new State()): Seq[Mention]
+  // This last argument is optional because not all finders need it and that way a simple default value can be provided.
```
bethard: This comment is presumably no longer relevant?

src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
kwalcock: I generally don't like `for`, but you have convinced me in this case!  I don't know how that could have happened.

bethard: =) You could also write it using the equivalent method calls. Something like:
```
val dctOpt = dctStringOpt.flatMap(dctString => timeNormFinder.flatMap(_.parseDctString(dctString))
```
Not sure I've got all my maps flatMaps right there, but hopefully you get the idea.

src/main/scala/org/clulab/wm/eidos/context/TimeNormFinder.scala
kwalcock: Doubtful but not out of the question.  I would have double indented it as a continuation line, so maybe it was the IDE.  Will change it back just to be nice.

src/main/scala/org/clulab/wm/eidos/extraction/Finder.scala
kwalcock: Indeed, but it was so very insightful that I unconsciously could not part with it.  Gone.

src/main/scala/org/clulab/wm/eidos/context/TimeNormFinder.scala
kwalcock: Thanks!  That's a remnant of the old code.  Gone.

src/main/scala/org/clulab/wm/eidos/context/TimeNormFinder.scala
EgoLaparra: In my case, I found that`Season-Of-Year` entities like "rainfall" produce compositions that are not necessarily invalid but cannot be interpreted. However, what @bethard says can happen if the parser misses the correct values of the components of the time expression.

kwalcock: Great catch!  The headOption took care of one exception, but if more are possible, I should deal with them here.

EgoLaparra: An example that could occur as a DCT: `18 Maya 2018`. The parser identifies `Maya` as a `Month-Of-Year` but is obviously a wrong value.

kwalcock: I added a Try and pushed the changes.  There will be at least one more push which updates the processors version.  You comments have been helpful, so please continue.

PR #640, opened 2019-08-07
Move elasticsearch to a different project

kwalcock:
The Elasticsearch app depended on Lucene, a different version than was being used in Eidos.  Separating it into a different project allows the two to use different versions.  The AWS and Elasticsearch.org dependencies also unnecessarily bloated Eidos and caused classpath overruns.

I was able to use `% luceneVer force()` to keep Lucene at 6.6.6 and compilation was OK, but there were runtime complaints.  Given all the other problems, I decided that a separate project was the way to go.


kwalcock:
Tagging @bethard and @MihaiSurdeanu who will certainly say if this is the wrong way to go about it and also so that they know of the subproject.  This change means that Eidos (root) is back to Lucene 6.6.6 and will get slightly different geoIDs than master recently produced.

kwalcock:
@bgyori, this will affect you, because after this is merged, you will need Lucene 6.6.6 instead of the 7.0.0 that you may have just configured.  At least it should again match what is specified in build.xml.

bethard:
Does moving it to a different repository really solve the problem? Won't both versions of Lucene just end up on the classpath together and cause problems when some future project depends on both Eidos and elasticsearch?

For geonorm, it would be pretty easy to upgrade to Lucene 7. I would like to upgrade to Lucene 8 actually. The only reason geonorm is on Lucene 6 is because @marcovzla was using that for odinson and I was trying to avoid exactly this kind of classpath conflict.

kwalcock:
I didn't explain that well.  They are both still in this eidos repository, but they are separate sbt (or IntelliJ or eclipse) projects that aren't dependent upon one another.  They result in separate jar files.  The webapp is similarly a separate project, although that one does depend on the main eidos project.  This elasticsearch app runs independently of eidos and fetches all the files that eidos should eventually work on in a separate run.

bethard:
So the elasticsearch project will only ever download files that are later used by Eidos, and Eidos will never have to directly call any methods from elasticsearch? Then yes, your solution seems fine to me.

kwalcock:
Some of those will be my "famous last words", but yes, that's the intention.  I can still work on Marco, but at least this component won't cause additional complication.

PR #643, opened 2019-08-09
Prime the Webapp

kwalcock:
This should prepare the webapp to give a quick answer when it is finally initialized and primed.  Docker installation is changed to run the webapp in production mode.


kwalcock:
There would also temporarily need to be a line in DockerfileProd
```
RUN git checkout kwalcock-webappPriming
```
to test this specific branch.

kwalcock:
It seems to work in docker.  I'll wait a couple of days to hear otherwise from anyone else before merging.

BeckySharp:
does the "production" mode vs "dev" mode make it so that changes to the source code are not seen live, i.e. the webapp would have to be reloaded?

kwalcock:
We can still use webapp/run during development.  The difference is in the preparation of the docker container which uses webapp/stage.  So, yes.  The priming code does run in both cases, so there isn't an extra wait for a response to the first query if you have waited long enough for the priming.  It delays the appearance of the blank form, so checking that is slower.

kwalcock:
So there are now three docker files:
* DockerfileRun - the old Dockerfile that uses "sbt run" as before.  The image builds faster but is slower to run.
* DockerfileRunProd - one that uses "sbt runProd" which is however much slower to start and should be avoided.
* DockerfileStage - one that doesn't use sbt at runtime and doesn't need compilation, etc., and is much faster once it is built

The last one is what should be used in production.

kwalcock:
This addresses #639.

PR #644, opened 2019-08-11
Show how to add an ontology

kwalcock:
Note: This is a temporary example not to be merged into master.

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -35,12 +35,12 @@ apps {
 
 ontologies {
   // W2V
-  useW2V = false
+  useW2V = true
   //  wordToVecPath = /org/clulab/wm/eidos/${EidosSystem.language}/w2v/vectors.txt
   wordToVecPath = /org/clulab/wm/eidos/${EidosSystem.language}/w2v/glove.840B.300d.txt
 
   // Activated Ontologies
-  ontologies = ["wm", "un", "props", "mitre12", "who", "interventions"] // , "icasa", "mesh"]
+  ontologies = ["wm", "un", "props", "mitre12", "who", "interventions", "example"] // , "icasa", "mesh"]
 
   // Caching, for quick loading, language dependent
   cacheDir = ${EidosSystem.cacheDir}/${EidosSystem.language}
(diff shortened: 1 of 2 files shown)


kwalcock:
1. Add an ontology file, preferably next to the others.  In this case I made a copy of src/main/resources/org/clulab/wm/eidos/english/ontologies/un_ontology.yml named example_ontology.yml.  To help distinguish the output, I changed the first line from UN to Example.
2. Configure it in src/main/resources/eidos.conf by adding a line in the ontologies section like
```
example = /org/clulab/wm/eidos/${EidosSystem.language}/ontologies/example_ontology.yml
```
where the right side identifies the location of the file and the left side is just a name.
3. In the same file configure the use of the new named ontology by updating the line
```
ontologies = ["wm", "un", "props", "mitre12", "who", "interventions"] // , "icasa", "mesh"]
```
to include the new name:
```
ontologies = ["wm", "un", "props", "mitre12", "who", "interventions", "example"] // , "icasa", "mesh"]
```
4. Make sure that ontologies are enabled by changing `useW2V = false` to `useW2V = true`.
5. Include the vectors file as described at https://github.com/clulab/eidos#configuring.  In other words, copy glove.840B.300d.txt to src/main/resources/org/clulab/wm/eidos/english/w2v.
6. Use Eidos, for example with `sbt runMain org.clulab.wm.eidos.apps.ExtractFromDirectory ../corpora/Doc10/txt ../corpora/Doc10/jsonld`.
7. You should observe lines like these in the resulting jsonld files:
```json
    }, {
      "@type" : "Groundings",
      "name" : "example",
      "values" : [ {
        "@type" : "Grounding",
        "ontologyConcept" : "Example/entities/human/livelihood",
        "value" : 0.6632381081581116
      }, {
```

kwalcock:
@brandomr, tagging you.

kwalcock:
P.S. We're back to using Lucene 6.6.6 instead of 7.0.0.

brandomr:
@kwalcock thanks for putting this together, it actually seems more straight forward than I expected. The key thing will be getting the ontology into the appropriate format. Thank you!

PR #645, opened 2019-08-12
Change LoadableAttributes into EidosComponents

kwalcock:
Build the components under varying conditions, allow components to come from places other than the configuration file via new constructor, block access to configuration inside Eidos, have components answer rudimentary configuration questions, unify adjective grounder construction, have callers of constructors ensure they are constructing from the same configuration.


kwalcock:
Hi @BeckySharp, this is just a courtesy call so that I don't feel solely responsible for having messed up the code.  You're welcome to ignore :-)

PR #646, opened 2019-08-12
Do not use = default value in constructor

kwalcock:
This is awaiting integration testing with Indra.


kwalcock:
Please see https://github.com/sorgerlab/indra/issues/941

PR #648, opened 2019-08-14
Use now as default DCT

kwalcock:
This would need to be rethought for general use.  Change EidosSystem.defaultDctStringOpt to None to undo, if needed.


zupon:
Thanks Keith!

PR #653, opened 2019-08-25
Restore original text after filtering

kwalcock:


kwalcock:
This is the first step in the process of making a more complicated filter that will replace the add_period script.  For now it just makes sure that the document text is not changed during processing.

kwalcock:
The first unintended consequence is with the TimeNormFinder, which works off the document text.  Since the text has been restored to the original by the time TimeNormFinder gets to it, output after this merge will have changed.  I ran the 52 documents and some changes were for the better and some for the worse.  There is no guarantee that the filter needed for better processing is also better for TimeNorm.  If need be, the TimeNormFinder could filter the text in its own special way, independently.  It could, for the sake of preserving output unchanged, use the same sanitizeText as the DocumentFilter.  So, I'm satisfied with going ahead with this, but others may want to disagree.  Therefore I'm tagging a few people.

kwalcock:
This is already obsolete, but is being checked in while a processors update is in the works.

PR #659, opened 2019-08-29
Push DocumentFilter functionality down into EidosProcessor

kwalcock:


kwalcock:
This is not yet ready to merge, but is here more for intermediate review.  This branch means to replace the external add_period program to Eidos in a way that is as transparent and correct as possible.  This version removes the DocumentFilter that Eidos needed to be aware of and which couldn't reach down far enough into the document and sentences, and replaces it with a custom Processor.  This gets most of the way there, except that the raw text still sees the added period.  I need to go another step down into the tokenizer or something to make that right.

kwalcock:
The version we're going to use is one that is even more refined.  This is being closed without merging.

Issue #654, opened 2019-08-25
add_period.py changes character offsets


kwalcock:
We're quickly losing the ability to specify which pdf to text converter to use ( we have preferred pdfminer) and what kind of special post conversion (add_period.py) must be performed.  Eidos will soon receive its text as the result of an elasticsearch query which provides already converted text that likely uses a different converter (pdfbox) and no post conversion.  It does look like section headers (e.g., header\n\nparagraph) are a general problem that we will have regardless of the pdf converter and that the functionality of add_period should be retained.  It would be very nice to do away with the explicit post conversion step by incorporating it as an implicit (optional if need be) pre annotation step within Eidos.  A secondary goal of the change would be to preserve the locations of words in the text so that the offsets that are output by Eidos correspond to offsets in the original text rather than the post conversion text.

kwalcock:
Note: That below is probably wrong because the filter is applied after tokenization and before annotation.  That means that the added period will not get tokenized and that the token/word will probably have to be added manually before annotation and removed afterwards.  I'll have to figure that out.  Another option is that this filter gets applied even before tokenization unlike the present DocumentFilter.

The current regular expressions used in add_period.py are
```python
            fixed = re.sub(r'\s*\n\s*\n\s*', '.\n\n', contents)
            fixed = re.sub(r'\.\.', '.', fixed)
```
and that's perfect except that the length of the text can change after substitutions, which will invalidate character offsets into the original text.  I propose to search for something more like `\.?\s*\n\s*\n\s*`.  If there is already a period present, replace with \.\n\n\s*\s*\s* with the \s, which may be a form feed or tab, normalized to a space.  If there is no period so that we want to add one, and there is any \s* at all, substitute with space\.\n\n\s*\s*\(s-1)*.  That is, add one of the spaces (to avoid the period becoming involved with abbreviations), the period, and then any spaces short one that are left.  This is slightly different from the original version, but I don't want that period getting incorporated into a word.  If there is no period and there are no extra spaces, then sacrifice a \n and do space\.\n.

The resulting text will be sent through processors.  After that happens, there will be "words" that use the added period.  These will be filtered out.  I hope that no Eidos rules depend on that period (and will check).  After processors is finished, the original text will be returned to the Document as is now being done with the updated DocumentFilter.

Since the current DocumentFilter changes \n and \f to space anyway, the new filter may take care of that.

MihaiSurdeanu:
I agree. In general, the principles we should follow are:

[...]

PR #660, opened 2019-08-31
Add a ParagraphSplitter

kwalcock:


kwalcock:
This isn't quite production code yet, but it digs deep into processors and it would be useful to get feedback before I clean it up.  There is more information at #654.

kwalcock:
Thanks.  I think the reason for not doing it by subclassing was so that it can support Spanish and Portuguese as well.  The available pre and post processing steps I noticed were TokenizerSteps and the argument to that is an Array[RawToken] which leaves it without access to the text containing the whitespace.  I may well be wrong, of course, and suggestions welcome.  I can also show how it would look with that better access to Tokenizer arguments.  It is in a delicate place and it does seem important to get it right.

MihaiSurdeanu:
Yeah, good point about the multilingual processors. In any case, this is not what I had in mind. Take a look at the custom tokenization postprocessor in BioNlpProcessor:

https://github.com/clulab/processors/blob/master/corenlp/src/main/scala/org/clulab/processors/bionlp/BioNLPProcessor.scala#L30

kwalcock:
Yes, let's talk about it at the meeting, because I don't readily see what you mean.

In the meantime, risking infecting the itch that I'm scratching,  I'll add two other options with the issue that would be possible with slight changes to processors. #654

kwalcock:
This was done in a different way which involves changes to processors.  This PR is being closed without merging.

PR #661, opened 2019-09-03
Add the ParagraphSplitter

kwalcock:
This won't compile until processors is updated


kwalcock:
The edits are difficult to follow.  It might be easier to skip to the end.

Issue #650, opened 2019-08-22
serialization tests failing in human migration branch


maxaalexeeva:
Hi @kwalcock, 

We have gone through our migration tests in `migration_schema-zupon` and marked our failing tests as `failingTest` (we are still working on fixing those). We also fixed some of the other tests that were failing because of our migration event changes. We now have six non-migration-event failing in this branch from these two test files:

org.clulab.wm.eidos.serialization.json.TestJLDSerializer
org.clulab.wm.eidos.serialization.json.TestJLDDeserializer

Is that because migration events are not yet exported (discussed here https://github.com/clulab/eidos/issues/594)? Or are they already exported, but the tests should be modified? I have found a few spots where there errors are coming from, but I don't think I know enough about serialization to complete the job. Do you think you could have a look at that? Thank you! 

@zupon, FYI

kwalcock:
Any code that I wrote around the time of the hackathon is probably very outdated if it was even correct in the first place.  Thursday is very busy, but the issue is on the short list.

maxaalexeeva:
Got it! Thank you! No rush---the branch is not completely ready on our end either.

BeckySharp:
Yes -- that rule creates a Relation Mention -- can we add the support for
them?

kwalcock:
I haven't forgotten about this yet.

PR #668, opened 2019-09-11
Get counts and migration tests to serialize and deserialize

kwalcock:


kwalcock:
This should address #650.

kwalcock:
Jenkins is acting up.  The tests pass on my computer.  I think the problem will go away after the merge.

PR #673, opened 2019-09-19
GroundingSteps and the Handler applies them

BeckySharp:


kwalcock:
We didn't overlap too much.  I'll work on merging.  You might want to look briefly at ConceptGrounder.scala in case I'm way off course.

BeckySharp:
do we need this now?  Is it that it's supposed to identify the key
concept?  Bc that could be done in EidosOntologyGrounder?

kwalcock:
I like how you are incrementally building and replacing Mention.groundings which I was trying to create in one overly complicated fell swoop.

I am worried that the OntologyGrounding is not the right data structure to hold the information about the core concept, modifiers, what I called category, and data that we might want to keep around long enough to get it into the JSON-LD and things like that.  A different return type seems to imply a different class and a different place to store the data, etc.  One thing leads to another.

That to me is the main difference between the implementations.  What do you think?  They could be in the same class hierarchy if, for instance, they updated the grounding rather than returned it.  The update might have to be to a different variable in the EidosMention, but not too many things would have to know about that.

BeckySharp:
OK, i see your point, let me think on the data structure and we'll touch
base?

BeckySharp:
fyi I haven't forgotten about this

BeckySharp:
> core concept, modifiers, what I called category, and data that we might want to keep around long enough to get it into the JSON-LD and things like that. A different return type seems to imply a different class and a different place to store the data, etc. One thing leads to another.

Yes -- this seems important, a different data structure/class is prob good

> That to me is the main difference between the implementations. What do you think? They (the grounders) could be in the same class hierarchy if, for instance, they updated the grounding rather than returned it. The update might have to be to a different variable in the EidosMention, but not too many things would have to know about that.

I'd like to keep it with not too many types of grounders... just several of them? (i.e., indicators are just one more EidosOntologyGrounder) does that make sense?  Would you like to take a stab at it?

kwalcock:
I"ll stab away.

BeckySharp:
thanks!!!

kwalcock:
This will take at least one more step to integrate

PR #674, opened 2019-09-24
Serialization of CountAttachments

kwalcock:


kwalcock:
This is not the final version.  In fact, it is very likely to change because it is very much a hack.  It at least shows that the problem has been identified.

maxaalexeeva:
@kwalcock, yep, looking at it now. Thank you!

PR #676, opened 2019-09-25
Bsharp patch2

BeckySharp:
Fixes an issue that @adarshp found where the generated indicator map included wm -> wm mappings


kwalcock:
Only patch 2?  I thought we were on 6 at least!  Please go ahead and press the button if this should indeed go to master right now.

PR #678, opened 2019-09-26
Make MigrationTester that can turn off attachments

kwalcock:


kwalcock:
This MigrationTester may or may not be useful overall, but there is an example in the first test of TestMigrationSchema.  Here useAttachments is set to false, so it does not catch the error in the specification of group, which has 40001 instead of the correct 40000 in the attachment.  There is a newTester method which could be used to set useAttachments false globally, but that causes more problems that it could solve.  Without checking the attachments, there are often multiple matches.  In object NodeSpec there is a filter that is configered by default to be NodeSpec.firstFilter so that when there are multiple matches, the first is taken without complaint.  Without the attachment check there are more matches and often the first one is not the right one.  So, I would only use this on individual tests.  I wonder whether some of the mentions need to be filtered so that one with the most arguments kept and less complete ones dropped or something.  Let me know what other functionality might be added or corrected.

BeckySharp:
That filtering is supposed to kinda be there, but if it’s not working and
you can fix it by all means! That would be awesome!

kwalcock:
I'm not sure why the merge was from migration_schema-zupon into this one.  I had expected the other way around.  I'm going to close this PR and possibly arrange for it to be done the other way around if it is still needed.

kwalcock:
@zupon, @maxaalexeeva, this is the PR I was wondering about.  If it is still relevant, then perhaps I should resubmit it.

maxaalexeeva:
@kwalcock, based on this thing you said,

> There is a newTester method which could be used to set useAttachments false globally, but that causes more problems that it could solve. Without checking the attachments, there are often multiple matches.

I think it makes sense to keep testing with the attachments included. My thought was test just the values (with no attachments) separately from testing the attachments, but if testing with no attachments might end up not being very reliable, let's just stick, for the time being, with what we are doing now. Thank you!

PR #679, opened 2019-09-26
Updates in ontology mapping

bgyori:
This PR makes the following changes to the mapOntologies function:
- The old BBN/Hume ontology isn't relevant anymore so I removed it
- Added docstrings for the two new arguments
- Propagated the providedOntName to appear in the mapping table


kwalcock:
I'm not in the loop on the ontology development, so I'm hoping that @BeckySharp will just acknowledge the update.

PR #680, opened 2019-09-26
Update echo command in docker files

kwalcock:


kwalcock:
I was able to confirm that on Windows using docker toolbox for some reason the original echo statement results in
```
-e
play.filters.hosts {allowed = [".:9000"]}
```
being placed in the configuration file.  The new version seems to work right on both Windows and Linux.  Mac is presumed so far.

PR #681, opened 2019-09-27
Merge back branch that contains batch processing code

kwalcock:
This is code that has been used to process large collections of documents.


kwalcock:
I'm going to try to make this look simpler.

PR #682, opened 2019-09-27
Merge batch processing code

kwalcock:


kwalcock:
It isn't obvious from the list of commits above, but very few changes to master files are involved here.  These are changes from the kwalcock-17k branch which was tracking master but also included all the extra "scripts" used for batch processing of large collections of documents.  That means that there are lots of one-off things related to maximum file sizes, error handling, ad-hoc metadata conventions, etc.  However, as our metadata format solidifies (via Jataware, DART), with the advent of DocumentAttachment, because we're getting more formal data sources, since other groups are needing the code to incorporate into a larger workflow, etc., I think it's about time to merge this into master.  I'll probably be able to prune some of the files in `batch` and document the others as it becomes more clear which will stay for the longer term.  In the future I hope that only a small branch will be necessary in order to record and tag the configuration files needed for a particular run.

kwalcock:
I'm going to tag a few people so that they are aware and in case they need to say it's just a bad idea or to ask questions.  We're going to be processing some large batches in the next few weeks and this will take care of tracking with master so that it need not be done at the last minute.

kwalcock:
I get the apps mixed up myself and will add more documentation next time they are run.  They are mostly separate from others because they have to deal with metadata that we don't usually have locally, but there is also multithreading, error handling, etc.

BeckySharp:
really glad you’re on the team keith! like, so so so glad

kwalcock:
I'm going to go ahead and do this.  The code includes ontology version output that will lead to conflicts with what I'm working on now.

PR #683, opened 2019-10-02
Update DART search for recent CDR changes

kwalcock:


kwalcock:
This is a separate subproject of eidos that is just tracking a new way we have to download documents from the so-called DART pipeline.  Since it does not affect eidos proper, I am taking the liberty of checking it in without much/any review.

Issue #671, opened 2019-09-17
Compositional Grounding API


kwalcock:
This is where we can discuss this topic.

kwalcock:
@BeckySharp, I made a first draft of the grounding as a post-processing stage with an optional var containing the result as we discusses just over a week ago.  I have not yet moved the canonical name into the grounding, but given the additional grounding algorithm which will have its own data, that seems to be the right thing to do.  I'll check that nothing else uses it and head in that direction.

MihaiSurdeanu:
@bgyori: this is so we can discuss the reader API for the compositional grounding.

What we had in mind:
- We do NOT remove anything from the current output, so we maintain backwards compatibility (for better or worse). 
- For concepts that include a core concept + a property/process/phenomenon in the same phrase, we include information from compositional grounding. In particular, this will include:
1. The core concept in the phrase. For example, in the phrase "price of oil", the core concept is "oil". This core concept will be associated with top K possible groundings.
2. A list of 1 or more modifiers, where a modifier will be a property/process/phenomenon. For example, for the phrase "price of oil", the modifiers will be: Modifiers(Property("price")). Same as above, each modifier will be associated with top K groundings.

@bgyori: does this sound Ok to you?
@kwalcock: can you please think of the data structure to represent this internally?
@BeckySharp: can you please implement the internal call to the compositional grounding algorithm to produce this data?
@zupon: can you please implement the above using the syn+sem heuristics we discussed today?

Thank you all!

bgyori:
Yes, that sounds good to me!

BeckySharp:
Small note:
I think that the above "price of oil" would actually be:  Modifiers(Property("price"), Phenomenon("oil")) and theoretically you can use the "Modifiers" flavor of grounding in place of the original (assuming you have an algorithm for it)

MihaiSurdeanu:
Good catch.

kwalcock:
@BeckySharp, in your example is the CoreConcept still "oil" in addition to the Phenomenon being "oil"?  It looks like I will also have to remember whether something is a property, process, or phenomenon.  The slashes weren't just a way to describe some vague P-thing, but the specific P is valuable information?

kwalcock:
The scaffolding is in branch kwalcock-reground.  I'm removing the pull request that demos it so that it doesn't accidentally get merged into main.

BeckySharp:
sounds great!

PR #684, opened 2019-10-03
Preview of CompositionalGrounder

kwalcock:
This is not yet intended for master


kwalcock:
I'm probably going to resubmit this shortly without all the intermediate steps so that the final changes are more obvious, but one can inspect and test it already  The most important changes are to OntologyHandler.scala and OntologyGrounder.scala.  I'm not terribly happy with the change to Map[String, Seq[OntologyGrounding]) because it has some long-distance effects.  This may still be influenced by how the CompositionalGrounder really works in contrast with how I guessed it might work.

src/test/resources/englishTest.conf
```diff
@@ -59,7 +59,7 @@ actions {
 
 ontologies {
           useW2V = false
-   wordToVecPath = /org/clulab/wm/eidos/${EidosSystem.language}/w2v/vectors.txt
+   wordToVecPath = /org/clulab/wm/eidos/${EidosSystem.language}/w2v/glove.840B.300d.txt
```
BeckySharp: if we load these we'll need more mem fyi

kwalcock:
FWIW I changed my mind again and am slightly changing how these are stored in the EidosMention.  It shouldn't affect doing the actual grounding, though.

kwalcock:
I'm closing the pull request so that it doesn't get merged into main.  For further discussion, see issue #671.

Issue #665, opened 2019-09-09
Negative time intervals


bgyori:
We found that in the latest output, several "negative" time intervals are reported in TimeExpressions. For instance, 

```
113943       "timexes" : [ {
113944         "@type" : "TimeExpression",
113945         "@id" : "_:TimeExpression_80",
113946         "startOffset" : 31959,
113947         "endOffset" : 31983,
113948         "text" : "November-to-January 2013",
113949         "intervals" : [ {
113950           "@type" : "TimeInterval",
113951           "@id" : "_:TimeInterval_53",
113952           "start" : "2013-01-01T00:00",
113953           "end" : "0014-12-01T00:00"
113954         } ]
```
in `17873e4eddc4aa7bcd5146b814ce1f42.jsonld` (@kwalcock should have a copy of this file) the end time stamp is before the start time stamp. 

Another example is 

```
227492       "timexes" : [ {
227493         "@type" : "TimeExpression",
227494         "@id" : "_:TimeExpression_190",
227495         "startOffset" : 64378,
227496         "endOffset" : 64392,
227497         "text" : "until November",
227498         "intervals" : [ {
227499           "@type" : "TimeInterval",
227500           "@id" : "_:TimeInterval_134",
227501           "start" : "2014-11-01T00:00",
227502           "end" : "2014-10-02T00:00"
227503         } ]
```
from the same file. 

Question is, are these considered bugs in time expression extraction, and should be fixed at the reader level, or should we take them as is, and handle them downstream?

kwalcock:
Thanks for the sharp eyes.  Our time gurus have been assigned.

EgoLaparra:
Can you share the text file, please?

kwalcock:
Here you go.

[17873e4eddc4aa7bcd5146b814ce1f42.txt](https://github.com/clulab/eidos/files/3593026/17873e4eddc4aa7bcd5146b814ce1f42.txt)

[17873e4eddc4aa7bcd5146b814ce1f42.jsonld.txt](https://github.com/clulab/eidos/files/3593035/17873e4eddc4aa7bcd5146b814ce1f42.jsonld.txt)

[17873e4eddc4aa7bcd5146b814ce1f42.json.txt](https://github.com/clulab/eidos/files/3593029/17873e4eddc4aa7bcd5146b814ce1f42.json.txt)

bethard:
[...]

Issue #664, opened 2019-09-05
Timenorm crashes 3


kwalcock:
Again this may only be with DCT and happens at least in branch kwalcock-EidosTokenizer2Crash.  The exception is different form the other two.

[902e1c36730958d429cfde0d213f6fae.txt](https://github.com/clulab/eidos/files/3580736/902e1c36730958d429cfde0d213f6fae.txt)

```java
Exception in thread "main" java.lang.UnsupportedOperationException: AfterRI does not yet support FractionalNumber
	at org.clulab.timenorm.scate.IRIN.integer(Types.scala:386)
	at org.clulab.timenorm.scate.IRIN.integer$(Types.scala:383)
	at org.clulab.timenorm.scate.AfterRI.integer$lzycompute(Types.scala:543)
	at org.clulab.timenorm.scate.AfterRI.integer(Types.scala:543)
	at org.clulab.timenorm.scate.AfterRI.x$21$lzycompute(Types.scala:548)
	at org.clulab.timenorm.scate.AfterRI.x$21(Types.scala:548)
	at org.clulab.timenorm.scate.AfterRI.start$lzycompute(Types.scala:548)
	at org.clulab.timenorm.scate.AfterRI.start(Types.scala:548)
	at org.clulab.wm.eidos.context.TimeNormFinder.$anonfun$find$17(TimeNormFinder.scala:208)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:32)
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:29)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:191)
	at scala.collection.TraversableLike.map(TraversableLike.scala:234)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:227)
	at scala.collection.mutable.ArrayOps$ofRef.map(ArrayOps.scala:191)
	at org.clulab.wm.eidos.context.TimeNormFinder.$anonfun$find$16(TimeNormFinder.scala:183)
	at org.clulab.wm.eidos.context.TimeNormFinder.$anonfun$find$16$adapted(TimeNormFinder.scala:182)
	at scala.collection.TraversableLike$WithFilter.$anonfun$flatMap$2(TraversableLike.scala:770)
[...]

Issue #663, opened 2019-09-05
Timenorm crashes 2


kwalcock:
This crash occurs with the text file below, but only if there is a DCT present.  That comes from metadata, so no app in the main branch will readily show it.  I've modified branch kwalcock-EidosTokenizer2Crash to produce the error.  It matches what is seen in the more complicated branch.  I hope that narrows it down some.

[713d49b4131b8dd38cc6ae3f92e6307c.txt](https://github.com/clulab/eidos/files/3580680/713d49b4131b8dd38cc6ae3f92e6307c.txt)

```java
Exception in thread "main" java.lang.UnsupportedOperationException: one of List(SimpleInterval(+44100-01-01T00:00,+44200-01-01T00:00), SimpleInterval(+44200-01-01T00:00,+44300-01-01T00:00), SimpleInterval(+44300-01-01T00:00,+44400-01-01T00:00), SimpleInterval(+44400-01-01T00:00,+44500-01-01T00:00), SimpleInterval(+44500-01-01T00:00,+44600-01-01T00:00), SimpleInterval(+44600-01-01T00:00,+44700-01-01T00:00), SimpleInterval(+44700-01-01T00:00,+44800-01-01T00:00), SimpleInterval(+44800-01-01T00:00,+44900-01-01T00:00), SimpleInterval(+44900-01-01T00:00,+45000-01-01T00:00), SimpleInterval(+45000-01-01T00:00,+45100-01-01T00:00), SimpleInterval(+45100-01-01T00:00,+45200-01-01T00:00), SimpleInterval(+45200-01-01T00:00,+45300-01-01T00:00), SimpleInterval(+45300-01-01T00:00,+45400-01-01T00:00), SimpleInterval(+45400-01-01T00:00,+45500-01-01T00:00), SimpleInterval(+45500-01-01T00:00,+45600-01-01T00:00), SimpleInterval(+45600-01-01T00:00,+45700-01-01T00:00), SimpleInterval(+45700-01-01T00:00,+45800-01-01T00:00), SimpleInterval(+45800-01-01T00:00,+45900-01-01T00:00), SimpleInterval(+45900-01-01T00:00,+46000-01-01T00:00), SimpleInterval(+46000-01-01T00:00,+46100-01-01T00:00), SimpleInterval(+46100-01-01T00:00,+46200-01-01T00:00)) is outside of SimpleInterval(2019-06-05T00:00,2019-06-06T00:00)
[...]

Issue #662, opened 2019-09-05
Timenorm crashes


kwalcock:
On the kwalcock-EidosTokenizer2 branch, which will soon be merged into master, the attached file takes a very, very long time to process and then finally crashes with the stack trace below.  I'm running the `ExtractFromDirectory` app with program arguments `../corpora/DocDart/txt ../corpora/DocDart/jsonld`.  The same thing seems to happen with or without DCT.  The other problem files *seem* to only have problems with a DCT.  Can a "time person" check it out?

[63d8dfc516fb3ee1c9617667c705e43d.txt](https://github.com/clulab/eidos/files/3580443/63d8dfc516fb3ee1c9617667c705e43d.txt)

```java
Exception in thread "main" java.time.DateTimeException: Invalid value for Year (valid values -999999999 - 999999999): 1000000000
	at java.time.temporal.ValueRange.checkValidIntValue(ValueRange.java:330)
	at java.time.temporal.ChronoField.checkValidIntValue(ChronoField.java:722)
	at java.time.LocalDate.plusYears(LocalDate.java:1277)
	at java.time.LocalDate.plus(LocalDate.java:1241)
	at java.time.LocalDateTime.plus(LocalDateTime.java:1194)
	at org.clulab.timenorm.scate.IntersectionRI.$anonfun$following$8(Types.scala:860)
	at scala.collection.Iterator$$anon$9.next(Iterator.scala:161)
	at scala.collection.Iterator$$anon$10.next(Iterator.scala:448)
	at scala.collection.TraversableOnce$FlattenOps$$anon$1.hasNext(TraversableOnce.scala:464)
	at scala.collection.Iterator$$anon$15.hasNext(Iterator.scala:636)
	at scala.collection.Iterator.toStream(Iterator.scala:1403)
	at scala.collection.Iterator.toStream$(Iterator.scala:1402)
	at scala.collection.AbstractIterator.toStream(Iterator.scala:1417)
	at scala.collection.TraversableOnce.toSeq(TraversableOnce.scala:298)
	at scala.collection.TraversableOnce.toSeq$(TraversableOnce.scala:298)
	at scala.collection.AbstractIterator.toSeq(Iterator.scala:1417)
	at org.clulab.timenorm.scate.This.intervals(Types.scala:324)
	at org.clulab.timenorm.scate.This.intervals$(Types.scala:314)
	at org.clulab.timenorm.scate.ThisRI.intervals$lzycompute(Types.scala:337)
[...]

PR #685, opened 2019-10-03
timenorm 1.0.2. catching exceptions in normalization.

EgoLaparra:
- New version of the timenorm parser. It solves the crashes after long process time caused by impossible dates,  https://github.com/clulab/eidos/issues/662. It does not produce negative intervals, https://github.com/clulab/eidos/issues/665.
- A handler has been added to capture exceptions thrown during the normalization of incorrectly parsed time expressions. https://github.com/clulab/eidos/issues/663 https://github.com/clulab/eidos/issues/664


kwalcock:
Thanks!

kwalcock:
This was able to process the three problematic documents and it also worked through the collection of 52 without issue.  I'm happy.  Next time it will be about 1225.

PR #687, opened 2019-10-08
Allow polarity to be Double or Int

kwalcock:


kwalcock:
We have quite a few internal ontologies that use a Double for polarity.  Rather than changing them all to Int and maybe changing them back later, and so that we can deal with both the old and new versions of the changed ontology, I'm allowing either format for now: 1 or 1.0.

kwalcock:
@BeckySharp, in case you want to review this retroactively.

BeckySharp:
looks great, thanks!

PR #690, opened 2019-10-11
Add version information to ontologies

kwalcock:


kwalcock:
This adds a version (the github hash code) and date to each ontology and furthers the information as needed so that it appears in the json-ld output.  Unit tests may fail if the testing machine does not have access to git.

PR #695, opened 2019-10-16
Findgeo

BeckySharp:
Adds Location attachment to Location mentions found with a gazatteer, importantly this includes the geoloc lookup

FYI @bethard @EgoLaparra 
FYI @zupon -- note that I removed the generic "locations" like "settlement" -- we'll have to put those in a separate file (or handle another way) when your branch gets merged in

### DIFF
--- a/src/main/resources/org/clulab/wm/eidos/english/lexicons/Location.tsv
+++ b/src/main/resources/org/clulab/wm/eidos/english/lexicons/Location.tsv
@@ -1,21 +1,7 @@
-area
-areas
-camp
-camps
-country
-countries
 DRC
 Equatorias
 Nguenyyiel
 Pagak
-refugee camp
-refugee camps
-refugee settlement
-refugee settlements
-settlement
-settlements
-site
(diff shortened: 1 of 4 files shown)


kwalcock:
I didn't merge into master yet, but did add this to the branch that is rerunning along with the patched DCT code.

BeckySharp:
cool!

maxaalexeeva:
I merged this into migration_schema-zupon because I really really needed this. This thing has made my week, @BeckySharp!

BeckySharp:
great!

kwalcock:
I'm not sure why this had a yellow dot for so long, but resubmitting the Jenkins job fixed it.  On the one hand the code seems fine and on the other I'm not sure the output was evaluated for improved performance.  I assume that it can be quickly turned off if that's a problem, though.

In some future round perhaps miscellaneous string constants as in `_ matches "Location"` can be organized more like `m matches EidosSystem.COREF_LABEL`.

PR #701, opened 2019-10-25
Remove BuildInfoOption.BuildTime from built.sbt

kwalcock:
Add some scala switches for better error messages and then address the messages.


kwalcock:
FYI @marcovzla.  This may interest you, @Lnk2past.  It should result in fewer compiles and fewer rebuilds.

PR #703, opened 2019-10-28
bring back dep and fix tests

BeckySharp:
@kwalcock -- this will bring back some behavior we want, can we please merge ahead of the eval (i.e., very soon?) assuming things all pass here as they did locally....

fyi @MihaiSurdeanu


kwalcock:
Do I understand correctly?  Some of the values in reference.conf caused problems, but because they were not in englishText.conf, the problems weren't detected in test mode.  When the values were added to englishText.conf, some tests failed.  Three tests were updated and it was decided that "nmod:poss" should no longer be in the list.  That one is the only change that will make a difference in "run mode" and it should be used in some upcoming evaluation.

kwalcock:
Your email has arrived in the meantime and seems to match.

PR #707, opened 2019-11-20
Account for change of polarity format

kwalcock:


kwalcock:
@zupon, this isn't the problem you are looking for yet, but the code you are working on is somewhat old and it isn't accounting for the change made in the WorldModelers/Ontologies project.  Perhaps that jitpack isn't getting you the most recent version of the ontology.  This change should work with both the new and old version.

kwalcock:
This on the fly creation of rules in CompositionalGrounder.getModifierMentions can result in things like
```
   pattern: |
      trigger = °
```
which odin can't process.  Maybe you've already noticed.  Some escaping is going to be necessary, perhaps quite  a bit because valid yml is required.  I'm thinking of SQL injection kinds of problems.

zupon:
> This on the fly creation of rules in CompositionalGrounder.getModifierMentions can result in things like
> 
> ```
>    pattern: |
>       trigger = °
> ```
> 
> which odin can't process. Maybe you've already noticed. Some escaping is going to be necessary, perhaps quite a bit because valid yml is required. I'm thinking of SQL injection kinds of problems.

@kwalcock Can you elaborate on this? I don't follow how the rule in `CompositionalGrounder.getModifierMentions` will give us something like your example. What sorts of changes/fixes need done to remedy this?

kwalcock:
In this case `${synHeadWord}` must have been °.  It came from whatever test file I happened to have been using last.  Pasting it in the middle of a rule seems to result in an invalid rule as far as Odin is concerned and the program crashes.  Perhaps it doesn't deal well with unicode.

In the case of `[chunk=/NP$$/ & !word=${synHeadWord} & !tag=DT]+`, if `synHeadWord` is something like `ha & word=ha ` or just `]` probably something undesired will happen at an inopportune time.

kwalcock:
BTW it is not intended to ground during testing, especially not on Travis/Jenkins.  The test will need to be disabled later (until we figure out a good way around this).

PR #709, opened 2019-11-26
Fix collecting of EidosMentions

kwalcock:
Adjust grounding name

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/groundings/OntologyGrounder.scala
+++ b/src/main/scala/org/clulab/wm/eidos/groundings/OntologyGrounder.scala
@@ -31,6 +31,7 @@ trait OntologyGrounder {
   def name: String
   def domainOntology: DomainOntology
   def groundOntology(mention: EidosMention): Seq[OntologyGrounding]
+  def groundStrings(strings: Array[String]): Seq[OntologyGrounding]
 }
 
 abstract class EidosOntologyGrounder(val name: String, val domainOntology: DomainOntology, wordToVec: EidosWordToVec, canonicalizer: Canonicalizer)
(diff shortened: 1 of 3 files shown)


kwalcock:
@zupon, here's an update for string tests.  Let me know how close or far that is from what you need.

kwalcock:
@zupon, @BeckySharp, to ground against simple text, I avoided some things in EidosOntologyGrounder.groundOntology because they involve use of the mention type.  I can add that as a parameter that can be passed in along with the text, though.  I do see that patterns are matched against odinMention.text while wordToVec gets run on the canonicalNameParts.  This was one of my concerns with regrounding.  If Ben calls it without a proper mention, it will probably not get the same result if it was a pattern that matched when grounded previously.

BeckySharp:
OK, makes sense, I wonder if we need to receive both when he calls
"reground".... or we can parse on the fly and repeat the canonicalization?

zupon:
I added some stuff to `groundStrings` in `EidosOntologyGrounder` so that we can match the regex patterns for the `property` branch without needing to invoke any kinds of mentions. I also changed around some of how I was checking the strings, and things seem to be working better. @BeckySharp  we can look at it when we meet tomorrow (12/5) and @kwalcock I'll keep you posted on what's going on.

kwalcock:
Cool.  I was going to do something like that five minutes from now.  Instead, I'll wait and see what you came up with.

BeckySharp:
sounds great!

kwalcock:
IIRC canonicalization requires things like POS which won't be available either.
```
 def isCanonical(lemma: String, tag: String, ner: String)
```

BeckySharp:
true, though we could parse them again, or at least POS tag them with
processors....

kwalcock:
Wouldn't we need the entire sentence then?  That may not be in the mention text or canonicalName.

BeckySharp:
I think for POS tagging/NER/lemmatization we'd do decent with just the
mention text, but I can try to check empirically if you want

PR #715, opened 2019-12-05
Moving Exporters to a package, added app to reconstitute jsonld output

BeckySharp:
I think it's fairly innocuous, and i'd like to use the app for some analysis MITRE wanted


BeckySharp:
is Jenkins down still? I don't see the building/checking happening

kwalcock:
It looks like it has happened in the meantime.  Jenkins was restarted yesterday.

PR #716, opened 2019-12-05
Translate migration to canonical Scala

kwalcock:


kwalcock:
I left some things in there so that the output changes minimally.  Merged mentions, for example, lose their order and ties are less well handled.  Still in 52 documents, there were two changes of order, possibly because of the tiebreaker on orderMentions.

BeckySharp:
@kwalcock  --
@maxaalexeeva  and I just modified the 2 failing tests, so that should fix things with that

BeckySharp:
ps -- hope you didn't mind us mucking in _your_ branch, seemed like a good way to make sure they survived and fixed the tests overall

kwalcock:
Not at all, especially if nobody else minds me having ransacked their branch when I merge this one.

PR #718, opened 2019-12-08
Make sure geonorm doesn't trip on empty raw text

kwalcock:


kwalcock:
FYI @EgoLaparra, @bethard.  There are pieces of raw text and have zero length because they aren't really there.  Something in the GeoNormFinder doesn't like them so they will be skipped.

PR #719, opened 2019-12-08
include trigger span in text,  add a cross sentence test

maxaalexeeva:


kwalcock:
Thanks so much for the test case.  If the trigger is not in the first sentence or last sentence, and it can't be in both, then its interval is being compared against intervals from other sentences.  I think that intervals are counted per sentence.  If an argument from sentence 0 starts at 10 and the trigger interval starts in sentence 1 at 5, then start will end up sentence 0 at 5 even though that doesn't seem right.  If the trigger is in sentence 3 starting at 20 and the last argument is in sentence 4 that is only 10 words long, the program will probably crash.  Does the trigger need to be combined with all the arguments in sentenceAndMentionsSeq instead?

kwalcock:
I'm also concerned about the tokenInterval coming in.  The constructor is only ever called from MigrationUtils.copyWithNewArgs with the new token interval calculated from

```
    def getNewTokenInterval(intervals: Iterable[Interval]): Interval = Interval(intervals.minBy(_.start).start, intervals.maxBy(_.end).end)
```

which doesn't take the trigger into account while I see elsewhere like in org.clulab.package.odin the trigger interval being taken into account.

```
  // token interval that contains trigger and all matched arguments
  def mkTokenInterval(trigger: TextBoundMention, arguments: Map[String, Seq[Mention]]): Interval = {
    val allStarts = trigger.start +: arguments.values.flatMap(_.map(_.start)).toSeq
    val allEnds = trigger.end +: arguments.values.flatMap(_.map(_.end)).toSeq
    Interval(allStarts.min, allEnds.max)
  }
```
These things seem quite complicated and I hope that we can eventually simplify them.

maxaalexeeva:
>  Does the trigger need to be combined with all the arguments in sentenceAndMentionsSeq instead?

You are right. Thank you. I missed that. 

>the trigger interval being taken into account.

That makes sense. Didn't know. 

I'll try to adjust the code and push again.

kwalcock:
That looks good.  How certain are we that
```
      odin.mkTokenInterval(mention.asInstanceOf[EventMention].trigger, newArgs)
```
is valid?  We have lots of unchecked assumptions and maybe it's a good one.  (I don't know.)  There is also some code for
```
  def mkTokenInterval(arguments: Map[String, Seq[Mention]]): Interval = {
```
I wonder if there needs to be something to handle either case, perhaps a match.

maxaalexeeva:
> I wonder if there needs to be something to handle either case, perhaps a match.

Not sure I understood this part. A match between what and what? 

Also, with either the odin token interval method or the token interval method I had before, the token interval in a crossSenteneceEventMention is still not great since the token int's are per sentence. Is there a way/does it make sense to not have token interval at all? That's sort of what this todo was about:
https://github.com/clulab/eidos/blob/b87a0697e33ac9b4f272c51d60be60da689fa927/src/main/scala/org/clulab/wm/eidos/mentions/CrossSentenceEventMention.scala#L9

kwalcock:
Thanks!  I think you're right about the tokenInterval.  Now I recall the CrossSentenceMention that doesn't have one and besides that I was just pointing out how the trigger comparison would be a problem across sentences.  What to do?  For a standard Mention, perhaps it should just max out the token interval for the sentence it is in, i.e., use only the args and trigger that are in the same sentence.  Becky probably knows better.

Not that it matters much then, but something like this might work for the match:
```scala
      mention match {
        case mention: EventMention => odin.mkTokenInterval(mention.trigger, newArgs)
        case _: Mention => odin.mkTokenInterval(newArgs)
      }
```

kwalcock:
I'll go ahead and merge it so that I have your changes.

PR #721, opened 2019-12-09
Use symmetric comparison, restrict tokenInterval to sentence

kwalcock:


kwalcock:
Here's a first pass.  Ordering is next.  What I forgot is deserialization.  That's something that I've been waiting on and haven't had time to do.  My fault.

kwalcock:
That's my suggested implementation.

kwalcock:
If it doesn't look unreasonable, it can go into the migration branch and then we can see what happens.

maxaalexeeva:
There are a couple of parts I don't completely understand (e.g., the part about hi/lo indices---I'll reread the code tomorrow), but I think it looks good?/ I don't see a reason not to merge it into the migration branch. 
One question: so the cross sentence event mention does not have tokenInterval anymore, but you still need the method to get the token interval (inside the crossSentEventMention object) because crossSentEventMention extends EventMention, and the latter does have to have a token interval?

kwalcock:
Thanks for looking.  Maybe I should name them left and right instead of lo and hi.  Anything but i and j.  The CrossSentenceEventMention does have a tokenInterval.  It just calculates it on its own instead of receiving something from MigrationUtils that is pre-calculated.
```
class CrossSentenceEventMention... extends EventMention(labels,  CrossSentenceEventMention.calcTokenInterval(sentence, trigger, arguments, document.sentences(sentence).startOffsets.length),...
```

maxaalexeeva:
Oh I see. The token interval is there but it comes from 'extends' . Thanks!

PR #722, opened 2019-12-10
Transparent

BeckySharp:
@bgyori -- this addresses your desire for consistency (gosh, so demanding...)
Essentially, when we "reground" we will pass text through the canonicalization process.

However -- this also moves some of the "transparent" words to the stop.txt file so that we can use the transparent list to (a) not count towards "contentfulness" but (b) be included when doing grounding (i.e., `children` isn't enough to make it a mention, but it matters when choosing a node...)

@kwalcock - i just used the isStopwordStrict you long ago added for such a time as this -- did I break anything? I guess jenkins will let me know...

fyi @MihaiSurdeanu


kwalcock:
Is this so that whatever is regrounding doesn't need to pass in both the text and the canonical name (and the mention type to support `EidosOntologyGrounder.groundableType(mention)`)?

BeckySharp:
it's so that the canonical name doesn't need to be passed, yes.  Ben indicated that he sometimes passes canonicalized text and sometimes not, so here if it's re-canonicalized canonical text nothing is lost, and if it's not already canonicalized, then it will be

BeckySharp:
but, @kwalcock one thing is that i changed the type of the processer from a SentenceSplitter to a EidosProcessor, and idk if it's a problem

kwalcock:
It looks like the text of a mention is taken from the raw text and not the words.  I think that implicitly inserted periods will not be included and they won't be reinserted if the SentenceSplitter is not used.  I'm not too worried about that.  I would be more concerned that the processor is trying to treat canonical text like the real sentence text and will get different lemmas, ners, and tags.  I wouldn't want to guarantee that the new grounding on the same text will match the old.

BeckySharp:
@kwalcock fair point.

@MihaiSurdeanu  would you like to weight in on how you feel about this change?
And @bgyori -- note the above point

kwalcock:
...not to say that it wouldn't be useful.  I just note the limitations.  I'm tempted to PR your PR.

MihaiSurdeanu:
Yeah, this is a bit tricky...
On one hand, we want to be close to the original text (raw), so downstream apps can match our extractions back to text. But the raw text is sometimes not useful, e.g., it may include weird Unicode characters such as Greek characters, which are not grounded well. This was a problem in Reach. However, I don't think this is a major issue here, is it?

BeckySharp:
well, the larger issue is that the lemmatization/NER/pos tagging of a
mention in isolation may be different than what it originally was when
parsed as a whole sentence.
I think we're fine for this eval (just a hunch), but it's worth noting....
an alternative would be to add a boolean to the API to allow Ben to
recanonicalize when he wants to...?

bgyori:
Yes, an additional argument to control whether or not canonicalization should be done would work well. I can control that from the outside as appropriate.

kwalcock:
I like the additional argument.  The computer won't need to unnecessarily canonicalize and it won't unnecessarily make a mistake.

BeckySharp:
Keith -- any chance you can add that while I'm at the ASIST kickoff?? I
can't work much this afternoon

kwalcock:
Yes

BeckySharp:
THANKS!!

BeckySharp:
@kwalcock what's the status of this bad boy?

kwalcock:
Giving Ben adequate time to comment.  Just achieved.

BeckySharp:
thanks! :)

kwalcock:
```
  override def containsStopwordStrict(stopword: String): Boolean = stopwords.contains(stopword)
```
seems to lead to many calls to stopwords.  Stopwords is a def which reads the entire stopwords file, many, many times.  I'm going to try to change it to a lazy val instead.

PR #726, opened 2019-12-12
Work on serialization

kwalcock:
Add lots of tests, deserialize the CrossSentenceEventMention, output groupModifier


kwalcock:
The serialization has just a little ways to go.  It has small discrepencies in ordering on a round trip test.  That's why it isn't quite ready for master.  It is, however, being used for the DART reading.  There we're not deserializing yet.  It's getting really close.

PR #727, opened 2019-12-12
Only read stopwords file once

kwalcock:
It was a def before so that the results could be thrown away.  Now they are used repeatedly.


BeckySharp:
thanks!

PR #728, opened 2019-12-13
Test CrossSentenceEventMention

kwalcock:
Serialize and deserialize the migration events, including CrossSentenceEventMention, and add tests for remaining attachments


BeckySharp:
thanks keith!

PR #730, opened 2019-12-24
Full ontology tree experiment parent name filtered no root

BeckySharp:
Let's get this merged in so we can re-ground the reader output before the experiment?

fyi @MihaiSurdeanu  @Fan-Luo

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -40,11 +40,11 @@ ontologies {
   wordToVecPath = /org/clulab/wm/eidos/${EidosSystem.language}/w2v/glove.840B.300d.txt
 
   // Activated Ontologies
-  ontologies = ["wm", "un", "props", "mitre12", "who", "interventions"] // , "icasa", "mesh"]
+  ontologies = ["wm"] // , "un", "props", "mitre12", "who", "interventions"] // , "icasa", "mesh"]
 
   // Caching, for quick loading, language dependent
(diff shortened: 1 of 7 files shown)


kwalcock:
This branch is not meant for checking into master and I should delete the PR.  It is the base branch for the grounding experiments.  useW2V = true cannot be used in the master branch because we can't test or run with w2v.  Full caching cannot be used with the grounding experiments because the ontology caching was highly optimized and doesn't support the operations needed for the experiments.  If master needs to be cacheable, then the compact ontology representation needs to be reworked.  I believe the code here was hacked so that the vector file could still be cached while the ontologies wouldn't be.  It might not be possible to use a configuration setting for parents on/off and still cache in as an efficient manner as previously.  The code used for the recent reading is in DART-42k* and includes some of this code.  It should be able to reground.

BeckySharp:
OK -- so then we'll need to reground?  I looked on the DART4-42k* branches and it seems that DART4-42k-B and DART4-42k-B1532 both have the parentName grounding implemented, 
(a) was the latest output we sent to Ben run with one of these branches and if so, was this non-leaf node grounding activated? thanks!

BeckySharp:
I think so -- since we got "conflict" as one of the groundings -- thanks!

kwalcock:
It was at least intended that grounding was activated and to the flattened version at that, and not to the props at all.  I did some checks that indicated that the settings were effective.

BeckySharp:
Awesome, thanks!

PR #734, opened 2020-01-11
Run the tests

### LINKED ISSUE #708: Some tests aren't testing
Some tests are missing their `should be(successful)` so that they really aren't checking the result of the test.  Lots of them are failing and we don't know it.

BeckySharp:
This is the PR addressing Issue #708 
fyi @MihaiSurdeanu 

Most of the tests were original, meaning that they had been sneakily not running ever since we first started developing rules, so many were about aspects of reading we didn't end up handling (i.e., different dimensions of entity modification, etc).

I modified some tests, others I tweaked rules to capture the info. Some I removed bc they weren't within the scope of what Eidos does

I guess we're not allowed to merge until after the freeze is lifted?  Or can we?  There are new rules, so I'd imagine we can't.

Also, idk why it can't automatically merge, i'll check in a sec, once I open it.

Closes #708

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -44,7 +44,7 @@ ontologies {
 
   // Caching, for quick loading, language dependent
   cacheDir = ${EidosSystem.cacheDir}/${EidosSystem.language}
-  useCache = false
+  useCache = true
 
   // Primary
   // Note that this first one is included via build.sbt.  It is not directly part of eidos.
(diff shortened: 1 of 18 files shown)


MihaiSurdeanu:
Thanks @BeckySharp !

kwalcock:
No thanks to me, @BeckySharp.  Can you look at this conflict?  These monster strings are not good for merging.  Is there any other way they can be coded?  Lists would work better.

BeckySharp:
> No thanks to me, @BeckySharp. Can you look at this conflict? These monster strings are not good for merging. Is there any other way they can be coded? Lists would work better.

I think it's fixed! thanks Keith!

Issue #735, opened 2020-01-18
webapp brat syntax display broken 


BeckySharp:
The syntax brat on the webapp is no longer displaying (on master).
Did we recently update something?  I looked through the "blame" on github and can't find any recent changes to HomeController or index.scala.html that would have done it... :(

@kwalcock can you please help!? :(

kwalcock:
Looking now

kwalcock:
Can you give an example?  My one test gives this output.  Is something missing?

![brattop](https://user-images.githubusercontent.com/8679738/72672377-2b65ba00-3a16-11ea-98cd-794ca6031863.png)

![bratbot](https://user-images.githubusercontent.com/8679738/72672380-37517c00-3a16-11ea-897d-379a6c95d93b.png)

BeckySharp:
it's a Martin Luther King Day miracle!
I was getting an empty blue outline box thing instead of the display... i tried switching branches etc and it was there.  But now, it's working again? though I changed nothing? 
I'd think it was a browser caching thing, but I even tried in chrome....
ok -- well, then either a tesseract or a miracle or just some weird user error... 
THANKS!

PR #736, opened 2020-01-19
Alleviate as a trigger (plus some of his friends... and enemies?)

BeckySharp:
Added triggers for inc/decrease, including the ones that @MihaiSurdeanu sent by email.
_In fact_, those were turned into tests! I know, right?!?
This is in advance of the planned rearrangement of semantically non-neutral triggers... hopefully there is more fun to come!

Also -- affirmed by the miraculous healing of the webapp (#735), I decided the stars were aligned, so I boldly moved the syntax up in the webapp display (to just below the extraction brat), something masha suggested in a diff project maybe and I liked...

ps -- i've been wanting to add more triggers for so long, this PR feels like scratching an itch! Is there a github PR label for that? :)

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -44,7 +44,7 @@ ontologies {
 
   // Caching, for quick loading, language dependent
   cacheDir = ${EidosSystem.cacheDir}/${EidosSystem.language}
-  useCache = false
+  useCache = true
 
   // Primary
   // Note that this first one is included via build.sbt.  It is not directly part of eidos.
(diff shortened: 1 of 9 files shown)


kwalcock:
I'll see if that merge at the end went OK as soon as I try to merge in the compositional grounding which is in the middle of its own merge.

kwalcock:
Drat.  I was on the wrong screen when I pressed that button.  Doing this backwards...  I'm going to remove that return statement when I get a chance.

BeckySharp:
thanks!

PR #737, opened 2020-01-20
Merge compositional grounding

kwalcock:


kwalcock:
Because of the large number of merge conflicts with main, this branch was forked from zupon-compositional-grounding so that it could be worked on in stages rather than a single session.

kwalcock:
This in OntologyGrounder.scala is a no-no:
```
|      [chunk=/NP$$/ & !word=$synHeadWord & !tag=/DT|JJ|CC/]+
```
One cannot just paste some text into a yaml string.  This breaks very quickly with real documents.

BeckySharp:
> Because of the large number of merge conflicts with main, this branch was forked from zupon-compositional-grounding so that it could be worked on in stages rather than a single session.

good plan! 👍 and we feel for you!!

kwalcock:
I have the yaml escaped.  The regular expression will soon follow.

kwalcock:
@zupon, @BeckySharp,

If the line
```scala
val headText = mentionHead.map(_.text).getOrElse("<NO_HEAD>")
```
gets executed, what is the expected outcome of
```scala
val modifierMentions = getModifierMentions(headText, mention.odinMention, allowedMods.mkString("|"))
```
I would like to short circuit the calculation if possible.  It would make the program faster and spare us from yet another placeholder to avoid in the real text.

kwalcock:
@zupon, @BeckySharp ,

Can I assume that the

```
         |      modifier: Chunk+ = >/^$pattern/{0,2} >/amod|compound/?
```
is meant to be
```
         |      modifier: Chunk+ = >/^($pattern)/{0,2} >/amod|compound/?
```
That is, ^ must apply to whichever of `compound|nmod_of|nmod_to|nmod_for|nmod_such_as` in the pattern it turns out to be and not just the first one.

BeckySharp:
> @zupon, @BeckySharp ,
> 
> Can I assume that the
> 
> ```
>          |      modifier: Chunk+ = >/^$pattern/{0,2} >/amod|compound/?
> ```
> 
> is meant to be
> 
> ```
>          |      modifier: Chunk+ = >/^($pattern)/{0,2} >/amod|compound/?
> ```
> 
> That is, ^ must apply to whichever of `compound|nmod_of|nmod_to|nmod_for|nmod_such_as` in the pattern it turns out to be and not just the first one.

I think yes, good catch!

BeckySharp:
> @zupon, @BeckySharp,
> 
> If the line
> 
> ```scala
> val headText = mentionHead.map(_.text).getOrElse("<NO_HEAD>")
> ```
> 
> gets executed, what is the expected outcome of
> 
> ```scala
> val modifierMentions = getModifierMentions(headText, mention.odinMention, allowedMods.mkString("|"))
> ```
> 
> I would like to short circuit the calculation if possible. It would make the program faster and spare us from yet another placeholder to avoid in the real text.

@zupon please correct me if I'm wrong, but I think that the goal here is to end up with the _head_ and the _modifiers_ of that head to exist as indiviual *Mentions* to pass to the grounding algorithm.  That said, it may well be the case that we can just ground the strings, but in any case we're trying to "ground" the head (e.g., _access_) separately from the modifers (e.g., _fresh water_)

kwalcock:
Oops, a little.  I mean if it is executed and the result is `<NO_HEAD>`.

BeckySharp:
fair point, though there would be no modifiers found, so we'd just be grounding "<NO_HEAD>" which wouldn't be in the vocabulary of the embeddings.  Though we can try to catch it... @zupon ?

kwalcock:
So I can do something like
```scala
val modifierMentions =
    if (headText.isDefined) getModifierMentions(headText, mention.odinMention, ...)
    else Seq.Empty
```

BeckySharp:
yes

kwalcock:
This is still going to undergo a better stress test.

What should the default ontologies now be in master?
What should that PRIMARY_NAMESPACE be?

kwalcock:
An EidosMention now has two vars.  I should really fill them in when they are deserialized from jsonld instead of having them recalculated under potentially different circumstances.  The change from lazy val makes this possible and almost necessitates it.  I guess it's not urgent, though.

kwalcock:
There is a case where headTextOpt = "C\nREPUBLIC\nO F T H E CO N G O\n\n".  Is whitespace expected there?

kwalcock:
This is about to go into master, so any reviewers need to object soon.

zupon:
Sorry for my radio silence on these issues. I've been wiped out by a cold all week so I haven't been checking in as often as I should. I agree with @BeckySharp on her comments above. I don't know the answer to the questions about PRIMARY_NAMESPACE or the default ontologies.

BeckySharp:
I'll try to review this tomorrow morning, sorry

BeckySharp:
> What should the default ontologies now be in master?
> What should that PRIMARY_NAMESPACE be?

It should be the `wm_flattened` one

BeckySharp:
> An EidosMention now has two vars. I should really fill them in when they are deserialized from jsonld instead of having them recalculated under potentially different circumstances. The change from lazy val makes this possible and almost necessitates it. I guess it's not urgent, though

This makes sense, esp as the code is changing very rapidly and we anticipate more changes

BeckySharp:
> "O\n\n"

Not sure -- the text comes from the Mention.text, the mention span comes from just the found head, and the head is found with the code in Processors, so idk.  I guess if we're trying to ground things, we should maybe strip it?  Do we do some text normalization in the grounding process? Should stripping (if we decide to use it) live there?

BeckySharp:
> > What should the default ontologies now be in master?
> > What should that PRIMARY_NAMESPACE be?
> 
> It should be the `wm_flattened` one

for now -- we may want to have the compositional one in the future, but IDK what the ramifications of that would be

PR #740, opened 2020-01-28
adding positive and negative affects...

BeckySharp:
added to the concept states as well as relations between concepts.
Polarity will need to be resolved downstream to add relations to a
CAG.


BeckySharp:
@kwalcock not sure if I handled the merge conflict correctly in JLDSerializer, FYI

BeckySharp:
Also -- doesn't yet have tests, maybe you can help me write some?

kwalcock:
Will look soon.  I'm still deserializing the groundings just now.

kwalcock:
Do you have a good example sentence that I can use for testing?  When you write positiveaffect, this is the rare use of affact as a noun and it doesn't mean positive effect, right?

kwalcock:
Probably quite a few POS and NEG were found, which is why the tests are failing.  You're probably working on them already.

BeckySharp:
actually I wasn't but I will!! :)
Would you mind updating the testing framework to include checking for Pos()
and Neg() modifiers as well as PosAffect and NegAffect relations?

BeckySharp:
@bgyori if/when this gets merged, concepts will also have positive and negative change attachments (along with the increase/decrease), but to make use of them, you'll have to resolve it with the ontology polarity.
Speaking of that, I'll go through the ontology really soon to make sure the polarities are set.  Are we going with -1/1 @bgyori ?

bgyori:
Sounds good, yes, -1 and 1, and if it's missing, we just assume it's 1 by default.

kwalcock:
@BeckySharp, thanks so much for working on this.  I'm going to merge it and put this on my todo list if necessary.  Can you tell me a little about HomeController$.statefulRepresentation?  It's on an object and public as if it is ready to be called from elsewhere besides a HomeController class, but it isn't.  If it was to be called from elsewhere, I'd want to put most of the logic in the EidosMentions to avoid all the class checking.  If that's not the intention, it could be a protected class method.  So, I'm just wondering where it's going.  Thanks.

BeckySharp:
It can be protected for sure (and I just pushed it), the method is only
there to add a suffix to the label for display purposes in the webapp.
It's supposed to trigger a color change and everything (though last I
checked that wasn't working the way I expected it to... but it could simply
have been from needing a reload and isn't urgent)

kwalcock:
Thanks for helping me to resist the urge to rewrite it!

kwalcock:
That was some extensive merging.  Thanks for handling it.

PR #741, opened 2020-01-28
Deserialize groundings

kwalcock:
Groundings are now stored in an EidosMention var, so they can and should be set upon deserialization.  This was previously impossible because they were a lazy val.


kwalcock:
FYI, @BeckySharp.  The deserializer does not have to have Eidos fill in the canonical name or groundings.  They are read from the file.  Eidos can re-postprocess or just reground if it wants to.

PR #742, opened 2020-01-28
Make Pos and Neg testable

kwalcock:


BeckySharp:
Keith -- you are so great!

PR #744, opened 2020-01-30
Test the grounding

kwalcock:


kwalcock:
I'm about to conclude that using data from this source, grounding_sample_v2 MS.xlsx, is not going to work well enough.  Only the original text is included, not the canonical text, and we cannot get the canonical perfectly from the text without rereading.  When that is done, the text of the mention doesn't necessarily match what is there so that it is difficult to figure out which mention to use.  We only have 13 subject groundings annotated and 14 object ones.  We're skipping 2 subjects and 6 objects because of the mismatches.  That's very little data to go on and not much of a head start over the previous way we were doing it.  I had hoped that some of the findings from other readers would match more readily.  One from hume did, but none of the others.  I could relax the match and if only a single mention is found, assume that it is the one in question.  For the other way we were doing it, the original data included the canonical text.  If we don't change the way that is calculated, then we can ground to that directly and separate the evaluation of grounding from the finding of the events.  If we're going to have someone spend time checking on the groundings, I would start from data for which we have recorded the canonical texts.

BeckySharp:
however, part of what i’d like to be able to evaluate is different ways of
making the canonical text... we can serialize while mentions, the Document
is included

kwalcock:
I updated the spreadsheet, mostly, so that the mentions could be matched better.  It turns out that the spreadsheet has tokenized text for the sentence, but raw text for the cause and effect, which was a problem.  The updated Excel spreadsheet is in the directory below along with the text version that the little app runs on.  If someone is going to check the groundings, then I'd definitely concentrate on the ones marked for the reader eidos.  The others seem unlikely to be found.  There are some 274 groundings for eidos, and then times two for cause and effect.  Mihai marked about 40 for eidos.  Who is next?

https://drive.google.com/open?id=153R-AQFsxZl8qo-aUTAY8u8eYMfIFDoP

kwalcock:
This PR just adds a little app, independent of eidos proper, so I'll go ahead and check it in.

MihaiSurdeanu:
Thanks @kwalcock !
Let's use *just* the rows I annotated for now.

PR #745, opened 2020-01-30
Zupon compositional grounding

zupon:
Changed `AllWords` rule from marking multiple words as one chunk. Instead, the rule will make each word a separate chunk. This will stop things like `water transportation` from being treated as a single modifier, letting us ground `water` and `transportation` separately.


zupon:
@BeckySharp I fixed the merge conflict. I think it was something that changed after I made the original PR, but now it matches.

kwalcock:
@zupon, if you would like to delete the branch on the server, go ahead.  I'll refrain in case it deserves to live on.

PR #746, opened 2020-01-31
Access glove via library dependency

kwalcock:
No separate download is required.  Download is now automatic, but perhaps superfluous.


kwalcock:
This is not working, because travis cannot speak to artifactory when they are both running on river.  It may have something with the reverse proxy creating some kind of loop.  Eric and I are working on it.

kwalcock:
The testing is working, although I still need to automate the configuration.  With a few lines of code changes, we can skip messing around with the glove file and do unit testing of grounding--at the expense of people downloading the vectors file who may never actually use the grounding.

kwalcock:
The servers seem to be working with automatic configuration, so we'll soon need to decide if this strategy is worth a try.

BeckySharp:
So does this mean that running something (i.e., building the jar) will always have the slow step of loading up the embeddings?  Do we benefit from the cached version you made?

kwalcock:
We can still cache as before.  If something like sbt is building the eidos jar in order to run it, it does not have to take the vectors from the resource directory and stuff it into the jar file.  It should have access to the version in the ivy2 directory.

BeckySharp:
nice!

kwalcock:
Documentation has been added to the CLU lab's wiki about Artifactory: https://github.com/clulab/clulab/wiki/Artifactory.  In addition, we have more space now.  I'm thinking about merging this finally.  Does anyone listening have reservations/concerns?

MihaiSurdeanu:
This is awesome!
Thanks Keith!!

Can you please email clulab this Wiki link? I think this is useful for
several projects.


On February 26, 2020 at 12:55:39 PM, Keith Alcock (notifications@github.com)
wrote:

Documentation has been added to the CLU lab's wiki about Artifactory:
https://github.com/clulab/clulab/wiki/Artifactory. In addition, we have
more space now. I'm thinking about merging this finally. Does anyone
listening have reservations/concerns?

—
You are receiving this because you are subscribed to this thread.
<https://github.com/clulab/eidos/pull/746?email_source=notifications&email_token=AAI75TXHQ7IZDDCX3ETBILLRE3CLXA5CNFSM4KOKDLOKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOENBU2FQ#issuecomment-591613206>,
or unsubscribe

kwalcock:
@bgyori, it turns out that this affects you.  You've been using vectors.txt rather than glove.  Glove is being declared a library dependency and vectors.txt not (at least not yet).  This means that the sed command won't work because the two files are not interchangeable.  However, if you are OK changing to glove, the command is not necessary.  In fact, these could all be removed:
```
RUN wget https://s3.amazonaws.com/world-modelers/data/vectors.txt
RUN mv vectors.txt src/main/resources/org/clulab/wm/eidos/english/w2v/
RUN sed -i 's/glove.840B.300d.txt/vectors.txt/' src/main/resources/eidos.conf
```

Glove will be included via `RUN sbt assembly`.  Are you OK with this?  I'll add the docker changes to the PR so you can see or try.  Thanks.

BeckySharp:
> @bgyori, it turns out that this affects you. You've been using vectors.txt rather than glove. Glove is being declared a library dependency and vectors.txt not (at least not yet). This means that the sed command won't work because the two files are not interchangeable. However, if you are OK changing to glove, the command is not necessary. In fact, these could all be removed:
> 
> ```
> RUN wget https://s3.amazonaws.com/world-modelers/data/vectors.txt
> RUN mv vectors.txt src/main/resources/org/clulab/wm/eidos/english/w2v/
> RUN sed -i 's/glove.840B.300d.txt/vectors.txt/' src/main/resources/eidos.conf
> ```
> 
> Glove will be included via `RUN sbt assembly`. Are you OK with this? I'll add the docker changes to the PR so you can see or try. Thanks.

@bgyori :
pro: GLOVE is a better set of embeddings
con: GLOVE take up (a lot) more memory

kwalcock:
It doesn't make a difference yet, but it looks like we're going to start filtering Glove and keep only the words we're likely to encounter.  That's unlikely to happen at runtime, though, so there will probably be a different file to use.  It's hard (for me) to imagine we'd keep up a parallel development for vectors.txt.

bgyori:
This looks like a great change! If I understand correctly, `sbt assembly` will produce a JAR which still includes the file, it's just that we don't have to manually download it and place it at the configured path anymore for it to be found.

MihaiSurdeanu:
Yep. @kwalcock is awesome.

bgyori:
One more question came to mind: wouldn't it make sense now to set `useGrounding` to `True` by default? If the large w2v file will be there anyway, we might as well use it by default (and most use cases will have to turn this on anyway). Or is there some consideration why keeping it `False` by default is better?

MihaiSurdeanu:
What do you think @BeckySharp ?

kwalcock:
Program start times and memory usage increase significantly when grounding is turned on, especially if the vectors file has not been cached, which is still a manual task.  It seems like quite a few people would need to turn it off right away in order to do development work.  Maybe it's more important that production use can run directly on master.  Relatively soon the pared down file might simply be included in the resource directory without causing problems with the default set to true. Either way is OK, though, especially if someone else decides.

BeckySharp:
> Program start times and memory usage increase significantly when grounding is turned on, especially if the vectors file has not been cached, which is still a manual task. It seems like quite a few people would need to turn it off right away in order to do development work. Maybe it's more important that production use can run directly on master. Relatively soon the pared down file might simply be included in the resource directory without causing problems with the default set to true. Either way is OK, though, especially if someone else decides.

I'm fine with either -- it's more of a question of who we should inconvenience (when they inevitably forget to toggle it), people who are using Eidos to read and ground or people who are developing/testing.  Based on that, I guess my vote would be default on, but again, I'm fairly indifferent...

kwalcock:
Shall I go for the whole hog/enchilada/nine yards then?  FWIW, TwoSix recently built and found out afterwards that grounding wasn't included.

MihaiSurdeanu:
I think it's easier to train our lab to remove grounding than to train everybody else to enable it...
I propose to have the default behavior use grounding.

kwalcock:
@bgyori, if when you build the Docker file you want to do
```
RUN sbt "runMain org.clulab.wm.eidos.apps.CacheOntologies"
```
then you could turn on the useCached* values and get some shorter startup times.  That was possible before, but it may be more important now with the larger glove file.

kwalcock:
Here goes.

PR #749, opened 2020-02-06
Support includeParents in cached version of ontology

kwalcock:


kwalcock:
@BeckySharp, this is what you have been waiting for.  Eidos can now cache the trees when includeParents is true.  The serialized versions are *.fast.serialized.  The most recent algorithm, parent name filtered no root is being used.

BeckySharp:
AWESOME, thanks!!

PR #750, opened 2020-02-11
Maas

BeckySharp:
This has the APIs for Jataware to use the MaaSUtils grounding -- i.e. finding the best bottom up variables/inputs/ouputs/models for a given node

@brandomr  is currently using it in this (maas) branch, I think
If this is merged in, then master can be used (yay!) -- 
BUT @brandomr **NOTE: for this to work**, the webapp must be launched with the 
`useW2V = true` in `eidos.conf`

FYI @MihaiSurdeanu @bgyori


kwalcock:
Can you confirm that the configured values for ontologies.ontologies are what we want in master?  It changes so often.

kwalcock:
At some point I should do something about the now three different json libraries we're using in Eidos: play.json, json4s, and now ujson.

kwalcock:
In the next PR I am going to do something about the redundancies in the config file.

BeckySharp:
> Can you confirm that the configured values for ontologies.ontologies are what we want in master? It changes so often.

Yes -- the wm_flattened is (for now, will likely change soon) the default flat ontology, the wm_compositional is the (only) compositional one, and the un etc shouldn't still be used

BeckySharp:
> At some point I should do something about the now three different json libraries we're using in Eidos: play.json, json4s, and now ujson.

yes -- that sounds like a thing!

brandomr:
Awesome, thanks for merging this in!

PR #752, opened 2020-02-13
Clean up the configuration

kwalcock:
Among other changes, eidosTest.conf "inherits" from eidos.conf via an include and then indirectly from reference.conf after that. This way it doesn't skip directly to reference.conf and need to repeat lots of things in eidos.conf. It can just override properties involved in testing.


kwalcock:
You are too fast.  See two comments in testEnglish.conf.  I probably just missed these things in a merge conflict.

kwalcock:
Right on the above.  This line
```
expander = { include required("englishTestActionsExpander.conf") } // kwa: This is different from reference.conf!
```
in englishTest.conf and the associated file should not be necessary because they make the test environment different from the production environment.

Also, in EidosSystem of englishTest.conf (and also of portugueseTest.conf)
```
  conceptExpander = ${actions.expander} // kwa: This is different from reference.conf!
```
the conceptExpander is linked to actions.expander while in reference.conf
```
  conceptExpander = { include "englishConceptExpander.conf" }
```
the values are quite different (englishConceptExpander != englishActionsExpander).

So far I don't see a way to parameterize a configuration like EidosSystem(language).useLexicons.  I don't like the duplication for Portuguese.  Maybe next time.

BeckySharp:
Thanks!
OK -- so then as my action item: I should check this branch out, make the
confs match and fix tests that break?

kwalcock:
The quick fix is to make the main version the same as the test version, and I can handle that, but perhaps you have reasons that that wouldn't be a good idea.  So, in some cases where the main version becomes the test version, tests might break and would need to be fixed.  You're the guru there, but let me know if I can help.

BeckySharp:
Let me look more closely at the diffs... things *were* done intentionally
-- but I don't remember what those intentions were anymore.... :)

BeckySharp:
OK -- so now we are using the same expansion in the test portion as in the main portion of the code -- but there are _slight_ diffs between argument expansion and concept expansion... i basically didn't want the concepts to expand out to the whole sentence -- in the arg expansion the trigger kinda stops certain rampant deps ;)
If the tests pass here (they darn well should! -- famous last words) then we'll merge?
In general I LOVE the cleaning of the configs, @kwalcock !!!!!!

kwalcock:
This looks like a real error:
```
[info] TestEntityFinder:
[info] rainfall entity
[info] - should have leading preposition trimmed
[info] JJ entity
[info] - should only be kept if following word was avoided as a trigger *** FAILED ***
[info]   false was not true (TestEntityFinder.scala:25)
[info] TestQuicklyEqualable:
```

BeckySharp:
ok, i'll fix that

BeckySharp:
> This looks like a real error:
> 
> ```
> [info] TestEntityFinder:
> [info] rainfall entity
> [info] - should have leading preposition trimmed
> [info] JJ entity
> [info] - should only be kept if following word was avoided as a trigger *** FAILED ***
> [info]   false was not true (TestEntityFinder.scala:25)
> [info] TestQuicklyEqualable:
> ```

FIXED sir! :D

kwalcock:
I did a diff to spy on you and then decided to make the differences a tiny bit more obvious.  Only formatting changes.

PR #758, opened 2020-02-18
Add keepMigrationEvents to configuration.

kwalcock:
Try again with more space on Jenkins.


kwalcock:
This finally made it through Jenkins.  It's the first pass.  There will probably be a separate migration section in the config file soon, but for now I just added it to EidosSystem.

kwalcock:
That idea about preprocessors and postprocessors is looking good.  I snuck in an interface for the preprocessors and they might show up soon.

PR #759, opened 2020-02-19
added a couple of rules and create_triggers

BeckySharp:
Made some better extractions...


BeckySharp:
fair enough -- forthcoming.
FYI @MihaiSurdeanu these came out of the intervention doc that was sent around

kwalcock:
It looks like GitHub is having some kind of problem.  It won't merge.  I'll check back periodically.

PR #760, opened 2020-02-19
Reground sentences

kwalcock:


kwalcock:
@zupon, this isn't ready to go yet, but the intended functionality has been placed into OntologyHandler.reground(sentenceText: String, interval: Interval).  If that looks like the wrong place or the wrong shape, you can let me know.  The idea is to go through all the steps that would normally be taken with OntologyHandler.process(eidosMention) but not have each individual grounder need to deal with anything other than an EidosMention or have each separately convert text to a Mention.

kwalcock:
@zupon, after this is merged, you can change a couple of lines in ExtractAndExport.scala to something like

```scala
  val groundedAs = groundAs.flatMap { grounder =>
    grounder match {
      case "wm_compositional" =>
        CompositionalGrounder.branches.map(OntologyHandler.mkBranchName(grounder, Some(_))
      case other => Seq(other)
    }
  }
```

kwalcock:
There are problems if, for example, the sentence text coming in gets parsed into two different sentences.  I'm punting on that issue here.

zupon:
Thanks @kwalcock ! I think this should work for my purpose. @BeckySharp , does this look alright to you?

src/main/scala/org/clulab/wm/eidos/apps/GroundCanonicalNames.scala
```diff
@@ -40,6 +42,24 @@ object GroundCanonicalNames extends App {
         (name, value, nameToIsLeaf(name))
       }
     }
+
+    def ground(text: String, canonicalName: String): Option[(String, Float, Boolean)] = {
+      val canonicalNameParts = canonicalName.split(' ')
+      val start = text.toLowerCase.indexOf(canonicalNameParts.head.toLowerCase)
+      val stop = text.toLowerCase.indexOf(canonicalNameParts.last.toLowerCase, start)
```
BeckySharp: These seem problematic -- what if we had the last word appearing N times in a sentence, isn't `indexOf` going to return the first occurrence?

kwalcock:
I should have written a comment.  That method is never called.  It was just used temporarily to test the call to the regrounder.  It usually worked.  For the regrounding call, something else needs to decide on the real interval.

BeckySharp:
got it! thanks!

kwalcock:
Some documentation has been added so that I remember what it was there for.

PR #762, opened 2020-02-20
Try again on migration

kwalcock:


kwalcock:
Mihai wanted a way to configure turning migration on and off.  There was a previous merge in which I implemented that... incorrectly.  This time I'm going to wait for @zupon or @maxaalexeeva to OK the changes first.

kwalcock:
It seems that the way to go about it is to let Odin go ahead and find the migration events, but then to filter them out, on demand, in EidosActions.globalAction.  Does that sound (and look) right?

MihaiSurdeanu:
This is not fully efficient, but it is not bad, *if* these migration events are not blocking us in picking up other information related to causality. @BeckySharp: what do you think?

kwalcock:
I agree on both counts: inefficient and perhaps disadvantageous.  It seemed that trying to take them out later would be very difficult and more inefficient.  The next option is in the Migration handler, I think.

BeckySharp:
Ill look into it tomorrow morning

BeckySharp:
So -- I don't think that finding migration events will impede finding causal events -- the human migration events aren't entities, so they _shouldn't_ affect the entity boundaries involved in expansion etc.
That said -- if we want to disable them... we can maybe split out the grammars -- currently we're (1) finding entities with entitiy finders
(2) running the "master" grammar, which includes (in order):
 - finding context entities
 - finding human migration events, entity quantification, and temporal mentions
 - finding increase/decrease of entities
 - finding linking events (i.e., causal/correlation/coref)

Maybe we can make the context and migration steps be "finders" too (or something like that, i.e., they can have grammars and be odin based, but be independent extractor engines that run first) -- then transition into the rest?

MihaiSurdeanu:
I think making the steps in (2) be optional "finders" makes perfect sense. But how much work would that take?

BeckySharp:
@MihaiSurdeanu prob not too much, we already have a bunch of finders that we can use as "templates" and there's a generic loader that will load as many as you ask for, in the order that you ask for them
i mean.. famous last words and all...

MihaiSurdeanu:
Ok. Should we discuss it at the next meeting? Maybe Andrew has some cycles
next week?


On February 21, 2020 at 10:58:10 AM, Becky Sharp (notifications@github.com)
wrote:

@MihaiSurdeanu <https://github.com/MihaiSurdeanu> prob not too much, we
already have a bunch of finders that we can use as "templates" and there's
a generic loader that will load as many as you ask for, in the order that
you ask for them
i mean.. famous last words and all...

—
You are receiving this because you were mentioned.
<https://github.com/clulab/eidos/pull/762?email_source=notifications&email_token=AAI75TRGCVSNOS37MIDPUADREAI3FA5CNFSM4KYYW3QKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEMTRNVQ#issuecomment-589764310>,
or unsubscribe

BeckySharp:
sure

BeckySharp:
I think the finders option is a cleaner workflow...

kwalcock:
We discussed this very briefly this morning.  It sounds like in the interest of expediency, we may want to add this version at least temporarily to master.  I wasn't able to tell Mihai where exactly I had put the code.  It is in the EidosAcctions.globalAction.  @BeckySharp, do you know if that should at least be effective in preventing migrations from appearing outside these confines?

MihaiSurdeanu:
As I mentioned, I am Ok with this as a temporary solution.

BeckySharp:
AFAIK this will work here -- we also can modify stuff in these places bc we're already filtering mentions to make sure they're "cag relevant"
 here: https://github.com/clulab/eidos/blob/master/src/main/scala/org/clulab/wm/eidos/utils/StopwordManager.scala#L63-L89
and here: https://github.com/clulab/eidos/blob/master/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala#L205
used here: https://github.com/clulab/eidos/blob/master/src/main/scala/org/clulab/wm/eidos/EidosSystem.scala#L126

Maybe something that defines what is cag relevant through the config?

kwalcock:
No, Becky, I can't simply press a button once or twice.  This was not the PR I intended to merge this morning.  I hope it's OK, at least until an alternative solution is found.

BeckySharp:
hahahahaha my bad! sorry keith!!!

PR #763, opened 2020-02-20
Fiddle with EidosSystem

kwalcock:


kwalcock:
To run those intervention documents, I again had to fiddle with the metadata handling, so I decided to do a little something about it.  It was already an issue because whether documents had an ID or had a DCT and whether it was a string or not, etc., was cluttering up the interface.  It's only going to get worse and lots of the programs in the batch directory are already adding information behind Eidos' back in lots of duplicate code.  The main entry points into EidosSystem are changed here, keeping backwards compatibility while calling it legacy.  They now take a general Metadata object which they can pretty much ignore, and another general Options object which is currently only used for that cagRelevantOnly, but then can be expanded without impacting the method signatures.

kwalcock:
The more interesting and maybe controversial change is the use of poorly named PreProcessors to set up a pipeline.  Some methods are partially exposed (still marked protected) which hint at how the pipeline could be adjusted either in the configuration file or even at runtime.  It could be made nicer if the classes involved implemented a trait in the same was PostProcessing was done, but to limit the disruption, there is a collection of lambdas running things.  Anyone concerned, please take a peek and send feedback.  Thanks.

MihaiSurdeanu:
I think it's pretty clean!

kwalcock:
@BeckySharp, can you tell me about the use of State in EidosSystem?  In the recent past it appeared twice:

```scala
val mostCompleteEvents = components.actions.keepMostCompleteEvents(extractedEvents, State(extractedEvents)).toVector

val afterHedging = components.hedgingHandler.detectHypotheses(cagRelevant, State(cagRelevant))
```
As far as I can tell, the state was not actually used in either case.  That constructor looks quite expensive, doing a mkLut across all the mentions passed to it.  I've removed the State argument in both of the calls above and changed the parameter to default to new State().  Is that safe?  Can I go one step further and just remove the parameters and not unnecessarily make a new Map with each new State?

BeckySharp:
> @BeckySharp, can you tell me about the use of State in EidosSystem? In the recent past it appeared twice:
> 
> ```scala
> val mostCompleteEvents = components.actions.keepMostCompleteEvents(extractedEvents, State(extractedEvents)).toVector
> 
> val afterHedging = components.hedgingHandler.detectHypotheses(cagRelevant, State(cagRelevant))
> ```
> 
> As far as I can tell, the state was not actually used in either case. That constructor looks quite expensive, doing a mkLut across all the mentions passed to it. I've removed the State argument in both of the calls above and changed the parameter to default to new State(). Is that safe? Can I go one step further and just remove the parameters and not unnecessarily make a new Map with each new State?

Yes -- looks like we're not using it
in the `mostCompleteEvents` case, it's in the EidosActions, if no one is using it _as an action_ (i.e., during the runtime of the grammars) then we can remove the state arg, and prob move the method to a diff place.
in the other case, I think it prob was originally an action in Reach, is not an action in Eidos, and I think it's safe to remove the arg

thanks!

src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
```diff
@@ -28,140 +24,153 @@ class EidosSystem(val components: EidosComponents) {
...
   //                                 Extraction Methods
   // ---------------------------------------------------------------------------------------------
 
-  def extractFrom(doc: Document): Vector[Mention] = {
+  def extractMentionsFrom(doc: Document): Seq[Mention] = {
 
-    def extractEventsFrom(doc: Document, state: State): Vector[Mention] = {
-      val extractedEvents = components.engine.extractFrom(doc, state).toVector
-      val mostCompleteEvents = components.actions.keepMostCompleteEvents(extractedEvents, State(extractedEvents)).toVector
+    def extractMentions(state: State): Seq[Mention] = {
```
BeckySharp: I think the Vectors were vectors bc the Reach team got bit by something lazy? Or something? I wasn't involved, but food for thought... you can check with marco if you want more info

src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
```diff
@@ -28,140 +24,153 @@ class EidosSystem(val components: EidosComponents) {
...
     //val cagRelevant = keepCAGRelevant(events)
 
     events
   }
 
-  def postProcess(annotatedDocument: AnnotatedDocument): AnnotatedDocument = {
-    val lastAnnotatedDocument = postProcessors.foldLeft(annotatedDocument) { (nextAnnotatedDocument, postProcessor) =>
-      postProcessor.process(nextAnnotatedDocument)
+  def preProcess(preProcessors: Seq[PreProcessing], odinMentions: Seq[Mention]): Seq[Mention] = {
+    val lastOdinMentions = preProcessors.foldLeft(odinMentions) { (prevOdinMentions, preProcessor) =>
```
BeckySharp: ok, I think this is clearly elegant, but any chance you can add a comment for the people on the team (with a range of scala experience) that have a hard time reading this syntax (including me!)
If I understand correctly, here we're taking the odinMentions and applying a each of the given preProcessing steps to each mention -- sequentially?
Perhaps we can include some examples in the comments too -- of what "preprocessing" on an odin mention might could look like?

src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
```diff
@@ -28,140 +24,153 @@ class EidosSystem(val components: EidosComponents) {
...
+  // Grounding is the first PostProcessing step(s) and it is pre-configured in Eidos.  Other things can take
+  // the resulting AnnotatedDocument and post-process it further.  They are not yet integrated into Eidos.
+  protected def mkPostProcessors(options: EidosSystem.Options): Seq[PostProcessing] = Seq(
+    components.ontologyHandler
+  )
+
+  protected def mkPreProcessors(options: EidosSystem.Options): Seq[PreProcessing] = {
+    Seq(
+      (odinMentions: Seq[Mention]) => { components.conceptExpander.expand(odinMentions) },
+      (odinMentions: Seq[Mention]) => { nestedArgumentExpander.traverse(odinMentions) },
```
BeckySharp: IDK what the consequences of expanding everything after running the grammars are.  If an argument is expanded in one "epoch/priority" of the grammar, then as a result other rules may apply differently.
Maybe this is fine, maybe it's not but can be fixed with tweaks to the grammars, etc.
Did you notice a difference in extracted event quantity?

src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
kwalcock: Thank you for all the invaluable comments!  I will check on it.

kwalcock: I sent the message.  I would think that at most only the first toVector is needed to "materialize" any lazy stuff.

src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
kwalcock: Good point and a better name would help.  How about a MentionFilter for Pre and AnnotatedDocumentFilter for Post?  No.  Something with Stage?  The previous method, as far as I could tell, was just a series of prevMentions -> filter -> currMentions -> filter -> nextMentions filter -> ...

src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
kwalcock: Maybe I missed something.  The functionality is not supposed to change.  The old code was
```scala
val afterExpandingConcepts = components.conceptExpander.expand(odinMentions)
val mentionsAndNestedArgs =  traverse(afterExpandingConcepts, Seq.empty, Set.empty)
val cagRelevant =...
```
Maybe I've confused things with a misnamed nestedArgumentExpander.  I didn't have a good name for that.  Suggestions welcome.

kwalcock:
> I left some comments -- I agree that it's way cleaner, but please see notes, I would like to know how the changes affect the output
> Thanks for working on this @kwalcock !

Output is not supposed to change at all.  So far it hasn't in regression or unit tests (that I have yet noticed).

kwalcock:
> > @BeckySharp, can you tell me about the use of State in EidosSystem? In the recent past it appeared twice:
> > ```scala
> > val mostCompleteEvents = components.actions.keepMostCompleteEvents(extractedEvents, State(extractedEvents)).toVector
> > 
> > val afterHedging = components.hedgingHandler.detectHypotheses(cagRelevant, State(cagRelevant))
> > ```
> > 
> > 
> > As far as I can tell, the state was not actually used in either case. That constructor looks quite expensive, doing a mkLut across all the mentions passed to it. I've removed the State argument in both of the calls above and changed the parameter to default to new State(). Is that safe? Can I go one step further and just remove the parameters and not unnecessarily make a new Map with each new State?
> 
> Yes -- looks like we're not using it
> in the `mostCompleteEvents` case, it's in the EidosActions, if no one is using it _as an action_ (i.e., during the runtime of the grammars) then we can remove the state arg, and prob move the method to a diff place.
> in the other case, I think it prob was originally an action in Reach, is not an action in Eidos, and I think it's safe to remove the arg
> 
> thanks!

Good point again on the State.  To preserve the interface yet stay efficient, I'll probably have each of these classes hold an empty State in reserve and reuse it rather than making a new one each time.

src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
BeckySharp: OH! you're right, sorry I think I was thinking about an older workflow that is no longer applied! Disregard!!!

src/main/scala/org/clulab/wm/eidos/EidosSystem.scala
BeckySharp: Maybe it could be something like odinMentionPostProcessors and eidosMentionPostProcessors?
That's a mouthful, but really they're both post-processing more than filtering, but some are done before converting to an eidosMention and some post -- but we can also keep the name as is, but just add a comment to this effect?
It really is much much cleaner!

BeckySharp: thanks for doing all this!

kwalcock:
Thanks for the comments.  I'm starting a different PR because of that merge conflict.

PR #764, opened 2020-02-21
Fiddle with EidosSystem

kwalcock:


kwalcock:
I didn't mean to go down this rabbit hole.  One thing led to another.  This is better documented.  I used the term "refine" because process is too overloaded.  It can be changed.

BeckySharp:
this is so beautiful!!!!

bgyori:
I hate to be a party pooper, I think the changes are nice, but the API of EidosSystem is changed in a way that might break integration with INDRA. I'll try to see if there are any issues and will post an update here on what I find.

kwalcock:
Who is that lurking?  Let me know if you find problems.  I tried to be careful, but may have broken something.

bgyori:
Just tested it, and it looks like you anticipated this issue by implementing
```
   // Legacy versions                                                            
   def extractFromText(                                                          
     text: String,                                                               
     cagRelevantOnly: Boolean = true,                                            
     dctStringOpt: Option[String] = None,                                        
     idOpt: Option[String] = None                                                
   ): AnnotatedDocument = {                                                      
     extractFromText(text, EidosSystem.Options(cagRelevantOnly), Metadata(this, dctStringOpt, idOpt))
   }                                                                             
```
which is the interface INDRA currently uses. This seems to work well, thanks!

kwalcock:
I'll be doing some regression testing yet and merging.  Thanks for checking it out.

kwalcock:
Here goes.

PR #766, opened 2020-02-25
Separate cache of vectors and ontologies

kwalcock:


BeckySharp:
oh yay!

kwalcock:
This version was a hack and I should probably do it better to avoid debt.  What do you think about useW2V -> useGrounding, useCache -> useCacheForOntologies.

kwalcock:
If we are working on ontologies and therefore do not want to use a cached version of them, it is still very useful to have the vectors cached and this change enabled separate policies for the two caches.  By default, however, caching of vectors follows caching of ontologies.  At some point in the future, the ontologies may be used to pare down the vectors and this system may need to be revisited.

PR #767, opened 2020-02-25
Changes to Exporter for Grounding Eval

BeckySharp:
Modified the grounding exporter to make it more relevant

FYI @MihaiSurdeanu @danebell 

BTW @danebell -- if you ever want to make a new export format you just need to specify a new Exporter, and add a line in ExportUtils to select which to load.  That said, this app makes one export file per input file, so we can cat them at the end to get one file if we want

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -39,7 +39,7 @@ apps {
   //    ----------------------------------------------------------
   //    mitre : a tsv designed for early MITRE evaluations (legacy)
   //    serialized : java serialized mention objects
-  //    grounding : a tsv format used to evaluate the grounding of the cause/effect
+  //    grounding : a csv format used to evaluate the grounding of the cause/effect
   exportAs = ["serialized", "jsonld", "mitre"]
 
   groundAs = ${ontologies.ontologies}
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -64,7 +64,7 @@ ontologies {
   includeParents = true
 
   // Activated Ontologies
-  ontologies = ["wm_flattened"] //, "wm_compositional"]
+  ontologies = ["wm_flattened", "wm_compositional"]
   path = ${EidosSystem.path}/ontologies
 
   // Primary
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -29,10 +29,21 @@ actions {
 }
 
 apps {
-  inputDirectory = "."
-  outputDirectory = "."
+  inputDirectory = "/Users/bsharp/data/WM/2020-02-14_wm-intervention-feb-2020-withRule/jsonld"
+  outputDirectory = "/Users/bsharp/data/WM/2020-02-14_wm-intervention-feb-2020-withRule/groundingeval"
   inputFileExtension = ".txt"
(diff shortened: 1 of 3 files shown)


BeckySharp:
@kwalcock this is a "no space left" issue

kwalcock:
I spent more than an hour on that issue today.  It's going down hill.
Almost every day.

BeckySharp:
Sorry Keith! I guess we all need something really annoying in our lives to
help us appreciate all the rest?
More seriously -- is there stuff on river that we can delete and give you
more space? or is that a bandaid on a broken leg?

MihaiSurdeanu:
I think I found the main culprit for the used space on river. I will email him tonight and cc Keith to make sure cleanup is happening.

MihaiSurdeanu:
@kwalcock : if you can, please do not invest more time in this. This is something that you can't solve. But it has an easy solution; just needs implementing by the owner of that data.

MihaiSurdeanu:
Ok, spoke too soon.
Good news: I did manage to free up significant space on the local /data1 partition on river. We now have 1.5TB free there, with more to come soon.

BUT, root has a separate / partition, which is still at ~100% use, despite my trying to free up everything I could, from unused packages to old kernels... Plus, I am not sure where the big use happens, as no top folder comes up as big... I am lost. @kwalcock: any ideas of folders that take too much space on /? The good news is that we could move them to /data1 (and symlink to them) to free up space on /. I just don't know what to move.

BeckySharp:
a lot in in the `/var/lib//docker/volumes` folder (i.e., ~60G) -- but is
that just where jenkins drops stuff?

(base) bsharp@river /var/lib/docker]$ sudo du -hs volumes
60G volumes

kwalcock:
A couple weeks ago I didn't see much that could be removed on river, either.  Jenkins doesn't seem to be deleting old builds correctly, so if/when it fills up I do that--and also figure out how to automate it better.  Also the /tmp directory inside the Docker volume fills up.  This may be in part because of sbt leaving stuff there as it does elsewhere.  I usually delete that dir.  The Jenkins job don't restart themselves, so that's a required manual intervention.  If the disk has filled so that their are corrupted files or half completed build stages, usually the next jobs won't start, either, and that takes work.  Anyway, I try to spend as little time as possible on it.  There may be some old Docker containers and volumes that can be removed.  I'm sending a list by email in case they look familiar.

MihaiSurdeanu:
Good find! Thanks!

I think we move /var/lib/docker to somewhere under /data1 and create a symlink to it in /var/lib/docker. I can do this.
BUT, what are we doing with docker on river? Are we using it directly? Is it part of something else? 
In other words, does somebody need to stop it before I move this folder?

kwalcock:
Yay!  Finally green.  Let me know if I should run this.  I figure that others are better able to notice if something has gone wrong, though.

BeckySharp:
> Yay! Finally green. Let me know if I should run this. I figure that others are better able to notice if something has gone wrong, though.

Do you mean merge? I think you're good to merge, we can't use it until the ontology node files are ready (and we know which documents -- already read or not -- to run it on, but I think we know that much already, right?)

kwalcock:
I should have mentioned, before accidentally merging this, that there is a class TsvUtils with a very few useful functions.  It appears to be slightly outdated when compared to the LUM version.  I should do some backporting and maybe accommodate commas.

PR #769, opened 2020-02-29
Exporter fix

BeckySharp:
small bugfix on exporter i had done ahead of the last eval but forgot to push, also a python script for concating the eval files together for uploading to gdrive.

### DIFF
--- a/src/main/python/concat_grounding_eval_files.py
+++ b/src/main/python/concat_grounding_eval_files.py
@@ -1,0 +1,29 @@
+from glob import glob
+import argparse
+import os
+import sys
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('wdir', help='directory with the grounding eval files')
+    args = parser.parse_args()
+    return args
+
+
+
+
+if __name__ == '__main__':
+    args = parse_args()
+    files = glob(os.path.join(args.wdir, '*.ground.csv'))
+    print(files[:10])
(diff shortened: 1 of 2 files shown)


kwalcock:
A PrintWriter should flush itself automatically on close.  Did you observe otherwise?  Certainly won't hurt though.

BeckySharp:
> A PrintWriter should flush itself automatically on close. Did you observe otherwise? Certainly won't hurt though.

I did, but to be fair, I added the flush and close at the same time, I for some reason assumed the PrintWriter access method in Eidos made one that was "closeable" (or whatever the name was), I thought I had remembered you adding a bunch of syntactic sugar stuff so that things closed themselves, but I assumed wrong...

kwalcock:
You're keeping me honest.  I should have looked harder.  Rather than passing in a prefabricated PrintWriter into the constructor of that Exporter, you should probably just pass in `filename + ".ground.csv"` from ExportUtils.getExporter.  Then in the export function, locally do the
```scala
    import org.clulab.wm.eidos.utils.Closer.AutoCloser
    
    FileUtils.printWriterFromFile(filename).autoClose { printWriter =>
      // Header
      pw.println(header)
      annotatedDocuments.foreach(printTableRows(_, pw, reader))
      pw.println()
    }
```
Does that make sense?  close() to me is one of those code smell things.

kwalcock:
I'm handling the nitpicky thing above along with adding some CSV utilities that were almost ready to commit anyway.

PR #771, opened 2020-03-03
Zupon grounding eval

zupon:
Two apps to evaluate different grounders.
1. `GenerateGoldGroundingTSV` creates the initial tsv file to fill in with gold groundings.
2. `EvalGroundings` takes that filled in tsv and compares the current groundings with the gold groundings. It returns a summary accuracies for each grounder/branch and also makes a new tsv file with the gold and current groundings, along with information about whether they are correct or not.

Should I push the tsv files that I have already generated? If so, where do they belong?


BeckySharp:
can i defer to Keith on this?

kwalcock:
can i defer to Becky on this?

kwalcock:
There are some potentially useful Tsv and Csv utilities waiting to be merged right now.  As you continue working in this, you may want to migrate.  They'll especially do some escaping for you.

kwalcock:
If I understand correctly, this works by doing an eidosSystem.annotate(text) and then regrounds using the Document produced.  I was trying to make do with sentencesExtractor.extractDocument(sentenceText).  I need to see how their steps differ and modify extractDocument to take care of those missing steps.  This was at one time optimized to only perform the steps necessary for the other grounders.  Thanks for figuring it out.

kwalcock:
I thinks it's OK and will merge momentarily unless someone hollers STOP!

PR #773, opened 2020-03-03
Fix texts in compositional grounding

kwalcock:
Words go on per array index into w2v.calculateSimilarities


kwalcock:
This hasn't been tested, but you are watching quite closely and will very likely notice whether it is effective or not.  We're still working on getting unit testing enabled with grounding and then it will be much easier to notice if something gets broken.

kwalcock:
I'll check this in.  @zhengzhongliang, @zupon, you may (or may not) want to merge this into your branches for your testing.  You may want to "cherry pick" this change and not take everything that has gone into master in the interim.

PR #775, opened 2020-03-04
Rule mods

### LINKED ISSUE #325: Extraction of causal paths, preserving implied order
While reading some relevant documents, I encountered a very common pattern of "A leads to B via C".  Which implies the causal chain A->C->B.

A real example is:
```
Climate change and natural disasters, it has been argued,
represent a major threat to national and international security 
by increasing resource scarcity and competition and inducing health problems
```
A simple synthetic example to better highlight the issue is 
```
rainfall causes food insecurity by increasing floods
```
for which we currently get "rainfall" -> "food insecurity" and "rainfall -> "flood", equivalent to A->C and A->B given the pattern above. Would it make sense to extract this pattern as A->C, C->B instead? From the perspective of modeling, preserving the causal chain of these relationships would be very valuable.

BeckySharp:
- Modifications to the rules (and tests) for causal extraction.
- fixed the grounding, now using words instead of lemmas to match the glove embedding.
- Changed canonicalizing to use same function for both initial grounding and regrounding, though there is still a FIXME for a later time.
- Added MISC to the NER stop list
- updated the reference.conf to match changes in eidos.conf (they had diverged)

FYI @MihaiSurdeanu @danebell 

closes #325

### DIFF
--- a/src/main/resources/englishActionsExpander.conf
+++ b/src/main/resources/englishActionsExpander.conf
@@ -11,6 +11,7 @@
     "^advcl_because",
     "appos",
     "^case",
+    "ccomp",
     "^conj",
     "^cop",
     "^cc$",
--- a/src/main/resources/org/clulab/wm/eidos/english/filtering/stops.txt
+++ b/src/main/resources/org/clulab/wm/eidos/english/filtering/stops.txt
@@ -71,7 +71,6 @@ achievement
 activity
 addition
 adjustment
-administration
 agent
 aim
 algorithm
@@ -148,6 +147,7 @@ recognition
 referral
 role
 sample
+section
 series
 situation
 size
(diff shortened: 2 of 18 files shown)


kwalcock:
@BeckySharp, I finally see how the mentions are getting canonicalized with the words rather than the lemmas.  I don't see that a similar change has been made to the ontologies, in the methods realFiltered, for instance.  The texts will have the words rather than lemmas looked up in the vectors file, but the ontologies seem to still have the lemmas looked up.  Does that need to change as well?  I hope to have unit tests for grounding soon so that I can keep track of things like this.  In the meantime... can you check that I'm not imagining things?

BeckySharp:
Yeah -- IDK -- I guess we should use the words there too, but I don't think
people are particularly going to use all word forms, but sure.
This is not critical, but we should be consistent.  Going to submit a PR
with a tentative fix you can tweak to make scalaier

BeckySharp:
submitted

PR #785, opened 2020-03-11
Export class, triggers, and a stop

BeckySharp:
Main diff is to move around the Exporters and rename the package. 
I also made a package object and removed the ExportUtils object

Small diffs:
- removed a couple of triggers and added a stop word


kwalcock:
Great idea.  I noticed the step in this direction with the most recently added exporter.  BTW package objects are slated for deprecation and disappearance in Scala.  I wonder how it would be to put MitreExporter in its own file and moving most things from package.scala to an Exporter object (if that's allowed).  A couple of import statements would probably have to change, but the mapping of classes to files would be more canonical.  I'd volunteer to do it if need be.

kwalcock:
Should I worry about the test that went from passing to failing?

BeckySharp:
> Great idea. I noticed the step in this direction with the most recently added exporter. BTW package objects are slated for deprecation and disappearance in Scala. I wonder how it would be to put MitreExporter in its own file and moving most things from package.scala to an Exporter object (if that's allowed). A couple of import statements would probably have to change, but the mapping of classes to files would be more canonical. I'd volunteer to do it if need be.

thanks for the heads up! yes, that would be awesome if you took a stab at it! I would have moved MitreExporter but I want to deprecate it so it seemed less worth it but i guess files are free right?

and nope, no worries about the test that went to failing, i removed a trigger bc it seems to be really low precision, but as a consequence there's a failing test

thanks!

PR #786, opened 2020-03-11
up to and including migration as finders

BeckySharp:
up to and including migration as finders (to the `Context` and `Migration` grammars), so they can be disabled by omitting from the config.
This hasn't been tested by me -- @kwalcock I have to run, do you have an easy way to test this?

fyi @MihaiSurdeanu @zupon @maxaalexeeva


kwalcock:
Cool!  I hadn't realized how that worked.  Probably I should remove my garbage around the keepMigrationEvents setting.  This looks like the right way to do it.

kwalcock:
A bunch of tests failed, the first ones related to migration.  I suspect it has to do with the test configuration file overwriting the value for entityFinders.  I can figure it out as long as you don't do it at the same time.

BeckySharp:
> A bunch of tests failed, the first ones related to migration. I suspect it has to do with the test configuration file overwriting the value for entityFinders. I can figure it out as long as you don't do it at the same time.

it's all yours! :) thanks!!

kwalcock:
Among the many things I don't understand are Finders and how their rules differ from the rules used by the ExtractorEngine.  We just moved the migration rules out of a file used by the ExtractorEngine and put it into a file used by an Extractor.  It seems like the Finders all run just once, in EidosSystem.mkMentions.  They run in the order specified in the config file, independent of priority, and they are used for the initial state.   Thereafter, there is a components.engine.extractFrom that runs.  My impression is that this runs through all the rules from master.yml, in priority order, multiple times, because a rule could make use of something that is figured out by a later rule or by an earlier running of itself.  This all loops until nothing changes anymore.  I don't know the context and migration rules that have been moved from one place to another.  It doesn't seem like just any rule could be moved.  They would have to work when only one pass of the context rules had been made.

BeckySharp:
> Among the many things I don't understand are Finders and how their rules differ from the rules used by the ExtractorEngine. 

In principle they don't, we could (and maybe *should* make everything a Finder), but I went through migration so there would be an easy way to disable them.

> We just moved the migration rules out of a file used by the ExtractorEngine and put it into a file used by an Extractor. It seems like the Finders all run just once, in EidosSystem.mkMentions. They run in the order specified in the config file, independent of priority, and they are used for the initial state. Thereafter, there is a components.engine.extractFrom that runs. My impression is that this runs through all the rules from master.yml, in priority order, multiple times, because a rule could make use of something that is figured out by a later rule or by an earlier running of itself. 

Yes, but about the multiple times, IDK if it does by default, we'd have to ask @marcovzla or someone.  I thought I remembered that if you wanted an Odin rule to run over and over you had to make the priority something like `2+` vs `2` or something...

> This all loops until nothing changes anymore. I don't know the context and migration rules that have been moved from one place to another. It doesn't seem like just any rule could be moved. They would have to work when only one pass of the context rules had been made.

This is true -- and perhaps I should dig into this to make sure that it's not the source of the problem... Using the thing above I could probably make them run several times if they aren't already.  But in both the Finder I made and the main "master" rules we use for causality, there is an Odin extractor engine doing the extraction, so they should be equivalent.

kwalcock:
Thanks @BeckySharp.  Lacking time to look at it carefully, I was just grasping at straws.  For the record, here's the description of priority below.  Since you are having a new ExtractionEngine running the OdinFinders, it seems like the rules would be run multiple times as usual.  Migration is priority 2 and that used to be shared with entityQuantification and temporalAttachment in master.yml.  Now those two would be running separately from and after migration, which runs before them.  Could that be a problem?

priority The iterations in which this rule should be applied. Note that the
Odin runtime system continuously applies the given grammar
on a given sentence until no new rule matches (this allows grammars
that use recursive events, such as the one in Example 1
to work). Each of these distinct runs is called an “iteration”,
and they are all numbered starting from 1. Through priorities,
a developer can specify in which iteration(s) the corresponding
rule should run. Specifying priorities is not required, but it may
have an impact on run time, by optimizing which rule should
be applied when. A priority can be exact (denoted by a single
number), a range (two numbers separated by a dash), an infinite
range (a number followed by a plus +), or a list of priorities (a
comma separated list of numbers surrounded by square brackets.

BeckySharp:
i don’t think it’s the problem or a problem, but good to point out, thanks!

maxaalexeeva:
@BeckySharp @kwalcock, another fun finding: 

In the `as_finders` branch, numbers are found as just concepts, while in other branches (master, our old migration_schema-zupon), they were found as concepts and entities, and migration rules are written such that groups have to be entities. Do you know why numbers stopped being entities? If that was intentional, then changing migration rules to accept `Concepts` for group will fix some of the tests (still not the expansion-related ones). If that was not intentional, I can keep looking for why that happened.

kwalcock:
Thank you for the detective work!  (That's about all I can say about it just now.)

BeckySharp:
Ill look, this is prob important

kwalcock:
Thanks for digging into this.  Those look like some things I should have noticed or maybe even I caused them.

BeckySharp:
I disagree - I feel like I should have found it waaay sooner, but if you want to dig into something :) you can look at the current errors -- with JLD serialization? I don't think they were errors before...? so maybe I broke something...

kwalcock:
I can probably check a little later today.  JLDSerializer sounds like a job for me.

kwalcock:
It may be broken.  In each case the exception is thrown because there was no "rule" that can be used for the "foundBy" parameter of a Mention.  Is that enough of a hint?

```json
  "extractions" : [ {
    "@type" : "Extraction",
    "@id" : "_:Extraction_1",
    "type" : "concept",
    "subtype" : "entity",
    "labels" : [ "Concept", "Entity" ],
    "text" : "300 refugees",

*** there should be a rule here ***

    "canonicalName" : "refugees",
    "provenance" : [ {
      "@type" : "Provenance",
      "document" : {
        "@id" : "_:Document_1"
      },
      "documentCharPositions" : [ {
        "@type" : "Interval",
        "start" : 0,
        "end" : 11
      } ],
      "sentence" : {
        "@id" : "_:Sentence_1"
      },
      "sentenceWordPositions" : [ {
        "@type" : "Interval",
        "start" : 1,
        "end" : 2
      } ]
    } ],
```

BeckySharp:
how is that possible?

BeckySharp:
maybe something in the expansion? i'll start looking there

kwalcock:
One possibility is that the serializer didn't write them out for some reason, but that doesn't appear to be the case.  I see in the test of p5s1 that there is a mention found with `oldCorpus(0).eidosMentions(2).odinMention.text = "more than one in five people in South Sudan"` and labels of (Concept, Entity) that has an empty string for foundBy.  It's an EidosTextBoundMention. 
 Its attachments are a COUNT and DCTime.  Something isn't filling in that rule.  Maybe it was in the output before but the particular mention was removed as a duplicate and its twin did have a rule.  I see in the equivalenceHash for the TextBoundMention that foundBy is not taken into account.  Something may be making a copy and losing the rule.

kwalcock:
A conditional breakpoint on package org.clulab.odin.TextBoundMention line 195 with foundBy == "" does catch something in ExpansionUtils.attachDCT after the ArgumentExpander.  Getting closer...

kwalcock:
I notice that in ExpansionUtils.addSubsumedAttachments, compositionalFoundBy is being called with overlapping being empty.  Then completeFoundBy is empty and MentionUtils.withFoundBy is called with no rule.  That would cause a problem.  I hope that gets someone further.

BeckySharp:
Thanks! This is helpful!

BeckySharp:
Keith you're soooooooooooooooo good at this!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
THANK YOU SO MUCH!!!!
Thanks for finding this bug of mine!

image

kwalcock:
Hey @BeckySharp, did you replace
```
    expander = ${actions.expander}
```
with
```
    expander = { include required("englishActionsExpander.conf") }
```
because the former didn't work or was it some other reason?  It seems like the first version would be safer.

BeckySharp:
oh, it was in the process of debugging, if tests pass with the former,
let's revert
can you check?

kwalcock:
I wondered and checked, mostly to make sure that I understood what was going on.  There is PR in the works.

PR #787, opened 2020-03-11
Rearrange exporter code

kwalcock:


kwalcock:
This includes unnecessary but cool modifications to TsvWriter to support a few more of the apps which had to be touched anyway.  IntelliJ is smart enough to strike through the deprecated class.  That's some motivation to get rid of it!  Exporter has a nice apply method now.

PR #788, opened 2020-03-11
Hard-code rule priority to allow parse

kwalcock:
Remove keepMigrationEvents from config files

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -24,7 +24,6 @@ actions {
   useCoref = true
   corefType = "causalBasic"
   taxonomyPath = ${EidosSystem.taxonomyPath}
-  keepMigrationEvents = false // Turn this on when documents are known to deal with migration.
   useExpansion = true
 }
 
--- a/src/main/resources/org/clulab/wm/eidos/english/grammars/migration.yml
+++ b/src/main/resources/org/clulab/wm/eidos/english/grammars/migration.yml
@@ -4,7 +4,7 @@ vars: org/clulab/wm/eidos/english/grammars/vars.yml
 rules:
 
   - name: migration-verbs
-    priority: ${ rulepriority }
+    priority: 2
     example: "Since the beginning of September 2016, almost 40,000 refugees arrived in Ethiopia from South Sudan as of mid-November."
     label: HumanMigration
     action: migrationActionFlow # parse out quantity as an attachment
@@ -21,7 +21,7 @@ rules:
 
   - name: migration-arrive
     # arrive behaves differently with nmod_in than other migration verbs
-    priority: ${ rulepriority }
+    priority: 2
     example: "Since the beginning of September 2016, almost 40,000 refugees arrived in Ethiopia from South Sudan as of mid-November."
     label: HumanMigration
     action: migrationActionFlow
@@ -39,7 +39,7 @@ rules:
 
   - name: migration-displace
     # displace acts differently a lot of the time
-    priority: ${ rulepriority }
+    priority: 2
     example: "For many FGD respondents from Greater Equatoria, this was the second or even third time they or their family had been displaced out of South Sudan; many reported leaving during the 2013 displacement crisis and/or in the second Sudanese Civil War (1983-2006)."
     label: HumanMigration
     action: migrationActionFlow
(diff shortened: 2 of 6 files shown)


kwalcock:
This fixes the configuration problem, but still a lot of tests are failing in TestMigrationSchema.  Most seem to be related to counts.

BeckySharp:
hmmm... ok, I can try to dig in...

BeckySharp:
or maybe --- @zupon can you or @maxaalexeeva (IDK who worked on the counts) look and suggest somewhere to dig in?

maxaalexeeva:
Yes, I can try to look at that tomorrow!

maxaalexeeva:
Becky, I mean we can dig in---no need for you to at least yet.

kwalcock:
I'll keep looking as well then.  I wondered if it would ring a bell.

BeckySharp:
thanks all!!

maxaalexeeva:
wanted to start looking, but I guess now eidos comes with a glove vectors download (?) and my current internet is super slow (and I don't have access to jenkins output). Hope they will download by tomorrow and I can have a look at what's going on.

kwalcock:
We can get you access to Jenkins.  It will be better if you can use a debugger, though.  You can comment out the glove dependency in build.sbt.  Then turn off ontologies.useW2V in eidos.conf.  Vectors aren't needed then.

maxaalexeeva:
Thanks, @kwalcock!

maxaalexeeva:
@kwalcock, I am still looking. There are some cases where it looks like some rules just didn't work---I will have to look at those later (either tonight or tomorrow at the latest), but there are two test sentences* where it looks like the group just did not expand. Could that have anything to do with the separation of expanders into concept and action? (I can investigate more later but have to go now). 

_85% of the new arrivals originated from Upper Nile State (Nasir, Longechuk or Mathiang, Ulang and Maiwut Counties), whilst 14% came from Jonglei State (Uror, Akobo and Ayod Counties)_

The tests related to the sentence with at least some nodes failing: _migration-originated_ through _migration-originated-5_ (with the exception of migration-originated-2, which was ignored anyway)

_Civilians caught in the fighting between armed actors first displaced into remote areas of the bush, then flee to Uganda, Kenya or the DRC over the course of a few months._

Relevant failing tests: _migration-supplemental-3_

BeckySharp:
ok, expansion seems like a good place to start! and also maybe the global
action.
thanks -- are you still digging or should i tomorrow?

maxaalexeeva:
I should be able to look more tomorrow late-ish afternoon, but if you happen to have  some time before then, feel free to have a look especially if you're familiar with the expander changes. Thanks!

BeckySharp:
I went ahead and merged so we can consolidate things, but seems the issue is that the tests were written to have things like "85%" be represented as:
`val group1 = NodeSpec("85% of the new arrivals", CountSpec(85, CountModifier.NoModifier, CountUnit.Percentage))`
But what we get now is:

```
List(Concept, Entity) => 85% 
     Rule => simple-np 
     Type => TextBoundMention 
     Concept, Entity => 85% 
     * Attachments: COUNT("85 %", value=85.0, mod=NoModifier, unit=Percentage) 
```

Was there something explicitly being done to remove the `%` from the text? Or should we just change the test?

maxaalexeeva:
@BeckySharp the intention with the group was to extract the complete natural lang info about the group (e.g., 85% of the new arrivals) and then get the info about count using a couple of actions (e.g., this https://github.com/clulab/eidos/blob/6ecb938b3a75c2ce72af11e9714e02ed897836e5/src/main/scala/org/clulab/wm/eidos/EidosActions.scala#L93). Sometime between merging migration branch into master and now, we lost the expansion from 85% to _85% of new arrivals_ (the test was passing). We could delete the new arrivals part to have the test pass, but that is not what the test was intended as and something might be wrong.
(A similar lack of expansion is with the civilians example---only _civilians_ was extracted and not the _cuaght in the fighting_)

I am hoping to look into this today more. Is there a pressing deadline on this? (e.g., we need this tomorrow)

BeckySharp:
nope no deadline, and i into the expansion, thanks for letting me know the
intention

PR #790, opened 2020-03-12
regrounding exporter ++

BeckySharp:
adds wm_flat ontology in a semi-hack way for now...

@kwalcock is this reverting something in src/main/scala/org/clulab/wm/eidos/groundings/EidosWordToVec.scala ?


kwalcock:
I saw this and will get to it soon.

Issue #783, opened 2020-03-09
Adjective grounding needs to be added to the pipeline


kwalcock:
At one time and for various reasons the design had no vars for an EidosMention.  This was recently changed so that the canonicalName and groundings are vars.  I realize now that there is one more dangling piece of data: the adjective groundings.  These are being added optionally just for the jsonld export and they're the reason that in lots of apps there is this strange adjectiveGrounder being constructed in order to do the serialization.  It's quite a distraction.  IMO the adjective grounding should be added to the main pipeline.  If necessary, it could be turned off by default via the options.  This means that the serializer doesn't need to know anything about the grounding.  The deserializer can be made to read it back in if present.  The information is otherwise lost and we've been reconstituting on round trips.  The adjective grounder would become part of the EidosComponents and potentially accessible there for anything that still needs its own.  It would even be guaranteed to be the same one that the reader used.  Ensuring that was previously reason for excessive handling of config variables.

bgyori:
Done in #748

bgyori:
Oh no, so

kwalcock:
#791 addresses this issue.

PR #791, opened 2020-03-13
Clean up adjective grounding

kwalcock:


kwalcock:
The example on the readme page will changes from

```scala
  // Export to JSON-LD
  val corpus = new JLDCorpus(Seq(annotatedDocument), reader.loadableAttributes.adjectiveGrounder)
  val mentionsJSONLD = corpus.serialize()
```
to
```scala
  // Export to JSON-LD
  val jsonld = new JLDCorpus(annotatedDocument).serialize()
```

kwalcock:
This addresses #783.  Adjective grounding can be turned on and off in the configuration file.  The adjective groundings have been added to TriggeredAttachments.  EidosSystem does a pass over the EidosMentions to perform the grounding if it is configured.  The serializer just serializes what it has and doesn't do any additional processing.  Lots of apps and examples have been simplified.

I just remembered that the deserializer needs to be updated, so stand by.

kwalcock:
@bgyori, do you ever call serialize on a JLDCorpus or other JLDObject with an adjectiveGrounder?  The need, and also the ability, to do that is being removed.  EidosSystem knows enough to do that on its own with this PR.

bgyori:
The adjectiveGrounder will not affect what we do, however, the fact that 
```
JLDCorpus(Seq(annotatedDocument))
```
will change to 
```
JLDCorpus(annotatedDocument)
```
will need to be changed in INDRA after this is merged.

kwalcock:
Oops, maybe.  A Corpus is a Seq[AnnotatedDocument], so at least from Scala we can still call
```Scala
JLDCorpus(Seq(annotatedDocument))
```
through
```Scala
def this(corpus: Corpus) = this(new JLDSerializer, corpus)
```
Will it not work from Python?

kwalcock:
That seems to be the only one that has ever been available, through a Corpus.  There was never explicitly one for a Seq[AnnotatedDocument].  I don't *think* you have need to change code.

bgyori:
Alright, in that case we may not need to do anything further on our end. Thanks!

PR #792, opened 2020-03-16
Zupon grounding eval

zupon:
Added grounding eval tsv files to "resources" directory". Changed paths in apps to match.


zupon:
@kwalcock This also updates the local version of `wm_compositional_metadata.yml` to match the one in the Ontologies repo. If that's not needed/wanted, I can remove it from this PR.

kwalcock:
The app must have been run on a PC and with git not configured to take care of EOL irregularities.  This results in CRLF endings finding their way into source code or resources, which is generally a bad thing.  Usually we (I) stick to println and make sure git fixes it.  There are other options such as using print and print("\n").  The git settings for this are confusing and I don't know the right one offhand.  In GitKraken, it is a general setting called AutoCRLF.  HTH.  Do ask if not.

```
[info] - should not have any CrLf line endings in /var/jenkins_home/workspace/eidos_zupon-grounding-eval/src/main/resources/org/clulab/wm/eidos/english/grounding/gold_groundings_annotated.tsv *** FAILED ***
[info]   true was not false (TestCrLf.scala:38)
[info] - should not have any CrLf line endings in /var/jenkins_home/workspace/eidos_zupon-grounding-eval/src/main/resources/org/clulab/wm/eidos/english/grounding/groundingEvalEntities.tsv *** FAILED ***
[info]   true was not false (TestCrLf.scala:38)
```

zupon:
@kwalcock I tried changing the line endings from CRLF to LF in IntelliJ, but that didn't seem to pass the check (unless there's something different now, but I can't see the Jenkins output).

kwalcock:
Stand by.  I was in the middle of a response.  Also we need to get you access to Jenkins.

kwalcock:
One has to click on [Full Log](http://jenkins.cs.arizona.edu:8090/job/eidos/job/zupon-grounding-eval/2/consoleFull) to see the problem with Unicode characters.  The test that fails is org.clulab.wm.eidos.system.TestResources, which you can probably run locally to see the line numbers with the problematic characters.  I see in the code a `val pw = new PrintWriter(outFilename)` which AFAIK may not be correct for Unicode.  (Environment variables can come into play so that the result is different on different computers.)  For that in FileUtils there's
```scala
  def printWriterFromFile(file: File): PrintWriter = Sinker.printWriterFromFile(file, append = false)

  def printWriterFromFile(path: String): PrintWriter = Sinker.printWriterFromFile(path, append = false)
```
that should be used just in case.  However, that's probably not sufficient.  The test will likely fail for even correct Unicode characters.  It looks like they are mostly bullet kinds of symbols and if it's easy to do, you might just remove them.  If not, you can exclude the files from the test by adding to `unwantedSuffixes`, for example.  In my experience, Excel has difficulty importing and exporting Unicode characters especially to simple tsv and csv files and gets different results when they are read from a file and when they are pasted in from the clipboard, so avoiding them can be the better way to go.

kwalcock:
If you are tired of writing `+ "\t"` and related code, you might consider using the TsvWriter.  I don't see any escaping being done and maybe you're certain there are no problems, but you get it for free with TsvWriter.  Excel likes to have commas escaped even if it is tsv and not csv.  It's a recent addition, so it's easy to have missed.  You might want to check for examples of autoClose and AutoCloser if you don't like having to close things.  Now I see you are reading as well.  There's a TsvReader for that.

kwalcock:
Of all the things you probably don't want to know about...  In `GenerateGoldGroundingTSV`

```scala
  val pw = new PrintWriter(outFilename)
  pw.println(header)
```
will get you a \r\n on some platforms.  Later on there is
```scala
        propertyScore + "\n"

      pw.print(row1)
```
which will get you \n.  Because these are project files, git comes into play, and it is trying to unify line endings.  Strange things can start to happen.  That TsvWriter is supposed to deal with it.  I can fix if this is tiring.

zupon:
It sounds like you know what all the right pieces are and how to put them. If you have the time to make some of these changes, I would really appreciate it. I _could_ spend my entire day trying to fix unicode problems, but you can probably get it done much faster (and much better).

kwalcock:
Maybe I should make you a branch and send a PR.  In one file I see
```scala
  val ontologyHandler: OntologyHandler = new
      EidosSystem().components.ontologyHandler
  val ieSystem = new EidosSystem()
```
which really looks like two instances of EidosSystem being created.  This seems like it could slow down the start up time by quite a bit and increase memory requirements...
OK, I'll take the comment that just arrived as permission to send in a PR instead.  I was trying to resist the temptation.  Alas.

kwalcock:
P.S. I recall my needing to turn this into a unit test, so I'll try to include that then.

kwalcock:
In EvalGroundings.scala there is code in this pattern:
```scala
      breakable {
        for (i <- top5flat.indices) {
          if (flatGroundingGOLD.contains(top5flat(i))) {
            flatGroundingsCorrect += 1f/(top5flat.indexOf(top5flat(i))+1).toFloat
            flatCorrect = true
            matchedFlat = top5flat(i)
            break
          }
        }
      }
```
I'm pretty sure that `top5flat.indexOf(top5flat(i)) == i` and that you really mean something more like `flatGroundingGOLD.indexOf(top5flat(i))`.  Can you tell me that I haven't gone crazy here?

I may have to eat my hat, but this change brings the numbers from
```
flat grounding accuracy:	0.7135416
concept grounding accuracy:	0.5197916
process grounding accuracy:	0.45
property grounding accuracy:	0.7875
```
to
```
flat grounding accuracy:	0.8166666
concept grounding accuracy:	0.64375
process grounding accuracy:	0.45625
property grounding accuracy:	0.7875
```

kwalcock:
And I may have to eat the hat.  Do you want the i, the index into the top5 of the first item that is in the golds, or do you want the index into the golds of the first item that was matched from the top5s?  I was thinking the later until I reread the comment `// keep track of matches, adding MRR score based on position in top 5`.  I'm doing position in gold.  Either way, `top5flat.indexOf(top5flat(i))` needs to change.

MihaiSurdeanu:
I think Keith is correct, as usual :)

For MRR, we need to compute 1/(index in flatGroundingGOLD), if I understand
the code correctly. So we need flatGroundingGOLD.indexOf(top5flat(i)).

kwalcock:
There's a PR #793 pending with the fix.  It will be updated or replaced after I look at GenerateGoldGroundingTSV and rearrange some things for a unit test version.

zupon:
As for the MRR part, shouldn't it be `1/(indexOf in top5flat)` ? Most of the gold groundings only have one entry, and we are trying to check in our current groundings (top5flat) which one matches the GOLD grounding.

If we do `1/(indexOf in GOLD)`, it will (almost) always end up `1/1`, which will improve our scores, but might not be doing what we actually want. NB: I could very well be completely wrong on this.

MihaiSurdeanu:
@zupon : you're right. Please ignore my comment.

kwalcock:
OK, then false alarm.  The value should be i then.

kwalcock:
@zupon, I had to remove the local `wm_compositional_metadata.yml` file because it conflicted with the one coming from github, especially for people who run build.sbt through the assembly stage.  It should be gone when this branch gets merged with master.

zupon:
@kwalcock I changed one minor thing in `RegroundExporter` from an `override def` to just a `def` because it wasn't overriding anything (and thus failing a Jenkins check). If that sounds ok to you, we can merge this in at last!

kwalcock:
OK.  Hold your breath.  The button is being pushed.

PR #793, opened 2020-03-17
Assimilate EvalGroundings.scala

kwalcock:
Remove extra comma in input data, change MRR calculation


kwalcock:
@zupon, for the Uncode I replaced most strange characters that probably came from bullets with * and there was one other which might have been a long dash which I replaced with -.  This was done in part so that no offsets changed.  Since the original characters were wrong, I decided not to exempt the files from the test.  I hope it doesn't cause problems.

kwalcock:
@zupon, I'm prepared to send this code back to the collective.  Please merge it if it looks reasonable (or send comments if not) and then it can be fixed up more in your branch before it hits master.  Thanks.

zupon:
Looks good to me! I'm glad you're here, @kwalcock , to deploy all the super cool things that I never knew existed.

PR #801, opened 2020-03-25
Just comment it out for now

kwalcock:
Regression test on just 10 documents passed


kwalcock:
It didn't seem to break.  I left the line there because there's going eventually to be a merge conflict with the LUM version.

PR #805, opened 2020-03-31
Egoitz geolocation grammars

EgoLaparra:
Putting rule for geolocation attachment back.

Some unit tests in `TestMigrationSchema` fail. It seems that duplicate mentions with location attachments are produced but I couldn't say why.


BeckySharp:
thanks! I think it's likely the same issue as whatever is happening in #786 -- so hopefully I can kill 2 birds with one stone, assuming I can find the stone! :)

BeckySharp:
ok @kwalcock @EgoLaparra and @maxaalexeeva 
I'm having trouble debugging this because in the webapp seemingly we get them:
e.g., 

image

for failing test in `migration-az-hack-9`

is it some kinda of interaction with the tests/test utilities?

kwalcock:
Now I know why you answer email during meetings.  If you don't, you'll never catch up.  I'll get to this soon.

EgoLaparra:
I'm not sure about that example but I checked  `Departures: 375 individuals were recorded leaving Juba (52%) of which the vast majority intended to reach refugees camps in Uganda (75%)` and it failed when testing the `humanMigrationEdgeSpec` because the mention for `Juba` in `testResults` was `Concept Entity` instead of `Location`.

I think that the system creates two mentions for concepts with location attachments, one `Concept Entity` and one `Location`. When the `NodeSpec` is tested it always passes successfully but only one of the mentions is kept in `testResults`. If the mention that is kept  is `Concept Entity` the test for the migration event fails.

Does it make any sense?

BeckySharp:
oooh -- ok, that is useful, thanks!!!!!!!! I will look into this, Keith you
may be off the hook, thanks Egoitz!!!

kwalcock:
A late start will do wonders!  Continued good luck!

BeckySharp:
Yeah. this is the problem.  I have racked my wee brain trying to come up with good ways to handle this, and I was wondering if @kwalcock can alter the test framework to not check for the found Node exatly, but one which matches the desired info?  The Event is semantically identical, just not the exact same java object :(

kwalcock:
So, for this one the problem is that two moveThrough nodes look almost the same:

```
[info] migration-az-hack-9
...
[info] - should have correct migration1 node *** FAILED ***
...
[info]   4: Pagak Location(Pagak)
...
[info]   14: Pagak Location(Pagak)")
```
The move1 test passes because the first one gets found, but the migration1 event from eidos uses the second one.  The first one is labeled Concept/Entity and the second one is a Location.  They compare as false, so they can't be considered duplicates.  In general, we've just been lazy and added a filter so that the first or the last node gets picked, but the filter could be based on, for instance, the label.  Changing code to this fixes the test, but perhaps it shouldn't pass because we shouldn't be getting the near duplicate:

```scala
    val moveThrough1 = NodeSpec("Pagak", GeoLoc("Pagak"), NodeSpec.lastFilter)
```
I'll add this change in a PR and then someone can check whether it is a problem.

BeckySharp:
i mean -- semantically they are the same -- the label (at least here) is
irrelevant.  What matters is that they have the same text and the same
attachments

BeckySharp:
woohoo! 🎉

kwalcock:
OK then.  Becky gave it one reindeer.  I'll press the button.  Thanks for the good work.  It will be easier next time.

EgoLaparra:
Thank you guys! 👏

PR #806, opened 2020-04-02
Port unicode handling from ScienceUtils

kwalcock:
This includes manipulating RawTokens to account for substitutions.

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/EidosProcessor.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosProcessor.scala
@@ -155,7 +155,7 @@ class EidosTokenizer(tokenizer: Tokenizer, cutoff: Int) extends Tokenizer(
         else {
           val text = Normalizer.normalize(string, form)
 
-          assert(text.size >= 1)
+          assert(text.nonEmpty)
           text.map { char => (char, (start, end)) }
         }
       }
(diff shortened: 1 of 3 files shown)


kwalcock:
This is all @zupon's fault ;-)  There were some unicode characters in gold_groundings_annotated.tsv and gold_groundings.tsv that were causing complaints with unit tests that check for unicode and complain about it.  (At least one of these files came from an external source, so there's not much we could do.)  These were legitimate complaints, I think, in part because the unicode in the files was messed up so that where characters should have been the same, they differed.  The more important problem was that Eidos didn't ignore the characters and incorporated them into its analysis, resulting in an incorrect reading.  To fix it, the ScienceUtils from processors was used, in modified form.  Here are before and after versions of webapp output:

ChineseOld

ChineseNew

kwalcock:
But there's more...  One of the unit tests ran fine under Windows with IntelliJ, but gave an error with sbt.  It turns out the processor's reading of the unicode and accent files does not specify the encoding of the files even though they are known to be UTF-8.  Java just uses the operating system's default setting or what the user happened to have last configured.  For the Window's command line that happens to be "windows-1252".  The test can be passed using an environment variable to override it: `$env:_JAVA_OPTIONS="-Dfile.encoding=UTF-8"`.  This should not be necessary.

PR #808, opened 2020-04-06
Make FoundBy a thing

kwalcock:


kwalcock:
I didn't change any functionality, but did make this practice of concatenating and splitting on ++ into methods in a class where I won't forget how it's done.

PR #810, opened 2020-04-07
Use filter to pick correct node

kwalcock:


kwalcock:
Was someone supposed to remind me to produce better diagnostic output, because figuring this out is a pain!  Sorry about that.  This last one was more difficult because there are 8 candidate event mentions.  It really looks as if all but one have group arguments that refer to mentions which are not in the list of mentions that is returned from eidos.  They obviously exist somewhere in memory, but it seems a lot like we've lost track of them.  They don't seem to be among the answered collection of mentions.  Could this be?  Does it ring a bell?  I'm thinking of adding a unit test to check for the phenomenon.

kwalcock:
Each except the last two changes are the result of merging master with my branch.  I'm going to redo that so that it doesn't look so scary.

PR #811, opened 2020-04-07
Use filter to pick correct node

kwalcock:
Add another filter

### DIFF
--- a/src/test/scala/org/clulab/wm/eidos/graph/GraphTester.scala
+++ b/src/test/scala/org/clulab/wm/eidos/graph/GraphTester.scala
@@ -7,6 +7,7 @@ import org.clulab.wm.eidos.graph.TestResult.TestResults
 import org.clulab.wm.eidos.mentions.EidosMention
 import org.clulab.wm.eidos.test.TestUtils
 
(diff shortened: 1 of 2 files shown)


kwalcock:
If it only passes...

One way around the problem with the edge needing specific nodes which might not be the nodes previously found, is to find the edge first.  I believe it will find its own nodes and those nodes should match.  The drawback is that if it doesn't match, you don't know which of the nodes or edges was wrong.

kwalcock:
This still applies:

Was someone supposed to remind me to produce better diagnostic output, because figuring this out is a pain! Sorry about that. This last one was more difficult because there are 8 candidate event mentions. It really looks as if all but one have group arguments that refer to mentions which are not in the list of mentions that is returned from eidos. They obviously exist somewhere in memory, but it seems a lot like we've lost track of them. They don't seem to be among the answered collection of mentions. Could this be? Does it ring a bell? I'm thinking of adding a unit test to check for the phenomenon.

kwalcock:
I think I'm just again confused about how the odin mentions work and that there isn't a recent problem that should hold back this PR.  I'm writing tests elsewhere to make sure everything has a valid value.

PR #812, opened 2020-04-08
Egoitz - SUTime as optional TimeNormFinder

EgoLaparra:
This PR allows to use SUTime in `TimeNormFinder` for a faster parsing:
- A new option has been added in  `eidos.conf` to switch off the neural parser.
- An additional `annotate` method with a new signature has been added in`EidosSystem` to allow to set the DCT of `Document`.
- `TimeNormFinder` is now a trait with two classes extending it:  `TimeNormFinderNeural` and `TimeNormFinderSUTime`.


kwalcock:
I'm finally looking at this now, in case anyone is wondering (and in case there are outstanding changes in a local branch).

kwalcock:
So I need to time this, but we probably need to know how well this works.  Does @EgoLaparra have some kind of measurements so that will help us know when to trade working well for working fast?

EgoLaparra:
I cannot think in anything besides a manual comparison between the outputs of sutime and the neural parser. I can do it if you send me a few documents that have been processed with both options.

MihaiSurdeanu:
Similar to grounding, I strongly recommend building a small evaluation dataset for TimeNorm (say 50 temporal expressions) with gold data, so we can track changes quantitatively.

kwalcock:
I had to eyeball the output anyway to make sure it wasn't broken.  Here is a sample of the differences.   Some are systematic.

neural | SUTime | notes
-- | -- | --
last few months | the last few months
[none] | the end of June
past month | the past month
coming weeks | the coming weeks
[none] | Fall | "impact of Fall Armyworm"
next 1-2 weeks | weeks may
past few weeks | the past few weeks
previous month | the previous month
[none] | the start of July
past month | the past month
coming weeks | the coming weeks
June 2017 | late June 2017.Overall | matches original puncutation
recent | [none]
the same period | [none]
corresponding period two years earlier | [none]
coming year | the coming year
since July | July
since 2014 | 2014
period | [none]

EgoLaparra:
@kwalcock, can you send me the document that produces `late June 2017.Overall`?

kwalcock:
Sometimes I see for the neural
```
      "timexes" : [ {
        "@type" : "TimeExpression",
        "@id" : "_:TimeExpression_5",
        "startOffset" : 31724,
        "endOffset" : 31739,
        "text" : "past five years",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_1",
          "start" : "2012-05-11T00:00",
          "end" : "2017-05-11T00:00"
        } ]
      } ],
```
and for the new
```
      "timexes" : [ {
        "@type" : "TimeExpression",
        "@id" : "_:TimeExpression_1",
        "startOffset" : 31720,
        "endOffset" : 31739,
        "text" : "the past five years"
      } ],
```
Is the "the" messing up some regular expression so that the intervals can't be calculated?

MihaiSurdeanu:
SUTime seems to include "the" in the span. Either way is fine. I don't think that is a major issue.
The bigger issue is that our converter does not convert this expression into a TimeInterval. Playing with shell (without a DCT), I see that "the past five years" is normalized to "P5Y", which should be normalized  to the same time interval produced by the neural model. @EgoLaparra, I think this is for you. Thanks!

kwalcock:
Other examples do work with "the", but this doesn't again: "the past two decades".

kwalcock:
Here are some strange ones.  Something might be off by one.

Neural:
```
        "text" : "Monday",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_1",
          "start" : "2012-07-16T00:00",
          "end" : "2012-07-17T00:00"
```
SUTime:
```
        "text" : "Monday",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_1",
          "start" : "2012-07-23T00:00",
          "end" : "2012-07-24T00:00"
```

Neural:
```
        "text" : "Aug. 2",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_3",
          "start" : "2011-08-02T00:00",
          "end" : "2011-08-03T00:00"
```
SUTime:
```
        "text" : "Aug. 2",
        "intervals" : [ {
          "@type" : "TimeInterval",
          "@id" : "_:TimeInterval_2",
          "start" : "2012-08-02T00:00",
          "end" : "2012-08-03T00:00"
```
It's working great and seems fast.  It's probably easy to run with that handy switch set to true and false and compare the outputs.  SUTime seems better in some cases.  Thanks!

[00a2f02f9dd2da4d2bc1e5ab555b629a - Copy.txt](https://github.com/clulab/eidos/files/4452904/00a2f02f9dd2da4d2bc1e5ab555b629a.-.Copy.txt)

MihaiSurdeanu:
Ok, this is generally good news. Thank you both!
@EgoLaparra: can you please look at the differences reported by Keith? Are you passing the right DCT to SUTime?

EgoLaparra:
@kwalcock, what are the formats of the DCTs? For SUTime, it should be YYYY-MM-dd

kwalcock:
I'm using whatever dct.text returns, which looks OK.  I did move your method
` def annotate(text: String, documentTime: String): Document = {` elsewhere because I'm trying to avoid complicating that interface with all sorts of metadata variations.  I'll do a PR for you to see.  The file I sent uses so-called `CdrText` and can most easily be read with `ExtractCdrMetaFromDirectory.`

EgoLaparra:
I'm fine with your changes and everything seems to work fine.

Can you send me the meta-data for the file you have used, please?

kwalcock:
@EgoLaparra, it wasn't clear what the previously attached file was about after my copying and renaming to make it upload.  It contains the metadata (and the text).  The important parts are
```
		"CreationDate": "2012-07-23",
		"ModDate": "2012-07-23",
	"document_id": "00a2f02f9dd2da4d2bc1e5ab555b629a",
```
This is "CDR" format, so `ExtractCdrMetaFromDirectory` should be able to deal with it.

EgoLaparra:
😅  I saw the txt extension and assumed it only contained the text, sorry.

The `the past five years`  case comes from a different document, right? Can you send me it too?

kwalcock:
Here it is.  Again it's really a json file.  It was so useful to be able to switch back and for between true/false.

[000d4a4905d26bd5e116ca9d765673b3 - Copy.txt](https://github.com/clulab/eidos/files/4453930/000d4a4905d26bd5e116ca9d765673b3.-.Copy.txt)

kwalcock:
It's very effective on the speed issue!

Component | Neural Time (ms) | SUTime Time (ms)
-- | -- | --
Distinct | 3 | 0
AdjectiveGrounder | 10 | 1
HedgingHandler | 17 | 15
MigrationHandler | 17 | 25
NegationHandler | 32 | 16
StopwordManager | 49 | 85
ConceptExpander | 53 | 46
NestedArgumentExpander | 60 | 38
MostCompleteEventsKeeper | 136 | 90
GazetteerEntityFinder | 307 | 395
OntologyHandler | 577 | 242
RuleBasedEntityFinder | 634 | 721
GeoNormFinder | 4518 | 5059
TimeNormFinder | 711658 | 161

MihaiSurdeanu:
Nice!

EgoLaparra:
I've checked the strange cases and they are not that strange.

- `the past five years` is labeled by SUTime as a duration (`P5Y` means `Period of 5 Years`) not a date, so it cannot be align to a calendar interval. For example, the same `P5Y` would be used to label `the next five years`. This is one of the limitations of the TimeML schema, it only can align time expressions to single calendar units, e.g. `the past year` would be properly normalized. 

- `Aug. 2` is interpreted as the Last `Aug. 2` by the neural parser and as the Next `Aug. 2` by SUTime. In this case SUTime is right: `The U.N. Security Council has ordered the two sides to reach agreement by Aug. 2`

- `Monday` is interpreted as the Last `Monday` by the neural parser and as This `Monday` (i.e. today) by SUTime. Again, SUTime is right, however this case is a bit different to the previous one. The sentence is `South Sudan on Monday offered to increase the transit fees` that seems to talk about a past event, but that `Monday` is actually the same day of the document creation time. This is very common in news articles and has been a TODO in the scala version of the neural parser for a while. Maybe, it's time to revisit it.

MihaiSurdeanu:
@EgoLaparra: on the last two bullets, I am not sure we need to do anything. 
On the first thing, however, I think we can quickly implement some heuristics to handle them. That is, if the phrase contains keywords that hint of the future, e.g., "next", then consider it the next 5 years. Otherwise, consider it the past 5. What do you think?

EgoLaparra:
We also need keywords for the past cases. Durations are usually not normalizable,  like in `it takes 5 years to get a PHD`, that would also be labelled as `P5Y`.  But I can try to solve those explicit cases like `the next 5 years` or `the last 5 years`.

MihaiSurdeanu:
Agreed. But let's start somewhere simple: maybe two lists with obvious keywords to indicate past/future?

kwalcock:
The default configuration setting is to keep using the neural version.  So, there's little harm in checking this in no matter what the accuracy is.

MihaiSurdeanu:
Ok!


On April 13, 2020 at 10:01:13 AM, Keith Alcock (notifications@github.com)
wrote:

The default configuration setting is to keep using the neural version. So,
there's little harm in checking this in no matter what the accuracy is.

—
You are receiving this because you commented.
<https://github.com/clulab/eidos/pull/812#issuecomment-612989085>, or
unsubscribe

PR #813, opened 2020-04-08
Use geonames as a resource

kwalcock:


kwalcock:
I'm just hesitating on this because we'll lose the wordas part.  See https://github.com/clulab/geonorm/issues/1.  Does @BeckySharp think that's a problem?

BeckySharp:
for the Ethiopia use case, it's a problem.
For other use cases, no.
how important is this?

kwalcock:
IDK.  If we need to do Ethiopia again, I don't want to resort to old code.  To get Ethiopia, I probably have to encourage Steve to add it back.

BeckySharp:
Yeah -- I wonder if there's a solution where a general resource is
available as a resource, and in there are use-case specific components,
those can be "added in" (and they'd almost certainly be much smaller).....

kwalcock:
The new geonames in combination with the new processors results in two failing migration tests.  I will look into it.

kwalcock:
This works so very much better than trying to download the files at runtime!

MihaiSurdeanu:
+1!


On May 20, 2020 at 9:23:52 AM, Keith Alcock (notifications@github.com)
wrote:

This works so very much better than trying to download the files at runtime!

—
You are receiving this because you are subscribed to this thread.
<https://github.com/clulab/eidos/pull/813#issuecomment-631581311>, or
unsubscribe

kwalcock:
However, I wonder about the geoIDs changing.  Is this expected?  In the table below, "before" comes from
```
    "org.clulab"                 %% "geonorm"                  % "0.9.8",
  geoNamesIndexURL = "http://clulab.cs.arizona.edu/models/geonames+woredas.zip"
```
and "after" comes from
```
    "org.clulab"                 %% "geonorm"                  % "1.0.0",
    "org.clulab"                  % "geonames"                 % "1.0.0+20200518T005330Z.gadmworedas",
```
Mostly I want to make sure that I haven't make a mistake in the matter.  There are quite a few more examples that can be collected, if necessary.  Thanks, @bethard, for checking if you can.


Text | Before | After
-- | -- | --
Leer County | 4161743 | 4433190
Uror | 10670095 | 11371354
Rabdit | 10994896 | 781854
Aiyella | 264350 | 98227
long-maturing | 8812290 | 2101495
States | 734014 | 12353
Pagak | 7405950 | 6370601
Dinka Bor | 11869855 | 7574759
outcomes | 2307970 | 3563719

bethard:
Yes, any changes to the index will cause ties to be broken in different ways. I spot-checked a few of these and these all look like reasonable confusions. For example, for "Dinka Bor", it changed from https://www.geonames.org/11869855/borodinka.html to https://www.geonames.org/7574759/pristan-borodinka.html which both have "Borodinka" as one of their alternate names. They're both probably wrong, since I assume the correct answer in your data is https://www.geonames.org/377231/bor-dinka.html, but it doesn't look like the model is getting any worse. Name normalization is hard. =)

kwalcock:
Thanks for looking, and not for the first time.  The answer might have changed since before, so a double check seemed prudent.

PR #816, opened 2020-04-08
Fix FoundBy invention

kwalcock:


BeckySharp:
good catch!

kwalcock:
For the record, world, it was my own mistake.

PR #818, opened 2020-04-09
removing old app

BeckySharp:
And perhaps @kwalcock  this is as good a place as any to start a conversation about cleaning out the app closet... since we're all stuck at home with nothing better to do ;)
I know i'm guilty of contributing to the app proliferation, and i'm about to add one more

Maybe some can be refactored to be diff configurations of a single app?


kwalcock:
OK.  Let's try an experiment.  Let me know when you regret having deleted the file ;-)  Just now I can still find better things to do.  Another option is to collect them somewhere and labelled legacy or unsupported, swept under a rug.

PR #820, opened 2020-04-10
synhead backoff for triggerOpt

BeckySharp:
I think this will meet your needs for getting a fake trigger from TextBoundMentions (i.e., the cause and effect of a causal mention)

fyi @danebell

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/utils/MentionUtils.scala
+++ b/src/main/scala/org/clulab/wm/eidos/utils/MentionUtils.scala
@@ -56,22 +56,28 @@ object MentionUtils {
     case _ => None
   }
 
-  def triggerInfo(em: EidosMention): TriggerInfo = triggerInfo(em.odinMention)
-  def triggerInfo(m: Mention): TriggerInfo = m match {
-    // if mention is an EventMention, it's easy
-    case em: EventMention => TriggerInfo(em.trigger.text, em.trigger.startOffset, em.trigger.endOffset)
-    // Otherwise, check to see if the mention has a syntactic head
-    case _ => {
-      val synHeadOpt = m.synHead
-      if (synHeadOpt.isDefined) {
-        // get the info from the syntactic head
-        val head = synHeadOpt.get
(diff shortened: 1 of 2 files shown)


BeckySharp:
and I even added a test!  @kwalcock , this is me trying to be more like you!

BeckySharp:
The method returns a `TriggerInfo` which (I think) contains the info you need for lum issue 73

kwalcock:
Thanks for the code and the test.  I'll be transferring this PR to the LUM version right away.  For the span object, the spec says

`"text": "unnormalized text referred to by the start/length" # Optional`

and in the code I see some use of "word", which I believe are to some extent normalized.  I'm not offhand certain of the text() method.  I think that some of these might need to be references to raw instead.  Could you add tests with things like apostrophes and commas and other things we manipulate and see if the original text is in the result?

BeckySharp:
> Thanks for the code and the test. I'll be transferring this PR to the LUM version right away. For the span object, the spec says
> 
> `"text": "unnormalized text referred to by the start/length" # Optional`
> 
> and in the code I see some use of "word", which I believe are to some extent normalized. I'm not offhand certain of the text() method. I think that some of these might need to be references to raw instead. Could you add tests with things like apostrophes and commas and other things we manipulate and see if the original text is in the result?

Sure, I can add more tests, good idea.  I put it here bc we may also need to evaluate on LDC text for WM, and so this may be useful.  I'll prob rename the method though, I woke up annoyed with myself over it :)

BeckySharp:
and I'll go back to the raw

kwalcock:
Thanks, @BeckySharp.  You make me feel lazy when you include javadoc comments!  My suggested punctuation marks turned out to be pretty bad.  I changed them from and apostrophe to quotation mark, which gets translated into a word of two characters, and to an n-dash, which gets replaced with a plain dash.  These might have tripped us up.  There's a PR in the works.

kwalcock:
This was updated in LUM, but it does not change output.

BeckySharp:
cool, thanks!

PR #822, opened 2020-04-13
Egoitz - SUTime with normalization of explicit Durations

EgoLaparra:
This is a follow-up of https://github.com/clulab/eidos/pull/812. 
This PR adds normalization of durations with the form 'the next|past N units'


kwalcock:
@EgoLaparra, is it OK with you if I take a couple things like
```
    val normalizableRegex = s"^the (${(pastOperators ++ nextOperators).mkString("|")})".r
```
and turn them into static values?  I'm especially concerned that this one might be repeatedly calculated and it looks rather expensive.

kwalcock:
This ZoneId.systemDefault() has entered the code recently and highlights the fact that I don't 100% understand how LocalDateTime works.  When we have a date, we generally do not know the location or the corresponding time zone.  My impression is that LocalDateTime is meant to deal appropriately with this vagueness.  If a time zone is somehow added here, don't we run the chance of something being recorded at midnight in London but some time a day earlier in Tucson?  Does it mean that our output can be different when run on different computers?  The documentation at https://docs.oracle.com/javase/10/docs/api/java/time/ZoneId.html#systemDefault() says "If the system default time-zone is changed, then the result of this method will also change."  That sounds like something to be avoided.  Can you clue me in?  It seems like a different strategy might be needed.  Thanks.

EgoLaparra:
Yes, you are totally right @kwalcock. I've been looking into SUTime and it seems to use UTC as default time zone, so I think we can use the same zone to avoid the ambiguity. I'm also going to include a better processing of the time part of the time expressions (hours, minute,...).

kwalcock:
Thanks, @EgoLaparra.

PR #824, opened 2020-04-14
Work around assembly merge problem

kwalcock:


kwalcock:
@bgyori, can you give this a try after I merge?  I'll try to remember to test the assembly each time we update a dependency.

PR #825, opened 2020-04-14
Attempt to add function without affecting processors

kwalcock:
Second attempt
Unused import
Add empty line


kwalcock:
@zhengzhongliang, your task is much more difficult than I realized and thanks for insisting on better answers.  If adding processors makes it too much more complicated, you are welcome to try this as a temporary solution.  It doesn't require any changes to processors, but can be ported there easily after you settle on how that code should look.

PR #826, opened 2020-04-15
Grounding zhengzhong readytomerge 20200407

zhengzhongliang:
Changes:
0, All changes are for the compositional grounding. 
1, Use the all mention tokens to ground the concept when the syntactic head does not contain "NN" POS tag. The weights of the tokens are determined by their POS tags, NN, NNS, NNP get 5 and others get 1 (values are tuned to the best performance). The composited vector is normalized.
2, Use competing grounding for the concept branch and the process branch.
3, The results are: concept MRR=0.5635, process MRR=0.675, property MRR=0.7875.
4, This uses the localPublish version of the processor of branch "localPublish_makeVector_20200415", I will also make a pull request of that branch of processor. The publish version is "8.0.2-SNAPSHOT". The "procVer" value in build.sbt of eidos should match that.


src/main/scala/org/clulab/wm/eidos/groundings/EidosWordToVec.scala
```diff
@@ -75,6 +79,28 @@ class RealWordToVec(val w2v: CompactWord2Vec, topKNodeGroundings: Int) extends E
...
+    // It could be that the composite vectore below has all zeros even though some values are defined.
+    // That wouldn't be OOV, but a real 0 value.  So, conclude OOV only if none is found (all are not found).
+    val outOfVocabulary = sanitizedNameParts.forall(w2v.isOutOfVocabulary(_))
+    val termWeights = posTags.map(tag=>if (tag.startsWith("NN")) nounVerbWeightRatio else 1.0.toFloat)
+
+
+    if (outOfVocabulary)
+      Seq.empty
+    else {
+      val nodeEmbedding = w2v.makeCompositeVectorWeighted(sanitizedNameParts, termWeights)
```
zupon, line 93: @zhengzhongliang I'm trying to run `EvalGroundings` to try out your changes, but I run into an error here with `makeCompositeVectorWeighted`. It doesn't seem to exist anywhere. I pulled your branch directly, too.

kwalcock: I'll be trying it out soon as well, but note step 4 above.  It requires a change to processors that has not yet been published, so doing that locally is required.

zupon: I pinged (pang?) Zhengzhong on Slack for help with step 4 and I got things running.

zupon:
I reran the evaluation and got the same results as Zhengzhong. I also did two quick experiments. 

First, Zhengzhong's branch is using an older copy of the gold groundings, which has only 80 entities evaluated. I tried the newest version, which has 130 entities evaluated and some of the gold groundings fixed. Here are the results:

```
130 entities evaluated
flat grounding accuracy:	0.72397435
concept grounding accuracy:	0.5878205
process grounding accuracy:	0.6615385
property grounding accuracy:	0.8769231
```

Second, in my earlier error analysis I found a few areas where the ontology needed some changes. These mostly helped Property, but also gave some small boosts to Concept and Process. Running Zhengzhong's branch with the updated ontology yields the following results:

```
130 entities evaluated
flat grounding accuracy:	0.72397435
concept grounding accuracy:	0.5955128
process grounding accuracy:	0.66923076
property grounding accuracy:	0.9076923
```

Comparing these results (even just with Zhengzhong's additions, not including the updated ontology) to my current branch, we see a ~4 point improvement in Concept and a whopping 16 point improvement in Process.

Other than whatever merge conflicts there were, these updates look good, and the results are definitely an improvement, especially for Process, which was our biggest problem before.

zhengzhongliang:
The conflicts are resolved.

MihaiSurdeanu:
I released processors 8.0.2, which should be available on maven in a couple of hours. As soon as it does, and the unit tests pass, we can merge this!

Good job @zhengzhongliang and @zupon !

MihaiSurdeanu:
processors 8.0.2 is now on maven so this should work now. Maybe we need another commit to this branch to force tests to run again? @kwalcock ?

kwalcock:
One thing that has to happen is that
```
  val      procVer = "8.0.2-SNAPSHOT"
```
needs to change.  Would @zhengzhongliang like to do the honors?  It should force another test.

MihaiSurdeanu:
The diff shows that this has been changed to 8.0.2 already, no?

kwalcock:
Yes, sorry.  I was looking at stale data.  I'll kick it then.

kwalcock:
It has all green lights now  I'm just going to double check before merging.

MihaiSurdeanu:
Yay!!

kwalcock:
Thanks, someone, for fixing my typo in allSimiliarities!

Here are a couple of questions/comments to make sure I understand this.

Is `m.text.split(" ")` here intentionally different from `m.words.toArray`?  m.text is actually a method that has to do substantial work to reconstitute a mention's text from the individual words and might even insert multiple spaces.  It looks like this is all immediately undone by splitting on the same spaces.  Are the words good enough?  I wish toArray was not necessary, but that's a different project.

Could the lines
```scala
      val mentionHeadTag = mentionHeadOpt.map(m=>m.tags.head.head).getOrElse("None")
      val conceptSimilarities = {
        if (!mentionHeadTag.startsWith("NN")) {
```
be combined into something like
```scala
      val mentionHeadTagIsNN = mentionHeadOpt.map(m=>m.tags.head.head.startsWith("NN")).getOrElse(false)
      val conceptSimilarities = {
        if (!mentionHeadTagIsNN) {
```
mostly so this temporary use of "None" can be avoided?

That SingleOntologyGrounding of mine isn't very transparent and should probably change.  It is hiding the fact that a Namer tries to keep track of the ontology branch.  So in
```scala
          .filter(_._1.name.startsWith("wm/"+branch))
```
if it's safe to assume that "wm" is already a given and that there are branches because compositional grounding is being used, then something closer to
```scala
            .filter(_._1.branch.get == branch)
```
would be preferable.

I just want to confirm that the `take` comes before `filter` because you want each branch to compete with the others for the top places.  (I should have paid more attention to your reports.)
```scala
          .take(effectiveTopN)
          .filter(_._1.name.startsWith("wm/"+branch))
```

For strings like "property", can constants like CompositionalGrounder.PROPERTY be used?

Thanks for putting up with the nitpicky critique.

kwalcock:
Oops, one more.  In calculateSimilaritiesWeighted, the `1.0.toFloat` can probably just be `1f` or `1.0f`.

kwalcock:
@zhengzhongliang, do you have time to look at the small issues above?  I can make changes if need be.

zhengzhongliang:
Hi Keith,
Thanks for the review, here are the responses:
1, m.words.toArray is indeed better than m.text.plit(" "), I changed to m.words.toArray.
2, I changed the way to get the tag of the syntactic head to the method you suggested.
3, The filter branch method is changed to what you suggested .
4, Yes the current order is correct. The filter should be applied after take because I want the branches to compete with each other. 
5, strings like "property" are replaced with what you suggested.
6, 1.0.toFloat is changed to 1.0f

I made these changes and tested the new code, and the result does not change, so I think this is ready to merge.

zhengzhongliang:
Please let me know if anything else I need to do to merge this. Thanks.

kwalcock:
This is excellent and oh so close.  Can @zhengzhongliang  and @zupon look at the merge conflict with evaluation_results.tsv and decide what needs to happen?  Perhaps the file even needs to be regenerated to match whatever master would output.  I haven't followed it closely enough to know.  Thanks!

zhengzhongliang:
I left a message to Andrew on slack, should be able to fix this soon.

zhengzhongliang:
Conflict is resolved. @kwalcock

kwalcock:
Thanks for everyone's hard work on this.

PR #828, opened 2020-04-18
Add special characters to stop list

bgyori:
This partially addresses https://github.com/clulab/eidos/issues/827 by adding % and ~ to the stop list since they are frequently picked up as entities.


BeckySharp:
may not be needed with #830 , FWIW

kwalcock:
Let's try #830 and see how it works.  It will apply to any 1-character value, not just the two added here, and there are additional checks performed.  Please sound the alarm it you notice unforeseen consequences, @bgyori.

PR #829, opened 2020-04-18
Grounding app

BeckySharp:
One more app bc apparently 137 aren't enough... :)
This one generates the grounding annotation sheets we've been using lately


kwalcock:
Thanks for dealing with the bad parses.  Next time I read a paper of yours, I'm going to look for periods that dangle three lines down from where they should be



.
😉

PR #831, opened 2020-04-20
added most up to date grounding eval tsv files

zupon:
Adding the most up to date tsv files for evaluating the compositional grounder. The main changes are (1) more entities being annotated and (2) some fixes to the gold groundings where there were errors (e.g. removing an extra non-leaf node that no longer exists in the current ontology).

NB:  the results shown here use the version of grounding that do not include Zhengzhong's updates to the grounding algorithm, which have not been merged into `master` as of this PR.

Before:
```
80 entities evaluated
flat grounding accuracy:	0.70979166
concept grounding accuracy:	0.5197916
process grounding accuracy:	0.45
property grounding accuracy:	0.7875
```
After:
```
130 entities evaluated
flat grounding accuracy:	0.72397435
concept grounding accuracy:	0.5470513
process grounding accuracy:	0.50384617
property grounding accuracy:	0.8769231
```

@kwalcock Assuming no unicode or other issues, are these two files okay to merge?


kwalcock:
Not Unicode, but close:
```
[info] - should not have any CrLf line endings in /var/jenkins_home/workspace/idos_zupon-updated-grounding-tsv/src/main/resources/org/clulab/wm/eidos/english/grounding/gold_groundings_annotated.tsv *** FAILED ***
[info]   true was not false (TestCrLf.scala:38)
```

zupon:
Just to double check, what should the line endings be? `CR` or `LF`? In principle there's a way to set it in IntelliJ to warn about these sorts of things when you commit a file, but I haven't tracked it down yet. There *is* a way to change the line endings in individual files though.

kwalcock:
If you don't have a handy way to do it, then I can.  When files get to github, they should be LF.  This file has CR/LF just now as if it had been in a Windows machine and not converted when uploaded, perhaps from a non-Windows machine.  It's a slight pain now but more pain later.

zupon:
I'm seeing `LF` on both files on my end, yet it's still failing a check.

kwalcock:
The gold_groundings_annotated.tsv in that branch is still CR/LF.  I think it has to do with the order things are being merged.  It looks like the two changes are competing and this PR will clobber the LF change you want to have made.  The sure way to check is to download the zip file of the branch and peek.  If it remains a mystery, I can stab at it.

image

zupon:
@kwalcock I've expended my knowledge/effort. Even going back (again) and making sure I had them as `LF` (again) didn't seem to fix it. Since you know how to make it work, I'll leave this in your hands.

kwalcock:
The horrible, no good, very bad day...  It looks a lot like the files are scrambled now, as if there had been a merge conflict.  Parts are repeated.  If there was ever an export out of Excel, then it may have inserted the CR/LF, it being a fairly Windows-centric program.  Third try will be a charm.

kwalcock:
Messages crossed.  OK.  It will probably be the version from "changed eol endings to LF".

zupon:
Yes, that's probably the one. I reverted to previous commits because with just changing the line endings on IntelliJ, git didn't think that there were any changes to the file, so I couldn't re-add and commit. Quite the mess. If/when I need to do this again I'll come to you before I start trying things out on my own to learn the real/good way of doing things.

zupon:
Thanks, @kwalcock !

PR #833, opened 2020-04-20
Try to convert to LF.

kwalcock:


kwalcock:
It took me two tries.  For the record, @zupon, this is being merged.

PR #836, opened 2020-04-21
Backport PMI changes from LUM

kwalcock:


kwalcock:
In the LUM version I removed the duplicate code that poor man's indra more or less provides.  I needed to rearrange a few things to make use of the functionality, though.  Backporting will hopefully make it easier to keep the versions in sync, especially if the functionality moves somewhere else.  The code is also slightly more efficient, just doing a count rather than a collect and size.

PR #843, opened 2020-04-24
Consolidate tag code

kwalcock:


kwalcock:
This consolidates code that compares tags against strings into the TagSet class.  Generally things like Finders, Actions, Expanders, and the like do not know which language is configured or even which Processor is being used.  Changes there can invalidate all the string comparisons.  It was also being difficult to keep strings from different tag sets from interfering with each other because values in one can be prefixes of values in another.  Comparisons for different language are now independent.

It is somewhat of a hassle to convey the TagSet to different parts of the program, but it wasn't too bad.  The TagSet class can also become a mishmash of unrelated is-functions, but it isn't yet a problem.  We should strive to keep them as generic as possible, sometimes combining existing methods.

kwalcock:
FYI @gcgbarbosa

kwalcock:
It would be nice for it or part of it to be closer to where the tags are actually issued, I agree.  That's where people will look and that's where the code might need to be maintained when something changes.  There are a couple of reasons it didn't make it that far.  For one I am just guessing at the tags and this list is not even complete or definitive.  The punctuation is certainly missing, for example.  There could be some simple methods in the processor's version like isVerb and isNoun to summarize tags that are subdivided.  It gets complicated quite quickly, though.  isConjunction would work for Portuguese: CCONJ and SCONJ.  For English, it starts with CC, but apparently only part of IN, which is a preposition or subordinating conjunction, would make sense.  There are several methods that are particular to Eidos, like isValidFinal, isStopwordContent, and isOntologyContent which would need to stay in Eidos anyway.  Changing both processors and Eidos to include these one-line methods seemed to be a little extreme.  I was going to leave them in Eidos initially and see whether they prove to be helpful or not.

MihaiSurdeanu:
Ok, thanks! I think the multi-linguality is addressable.
Maybe the basic functionality should be in processors? And the Eidos
version can inherit from this to add Eidos-specific functionality?
Anyway, this is not urgent at all.

kwalcock:
It not being urgent, I'm saving further refinement for another rainy day.  It seems OK to merge this for local use in the meantime.  If people need to consult the tags for something like grounding, there will be a more standardized or formalized way of doing so.

PR #844, opened 2020-04-29
Add wmuser project

kwalcock:
This consumes kafka messages regarding documents that need to be read and then consults the rest service to download them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,562 +1,76 @@
-[![Build Status](http://jenkins.cs.arizona.edu:8090/buildStatus/icon?job=eidos%2Fmaster)](http://jenkins.cs.arizona.edu:8090/job/eidos)
-
-# Eidos
-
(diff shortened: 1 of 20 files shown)


kwalcock:
This is the alpha version and is here in case anyone wants to track progress while I tidy up some things and write a README file, for instance.  wmuser is a subproject of eidos, similar to elasticsearch so that library dependencies needed only for wmuser do not bloat eidos.  wmuser is also independent of eidos itself in case it needs to be moved later and so that it remains very small itself.  The downside is that I did need to copy and paste a bunch of utils to the project to achieve the self-containment.  There will likely be a wmproducer namespace with at least a RestProducerApp added to wmuser in the relatively near future so that the output of eidos can be uploaded somewhere afterwards.

MihaiSurdeanu:
Thanks Keith! Very cool.

PR #845, opened 2020-04-29
Turn off neural timenorm by default

kwalcock:


kwalcock:
This will be followed up with some timing tests on the requested 1000 or so documents.  The timing code is in a different branch, though, so it will be addressed separately.

kwalcock:
Like I said, after I track down the set of changes for the tests.  It very probably has to do with the ordering of the mentions that are produced.  Somewhere I fixed them, but maybe there are more in the meantime.

PR #846, opened 2020-04-30
Egoitz - Scorer for time normalization

EgoLaparra:
This pull request includes an evaluation app for time normalization (`EvalTimeNorm`), an unit test (`TestEvalTimeNorm`) that runs the evaluation and the resources required (docs and test set) .

The unit set is set to `ignore` by default. 

`useNeuralParser` is now set to false, so unit tests have been adapted.


kwalcock:
FYI I will add an @unchecked somewhere to suppress this just noticed warning.

```Warning:(70, 74) match may not be exhaustive.
It would fail on the following input: (_, _)
      val predictTimex = mentions.flatMap(_.attachments).zipWithIndex.map{
```

kwalcock:
Changing from map to collect seemed to work better.

EgoLaparra:
Thanks @kwalcock

PR #847, opened 2020-04-30
Seek negative LOC

kwalcock:
Some things needed to be cleaned up


kwalcock:
People working closely with the grounding may have noticed that there were two majorly overlapping interfaces in OntologyHandler.scala: process and ground.  That's a historical remnant and the process one is finally removed here.  I'm just going to tag a few people working in that area in case this change messes up their plans or looks like a mistake: @zhengzhongliang, @zupons, @BeckySharp.  Thanks.

PR #850, opened 2020-05-07
Add glove vector culling app

kwalcock:


MihaiSurdeanu:
This really needs to be in the embeddings package in processors, not in Eidos...

kwalcock:
The case of the words in the vector file is not changed (or at least that isn't the intention).  In order to get the frequencies, the glove words have to be lowercased when looking them up in gigaword.  That's all.  Except for the altered first line and the added second line, all other lines are transferred directly from one file to the other.

MihaiSurdeanu:
You're right. Sorry about the false alarm!

kwalcock:
Parts of this are slated for migration to processors.  It may be some time before that happens, so I'm merging this into master for safekeeping.

Before the migration, I think that the first line of the file should be discussed, the one indicating number of rows and columns.  The second (or maybe it should be the last), the one with the OOV value, has some issues as well.

MihaiSurdeanu:
Can you please move this to the “srl5” branch of processors, in the
org.clulab.embeddings package?

On May 11, 2020 at 11:05:59 AM, Keith Alcock (notifications@github.com)
wrote:

Parts of this are slated for migration to processors. It may be some time
before that happens, so I'm merging this into master for safekeeping.

Before the migration, I think that the first line of the file should be
discussed, the one indicating number of rows and columns. The second (or
maybe it should be the last), the one with the OOV value, has some issues
as well.

—
You are receiving this because you commented.
<https://github.com/clulab/eidos/pull/850#issuecomment-626863521>, or
unsubscribe

kwalcock:
Done.  See https://github.com/clulab/processors/pull/392.

PR #852, opened 2020-05-11
Backport EidosSystem timing from LUM

kwalcock:


kwalcock:
This is step one of several.

kwalcock:
All steps complete.  This should not change any externally observable behavior except that the component loading now happens largely in parallel and it should therefore be faster.  This helps keep different repos in sync and might also influence the design of the processors server.  If useTimer is turned on, then timing information is collected for running and loading.  Most output is through the logging.  Summary information can also be included.

PR #854, opened 2020-05-12
Make this easier to compare to LUM version

kwalcock:


BeckySharp:
This reminds me -- we added the ability to Odin read yaml files with one thing per line, so I can reformat the triggers too to make them easier to compare -- this is me saying it out loud so maybe i actually get to it

PR #855, opened 2020-05-12
Calculate allEidosMentions just once

kwalcock:


kwalcock:
While reconciling code I noticed that in one place, for the adjective grounding, not all EidosMentions were being processed.  It was in a place where I have waffled about the interface and gone back and forth several times.  This ends that.  The Eidos refiners work against AnnotatedDocuments that have both the surface mentions and underlying ones.  This also removes the few times where those values were calculated more than once.  One error is fixed.

This would be a trivial refactoring with no change in output except there was a change.  It seems that for some reason in an edge case in one document, one mention is not getting processed.  I haven't been able to figure out the cause.  It is patched by calculating it the second way that doesn't miss the mention.  This preservs behavior, but does make it obvious that there is a problem.  I'll have to save that for later and in the meantime get the first error fixed.

kwalcock:
The patch was just covering up an underlying problem, so I'm rejecting this PR.  See instead kwalcock-verifyMentions.

PR #856, opened 2020-05-14
Use resourcer

kwalcock:


kwalcock:
This is just a proof of concept and check to make sure it works on Jenkins.  All (I should double check) of the resources that Eidos loads are being loaded from files if they are available, such as from within the development environment.  I am able to run the EidosShell from within IntelliJ, edit a yaml file to update a rule, have the shell reload, and then observe the rule change the next time text is processed.  It takes about 10 seconds.

This little change is really there just because IntelliJ must not be arranging for access to these resources directly by manipulating the classpath.  Here's a little discussion of that possibility on stack overflow: https://stackoverflow.com/questions/1464291/how-to-really-read-text-file-from-classpath-in-java.  Perhaps there is a setting for that somewhere.  It wouldn't work with sbt, though.  Perhaps it has its own setting.  I hope that at one time this was not necessary with eclipse back when I asserted that something like this was working.

The output from EidosShell is not the preferred output, so I'll check on getting something better from the webapp.

PR #858, opened 2020-05-14
Make more robust RestProducer

kwalcock:
This is a candidate for the "real" program to upload results.  The curl version was more to test the receiving end and was more appropriate for ~100 files.  This one, if it works at all, is better for the 10,000s of files that we are expecting.  This one will be used in the second round of testing which will commence shortly.


kwalcock:
Two Six has sent out another set of documents to read and this program will be used to upload the results.

PR #860, opened 2020-05-15
Generate html from Eidos and use in EidosShell

kwalcock:


kwalcock:
This is a poor person's IntelliJ/sbt/eclipse extension that helps people work on rules and see the results from within the development environment with relatively quick turnaround time.  In about 15 seconds a new rule can be applied and the result viewed either as text or as html in the browser.

To use, start up the EidosShell.  It can run within an IntelliJ debug session in a Console tab right next to the debugger tab.  Use the program pretty much as normal.  Just before the text output appears, the web page eidosshell.html will be updated with what would (almost) otherwise be the webapp output.  Open a web browser on it.  If you enter other sentences, just refresh the browser.

If you change a rule, like in entities.yml, have EidosShell do a :reload.  It takes under 10 seconds for me.  (Some of the reloading has be be double checked.  It is probably outdated.)  Type the sentence again (or better, paste it, or use the history), and the new rule should be used.  Just update the browser display if you want the HTML.  (I haven't yet dealt with the case for the rule change causing the program to crash, so be nice.)  Set breakpoints in the program to see what might be going wrong internally.

kwalcock:
This is displayed from a local file.  No web server is required, no webapp/run, no processors server, no sbt.

image

BeckySharp:
This is pretty cool!

kwalcock:
Can someone with a Mac try out the webapp in as many browsers as is convenient?  It's working in Windows Chrome, Firefox, Internet Explorer, and Microsoft Edge without any error messages in the console, especially not about missing fonts, exceptions, timeouts, etc.  The display should not have changed otherwise.

kwalcock:
This  was cancelled in favor of a different branch to avoid the cache problem.

PR #862, opened 2020-05-18
Main as finder

BeckySharp:
FYI @MihaiSurdeanu @zupon 

This is a PR that converts the main causal extraction to a finder, so we can enable/disable it (though if we disable IDK what we're doing...).  More importantly, it simplifies some code and allows for more modularity in the order of extraction if nothing else.

Since this is a pretty "major" change (less in terms of code than in terms of the config), I though we could go ahead and remove the MitreExporter that's been deprecated for a while... eh eh?

@Keith -- on the `weakIdentity` method -- I left a longish note in the commit message, but if you want that to be removed just let me know, it's not being used anywhere but I have a line commented out suggesting where it could be used to _maybe_ improve/speed up things


kwalcock:
As Becky points out, correctness comes before speed, so I think it's fine to leave that method.  I'm surprised that all the tests are passing.  I thought that this needed some careful experimentation before use, or was that just after it is used for something in particular, for the actors?  If that's the case, I will merge this branch first into lum's master before into this one.

kwalcock:
Sorry for how complicated that EidosComponents is.  There are some intricate dependencies there. 
 I'll backport after it is working.

PR #863, opened 2020-05-19
Use Expect header

kwalcock:


kwalcock:
When we return results to the Two Six server, the connection was being closed:
```
INFO: I/O exception (java.net.SocketException) caught when processing request to {s}->https://wm-ingest-pipeline-rest-1.prod.dart.worldmodelers.com:443: Connection reset by peer: socket write error
May 18, 2020 12:07:25 PM org.apache.http.impl.execchain.RetryExec execute
```
It turns out that this happens only for larger files, of size somewhere between 190KB which worked, and 380KB which didn't.  At some point the server won't take a files unless the client has stated that it will accept this continue header thing.  The server uses that to handshake with the client regarding whether it will accept the relatively large file.  The client can then go ahead and upload.  This behavior isn't part of the default setting for the Apacha HttpClient and needed to be added.  Thank you to the curl code which showed this header information, I think with the -D flag:
```
HTTP/1.1 100 Continue

HTTP/1.1 201 Created
Content-Length: 61
Content-Type: application/json;charset=utf-8
Date: Mon, 18 May 2020 19:57:20 GMT
Server: Jetty(9.4.z-SNAPSHOT)
```

PR #865, opened 2020-05-20
Add program to run a simple string

kwalcock:


kwalcock:
This was lying around elsewhere and I needed it to make sure the main (not the test) configuration was working well without running any of the other existing programs.  It doesn't need any arguments, doesn't write anything to disk, can be started by right clicking, etc.

PR #866, opened 2020-05-20
make attachmenthandler and corefhandler into refiners

BeckySharp:
Included:
 - Removing attachment merging and coref from the global action (and EidosActions, no longer needed a CorefHandler) and moving them to the sequence of OdinRefiners
- created AttachmentHandler
- removed some unnecessary code
- updated tests
- removed a few things from `eidos.conf` that are very infrequently (if ever) used/modified, they still live in `reference.conf`, so we can override them if needed

more modularity!


kwalcock:
Thanks especially for the comments around the refiners and for the test files.  I'm really wondering if it would be better to go ahead and merge both this and the previous one.  I can always do it tomorrow.  Remind me of that in a week.  In the meantime, I'll add it to LUM and see what happens.

kwalcock:
I think that changes in the geonames cache were causing problems, so a branch was merged instead.

PR #874, opened 2020-06-08
Remove migration

kwalcock:
Edit changes


kwalcock:
Agreed.  I don't like losing code, but there's a record of it here.  It looks like I may have missed some methods which are now never called, but then we probably don't lack for them elsewhere.  I don't see that IntelliJ has a way to search for them, so I won't worry too much.

kwalcock:
Goodbye, migration.

PR #881, opened 2020-06-17
Clean up EidosMention

kwalcock:


BeckySharp:
👍

kwalcock:
Although there is automatic deduplication of top-level, surface Odin Mentions, there is significant duplication of the "underlying" Mentions in triggers and arguments hiding under the surface which  EidosMention is left to contend with.  Those have been sorted out haphazardly in the past, with ties often being broken based on which order the graph of Mentions is traversed, and decisions about multiple contenders being spread out across code execution.  This PR addresses the situation, concentrating all deduplication information into two local variables of EidosMention.asEidosMentions and calculating their values in two separate recursive passes which do not need to be coordinated.  This prevents different (by identity) Mentions from being displayed when only one (by equality) is meant to have deduplicated them.

kwalcock:
A bunch of distracting, potentially reusable utility/library classes have been moved out of the way, variables and methods have been reordered and marked protected, Time and Geo are no longer dependent on EidosMention, some tests for equality are avoided when identity will suffice, some names have been updated for clarity, hopefully.  Behavior shouldn't change except in ambiguous situations, examples forthcoming.

kwalcock:
Since == for Mentions does not include the foundBy field, and probably rightly so, the usual way to observe problems with deduplication is to spot this value changing in a regression test.  Some examples are below.  It looks like the new algorithm is favoring the Mention with the longer list of rules.  If this isn't the desired behavior, there is a good place to change it.  A report of rule-changing deduplications might be output short of having to compare against previous runs.  Some values look strange and might be diagnostic.  Migration was not yet removed for this run.


Old foundBy | New foundBy
-- | --
simple-np | simple-np++GeoNormFinder++location-backoff-expand-common++location-backoff
simple-np | simple-np++migration-verbs
simple-np | simple-np++TimeNormFinderSUTime++time-stanford
simple-np | simple-np++number-stanford
simple-np | simple-np++location-expand-conjunction
simple-np++GeoNormFinder | simple-np++migration-verbs++GeoNormFinder
simple-vp++simple-np | simple-vp++migration-verbs++simple-np
simple-np++gazetteer++GeoNormFinder | simple-np++migration-verbs++gazetteer++GeoNormFinder
simple-np | simple-np++location-backoff-expand
simple-np | simple-np++gazetteer
simple-np++gazetteer++GeoNormFinder | simple-np++migration-pass++gazetteer++GeoNormFinder
simple-np | simple-np++migration-displace
simple-np | simple-np++GeoNormFinder

kwalcock:
@BeckySharp, now it's ready for a look.  You're too sharp (tm)!

BeckySharp:
love it! ❤️ ❤️ ❤️

kwalcock:
The mentions in the left column of the table above are now chosen over those in the right column.

BeckySharp:
hmm.... i worry about the choosing left vs right... something smells fishy
to me.
should we consider taking a step back and thinking about why we have so
many duplicate mentions and if there's a better way to handle things?

kwalcock:
Sure, but I would call it taking a step forward.  In the past the "choice" was made largely by chance, so I don't mind changing it willfully.  In the table above the left column is (except in maybe one case I discovered later) what Eidos was getting before I made the changes, so this is mostly a reversion to previous behavior.  I'm not clear on how deduplication is performed by Odin.  Does it ever happen that a mention found by Odin is made subordinate to another (as a trigger, argument, etc.) and then is no longer in the list which Odin would consider being a duplicate?  By Odin I might mean inside an Eidos Finder or GlobalAction or something.

BeckySharp:
Yes -- it's DEF a step forward!
re: how deduplication is performed, we try to pick the most
complete/attachiest mention from a given span, but that bit of code is like
our skeleton in the closet. you have made it more sound over the years FOR
sure with tie breaking and all, but it's really complicated, and not that
awesome, and we seem to still be continually dealing with issues with
"duplicates".
I guess I was wondering if we could do it fundamentally differently,
knowing what we "know now" (i.e., 3 years into the program).
Re: the subordinate question -- likely!  I know you dig in and retrieve the
subordinate ones, but that's after the fact, and I know that, for example,
I had to be careful to dig in and manually propagate attachments to args
etc., which felt gross... :)

PR #888, opened 2020-07-16
Zliang sentence classifier

zhengzhongliang:
Sentence classifier code is finished and unit test passed.
* The sentence classifier is accessed by eidosSystem.components.eidosSentenceClassifier.
* The sentence classifier uses eidos system's flat grounder (no new embeddings are generated for sentence classifier).
* When classifying the sentence, the classifier computes a IDF weighted average embedding of the sentence, and matches the embedding against the node embeddings of the flat grounder. 
* New resources are added to the resources file. Corresponding configuration entries are added.
* In the unit test, the accuracy is above 0.69 and f1 is above 0.77.


zhengzhongliang:
Seems like it is complaining unicode in the resource file. Trying to fix it.

kwalcock:
Hi @zhengzhongliang.  The code looks really good.  I'll check on the failing tests.  Still, I'm going to change a few things around.  I wonder about `private def getTop5Scores`.  If this is private and only used in classify, is there reason to take(5) and then right away to take the head of that?  Can the head be taken right away?  It would spare us from creating extra Seqs unnecessarily.

zhengzhongliang:
Thanks Keith,
I think you are right, maybe we can just take the top score, so please change it. 

Also, should I worry about the unicode thing? What I planned to do is to either (1) remove the unicode from the resource file or (2) stop the test from checking unicode of those particular resource files (I believe the two files that have unicodes are only relevant to the sentence classifier). May I know which I need to do or something else? Thanks!

kwalcock:
Thanks!  I hadn't gotten that far yet.  If the unicode is good characters and isn't the result of garbling of the files, then I can exclude them from the test.  I'll check the merge conflicts and then send you a PR with updates.

zhengzhongliang:
Got it. Thanks!

Keith Alcock <notifications@github.com>于2020年7月15日 周三下午7:24写道：

kwalcock:
Is it OK if I replace _SEP_ with a tab?  It makes a difference in split().  split("string") uses a regular expression and split('c') doesn't.  It seems to save a lot of time when we have measured before.  19,000 is a lot of extra work for the computer.

zhengzhongliang:
That would be great. Thanks!

Keith Alcock <notifications@github.com>于2020年7月15日 周三下午7:35写道：

MihaiSurdeanu:
Other than @kwalcock's comment, this looks good to me!
But there are some merge conflicts, and 1 test fails. @zhengzhongliang: can you please fix these?

zhengzhongliang:
> Other than @kwalcock's comment, this looks good to me!
> But there are some merge conflicts, and 1 test fails. @zhengzhongliang: can you please fix these?

Yes I think the failed test is due to the unicodes in the newly added resource file. I will fix those.

kwalcock:
@zhengzhongliang, I'm working on the merge.  These are probably changes that you are not aware of.

kwalcock:
@zhengzhongliang, the unicode in sentence_classifier_idf_weight_1000_20000.txt is not a problem, but it is problematic.  Most of the examples look like "1â 1\/4".  I suspect that "â " came from a special kind of dash or a non-breaking space that has been mangled.  One could track that down, remove the entry, leave it, etc.  I guess that's up to you.  Also in this example you'll notice that \ is being used to escape the /.  These aren't being unescaped when the file is read in.  I should probably just remove the \ since / is not a special character here and there is no good reason to escape it.  Also, there are characters like ()[]{} what have been encoded as in "-lrb-213-rrb-â 237-7832".  These aren't being decoded when read in, so I should replace them with plain text.  It looks like we use the "-lrb-" notation in tags, but not in the words that you are trying to match.

kwalcock:
The other file, SentenceClassifierEvaluation.tsv, just has nonbreaking spaces and special quotes.  It is only used during testing.  I will just exclude it from the check for special characters.

MihaiSurdeanu:
Thanks @kwalcock !

zhengzhongliang:
Thanks Keith!

kwalcock:
@zhengzhongliang, did you have an opinion about this from above: One could track that down, remove the entry, leave it, etc. I guess that's up to you.  Based on the tsv file, I'm inclined to change it to a nonbreaking space.

zhengzhongliang:
Hi @kwalcock, I think we can just remove them. It is very unlikely those special characters will affect the IDF weights a lot or affect the matching scores.

kwalcock:
Remove the entire line or just the character(s)?

zhengzhongliang:
I think removing the entire line is fine.

zhengzhongliang:
@kwalcock It passes all tests. Big big thanks for solving all of the problems!

kwalcock:
It's ready to go big time, I guess.  I'll wait a little while before merging into master in case someone says it doesn't yet belong there.  Thanks for working so hard on it.

kwalcock:
There haven't been any reservations expressed and Mihai approved twice earlier, so I'm pressing the button.  This will help other people stay up to date.

PR #892, opened 2020-07-16
Assimilate sentence classifier

kwalcock:


kwalcock:
@zhengzhongliang, can you look particularly at SentenceClassifier.scala and see if anything looks wrong?  Some things had to be added to protect against cases when we aren't grounding, there is no flat grounder, there was no grounding, etc.  Sometimes I called some library code that you didn't know about or used some obscure Scala feature.  It is supposed to get the same answer as before and follows the same outline.  In general we're trying, maybe unsuccessfully, to refer to config in the object rather than the class, so that provoked some rearranging.  Let me know if there is something you would like changed, or change it after the merge if need be.

MihaiSurdeanu:
+1 for "relevance"

kwalcock:
That's a good name change!  Thanks.

PR #896, opened 2020-08-10
Season Finder

EgoLaparra:
This PR includes a new Finder for season normalization.

- `SeasonFinder` identifies season expressions (e.g. _meher_, _lean season_, _short rainy season_), looks for reference locations and dates in the previous 5 sentences, and normalizes the expressions based on the content of `seasons-db.yml`.
- `seasons-db.yml` contains a list of trigger words that can point to a season expression (e.g. _season_, _meher_) and a dictionary of _start_ and _end_ months per season type in each _geolocation_.


kwalcock:
Can @EgoLaparra verify that the TimeModifier attachment code was purposefully abandoned and not lost in some merge?

kwalcock:
I'm about to spot check for negative cases, but are were certain that the pertinent values are always in the heads in code like this?

```scala
mention._2.attachments.head match {
  case a: LocationAttachment if a.geoPhraseID.geonameID.isDefined => true
...
.map(_.attachments.head).map(_.asInstanceOf[LocationAttachment])

```

Even though filterMentions has already been called before finding the geoLocation, it doesn't seem clear that the label is "Location" because filterMentions has not tested the label.  I don't want it to crash on the 19,999th document.

kwalcock:
In `seasonTokenIdx + 1 to sentSize` and some places nearby in SeasonFinder, should this be `until` in order to not run off the end or to avoid getting the seasonTokenIdx value?

EgoLaparra:
> 
> 
> Can @EgoLaparra verify that the TimeModifier attachment code was purposefully abandoned and not lost in some merge?

Yes, this was abandoned on purpose.

EgoLaparra:
> 
> 
> I'm about to spot check for negative cases, but are were certain that the pertinent values are always in the heads in code like this?
> 
> ```scala
> mention._2.attachments.head match {
>   case a: LocationAttachment if a.geoPhraseID.geonameID.isDefined => true
> ...
> .map(_.attachments.head).map(_.asInstanceOf[LocationAttachment])
> ```
> 
> Even though filterMentions has already been called before finding the geoLocation, it doesn't seem clear that the label is "Location" because filterMentions has not tested the label. I don't want it to crash on the 19,999th document.

I've added an additional test in `filterMentions `for the labels. 

`Location` and `Time` Mentions are created with a single `Location` or `Time` attachment, so _head_ should contain what we are looking for and `filterMentions `has already checked that _attachments_ is not empty.
https://github.com/clulab/eidos/blob/ba43f21459a617faca20e1e81ac6c1ef62b6707c/src/main/scala/org/clulab/wm/eidos/context/TimeNormFinder.scala#L232-L240
https://github.com/clulab/eidos/blob/ba43f21459a617faca20e1e81ac6c1ef62b6707c/src/main/scala/org/clulab/wm/eidos/context/GeoNormFinder.scala#L122-L130

EgoLaparra:
> 
> 
> In `seasonTokenIdx + 1 to sentSize` and some places nearby in SeasonFinder, should this be `until` in order to not run off the end or to avoid getting the seasonTokenIdx value?

You are right (as always). I've changed it.

EgoLaparra:
@kwalcock I've added a `TestSeason` that I forgot to include in the PR.

kwalcock:
Wow!  Did you have that test in the waiting?  For

```scala
case a: LocationAttachment if a.geoPhraseID.geonameID.isDefined => true
```
I was thinking
```scala
case a: LocationAttachment if mention._2.label == "Location" && a.geoPhraseID.geonameID.isDefined => true
```
because with `case l if l.equals("Time") || l.equals("Location") =>`, you wouldn't know which one it is.  Theoretically, the one labeled "Time" could have two attachments, the head being geo and the tail being time (or anything else) instead.  Then it would crash when calculating timeInterval later, unlikely but theoretically.

kwalcock:
It sounds like you know better the circumstances in which this code runs and I should stop worrying.  It does seem to be working just fine so far :-)

EgoLaparra:
> 
> 
> It sounds like you know better the circumstances in which this code runs and I should stop worrying. It does seem to be working just fine so far :-)

If something can happen, it will happen. 😓 

What if I change it to something like:

    val timeInterval = sortedMentions.filter(_.label == "Time")
      .flatMap(_.attachments)
      .find{
        case a: TimeAttachment => a.interval.intervals.nonEmpty
        case _ => false
      }
      .map(_.asInstanceOf[TimeAttachment].interval.intervals)
```

**Comment by EgoLaparra** [comment]
Probably better:
```
    val timeInterval = sortedMentions.filter(_.label == "Time")
      .flatMap(_.attachments)
      .map{
        case a: TimeAttachment if a.interval.intervals.length == 1 => Some(a.interval.intervals.head)
        case _ => None
      }.find(_.isDefined).flatten
```

PR #899, opened 2020-08-12
semantic role label grounding

BeckySharp:
@kwalcock this is **SUPER** not ready yet, but it involves the renaming i alluded to and, more importantly, I _think_ I needed to refactor the SingleOntologyGrounding to handle the new representation, so what I did (note -- this can still be thought of as a draft if you hate it) is to: 
- make a trait (IndividualGrounding)
- there are two case classes which inherit from it:
  -`SingleOntologyNodeGrounding` which is basically the old one, and 
  -`PredicateGrounding` which isn't implemented yet but will eventually hold the new representation

The "backend" of the latter is there, it's basically the `PredicatePackage` case class, but I am waiting to hear back from Ben if we can simply output the flat 4-tuple in the json-ld (i.e., if the `PredicateGrounding` can be based on the 4-tuple), or if we need to output the full structure.   @MihaiSurdeanu may also have an opinion.

OK -- so again, this is NOT done, there are several methods not implemented yet, but it involved a major enough refactor that I wanted you to have early preview ability if you wanted.

Also -- side note, I didn't try hard to work the refactor efficiently into the other CompositionalGrounder (it's added, but just surface assessment, rather than looking at a larger refactor to make better use of the refactor) because we're planning to deprecate it.  I also deprecated the InterventionGrounder while I was at it.

Whew.  what a message!

PS - don't judge me too hard on my attempt at recursion, I haven't tested it yet bc it's not implemented fully.  BUT while I don't want you to judge me, you CAN suggest fixes!!! :D 👍

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
+++ b/src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
@@ -1,0 +1,188 @@
+package org.clulab.wm.eidos.groundings.grounders
+
+import org.clulab.odin.Mention
+import org.clulab.processors.Sentence
+import org.clulab.processors.fastnlp.FastNLPProcessorWithSemanticRoles
+import org.clulab.struct.{DirectedGraph, Interval}
+import org.clulab.wm.eidos.groundings.{ConceptEmbedding, ConceptPatterns, DomainOntology, EidosWordToVec, OntologyGrounding}
+import org.clulab.wm.eidos.mentions.EidosMention
+import org.clulab.wm.eidos.utils.Canonicalizer
+import org.slf4j.{Logger, LoggerFactory}
+import sun.reflect.generics.reflectiveObjects.NotImplementedException
+import SRLCompositionalGrounder._
+
+import scala.annotation.tailrec
+
+case class GroundedToken(index: Int, grounding: OntologyGrounding)
+case class PredicatePackage(predicate: GroundedToken, agent: Seq[GroundedToken] = Nil, theme: Seq[GroundedToken] = Nil, other: Seq[GroundedToken] = Nil)
+
+class SRLCompositionalGrounder(name: String, domainOntology: DomainOntology, w2v: EidosWordToVec, canonicalizer: Canonicalizer)
+  extends EidosOntologyGrounder(name, domainOntology, w2v, canonicalizer) {
+
+  lazy val proc = new FastNLPProcessorWithSemanticRoles
+
+  def inBranch(s: String, branches: Seq[ConceptEmbedding]): Boolean =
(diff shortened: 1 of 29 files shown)


src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
```diff
@@ -0,0 +1,188 @@
...
+    val groundedPredicate = groundToken(predicate, validPredicates)
+    // get the arguments
+    val agents = groundArguments(predicate, AGENT_ROLE, srls, validPredicates)
+    val themes = groundArguments(predicate, THEME_ROLE, srls, validPredicates)
+    val others = groundArguments(predicate, OTHER_ROLE, srls, validPredicates)
+    PredicatePackage(groundedPredicate, agents, themes, others)
+  }
+
+  // Find all arguments from a given predicate of a certain role (e.g., A1, A0, etc) and ground them
+  private def groundArguments(predicate: Int, role: String, srls: DirectedGraph[String], validPredicates: Set[Int]): Seq[GroundedToken] = {
```
MihaiSurdeanu, line 132: In groundArguments, I think we need to syntactically expand the arguments rather than grounding a single token. This is a limitation of the CoNLL SRL annotation, that is, the argument dependency points to the *syntactic head* of the argument, rather than the whole span. We need to identify the span ourselves, using your expand magic.

MihaiSurdeanu: @BeckySharp: this was super fast, and is super cool. You are awesome.

BeckySharp: thanks! i had fun!  Your idea to expand is a good one, and i def wasn't thinking about it.  I may do some lightweight expansion (to the chunk) instead of full expansion which may be likely to suck up the other components....
ok, more soon

BeckySharp:
btw -- @bgyori mentioned he will assemble on the flattened, so we can output the 4-tuple instead of the full structure (yay!) and thanks @bethard for mentioning this option!

BeckySharp:
still in progress, but i think the only thing left is to decide how I want to output stuff
(oh, and test it and all...) :)

MihaiSurdeanu:
Nice!!

kwalcock:
The deserialization has been added, but one or two test files are crashing while grounding.  I'm still looking into it.

kwalcock:
See #907.

kwalcock:
It looks like the problem is with processors.  Also, #908 gets these tests to pass.  It's getting closer.

kwalcock:
This PR is being replaced by a kwalcock-srl_grounding -> master, which is #908 but to master instead of to here.

PR #903, opened 2020-09-08
20200807 add relevance score when annotate

zhengzhongliang:
Add relevance scores to a document attachment when annotating the document. I am sorry that I did not touch the JLD serialization and deserialization stuff, because I am not sure where to change the code. I am afraid I will have leave that to you Keith ...


MihaiSurdeanu:
@zhengzhongliang: this looks good to me, but:
- Some tests are failing,
- Why is the recall in the unit test so low (0.17)? I thought it was higher.

kwalcock:
The unit tests are being addressed in a successor branch #904.  However, I don't know about the recall and hope to hear about that before I merge.  Thanks, @zhengzhongliang.

kwalcock:
@MihaiSurdeanu, @zhengzhongliang can you check this recall issue?

MihaiSurdeanu:
I checked with @zhengzhongliang . It is Ok to merge as is!
Thanks @kwalcock !

Issue #905, opened 2020-09-12
Match error in season finder


kwalcock:
Eidos is having problems processing the attached file.  Remove .txt from the filename.  I'm running ExtractCdrMetaFromDirectory with these arguments: ../corpora/sep2020/input ../corpora/sep2020/output ../corpora/sep2020/time.txt 8.  There's also a VM option: -Xmx20g.  It could be lower.

```
11:56:35.522 [ForkJoinPool-2-worker-1] ERROR o.c.w.e.a.b.ExtractCdrMetaFromDirectory$ - Exception for file ..\corpora\sep2020\input\13e81eea8f1dc02197fe02719631ae94.json
scala.MatchError: season (of class java.lang.String)
	at org.clulab.wm.eidos.context.SeasonFinder.createMention(SeasonFinder.scala:99) ~[classes/:na]
	at org.clulab.wm.eidos.context.SeasonFinder.$anonfun$find$5(SeasonFinder.scala:126) ~[classes/:na]
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234) ~[scala-library-2.12.4.jar:na]
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:32) ~[scala-library-2.12.4.jar:na]
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:29) ~[scala-library-2.12.4.jar:na]
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:191) ~[scala-library-2.12.4.jar:na]
	at scala.collection.TraversableLike.map(TraversableLike.scala:234) ~[scala-library-2.12.4.jar:na]
	at scala.collection.TraversableLike.map$(TraversableLike.scala:227) ~[scala-library-2.12.4.jar:na]
	at scala.collection.mutable.ArrayOps$ofRef.map(ArrayOps.scala:191) ~[scala-library-2.12.4.jar:na]
	at org.clulab.wm.eidos.context.SeasonFinder.$anonfun$find$3(SeasonFinder.scala:125) ~[classes/:na]
	at org.clulab.wm.eidos.context.SeasonFinder.$anonfun$find$3$adapted(SeasonFinder.scala:123) ~[classes/:na]
	at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:241) ~[scala-library-2.12.4.jar:na]
[...]

PR #906, opened 2020-09-12
Add missing catch all in SeasonFinder

EgoLaparra:
Fixes https://github.com/clulab/eidos/issues/905


kwalcock:
Thank you!  We're reading 40,000 documents.  I'll see if any other surprises appear and then whether this fixes them all.

kwalcock:
Thanks again, @EgoLaparra.  In the end there were 7 documents of the 40,000 that exhibited this problem.  They were all fixed with your update.

PR #908, opened 2020-10-01
No longer snapshot processors

kwalcock:


kwalcock:
This is being replaced by the same PR but to master instead.

PR #909, opened 2020-10-05
SRL grounding

kwalcock:


kwalcock:
Runtime on four files without the SRL grounding is 11:40 minutes.  With the grounding it is 31:00 minutes.  This seems to be a very expensive grounding.  It will improve some with parallel dynet.

BeckySharp:
I think it will improve once we don't need to reannotate too.  any chance
you know the hot methods?

kwalcock:
This survived the recent reading test, so I'm going to merge it to simplify the branches because there are changes on top of it.  This grounding can be disabled in the configuration if speed is of the essence.  There is still an outstanding #912 to address, but it can be done in master as long as there isn't a crash, which should be prevented by the next PR.

PR #910, opened 2020-10-06
Egoitz season modifiers

EgoLaparra:
- Look for season modifiers, e.g. "late" in "late rainy season".
- Add or subtract months in the normalization of season expressions according to the values in `seasons-db.yml`.


kwalcock:
Thanks, @EgoLaparra.  It's too bad we didn't get to talk about this development this morning.  Is it something Mihai would want performance numbers for so that we know it's an improvement?  Should I want performance numbers to know that it only added 50 bytes to the file size and 0.1 seconds to processing time?  So far I did notice one strange `preamture: -1` and wonder whether `premature` is meant.  I will check and test the code fairly soon.

MihaiSurdeanu:
i agree. Numbers would be nice :)

EgoLaparra:
I fixed the typo. Thanks for catching it!

kwalcock:
@EgoLaparra, I've resolved the conflicts locally.  There's an EvalSeasonNorm that is called up via TestEvalSeasonNorm, which is normally disabled,  When the test gets enabled, EvalSeasonNorm tries to access a file `Ethiopia_Food_Security_Outlook_1-Feb-17.csv` that doesn't seem to be in the repo.  Can we add that?  It looks like it should be in `/org/clulab/wm/eidos/english/context/timenormeval/` next to `gold_timenorm.csv`.  An exception is thrown without it.

EgoLaparra:
Great, I forgot to add the test files. I've pushed `Ethiopia_Food_Security_Outlook_1-Feb-17.csv` and `Ethiopia_Food_Security_Outlook_1-Feb-17.txt`. The latter has the text input for the evaluation, it won't work without it.

kwalcock:
Thank you!  It seems to work now.

PR #919, opened 2020-11-09
Simplify startup

kwalcock:
Clean up
Update loading
Checkpoint
Use only one kind of ComponentLoader
Get serial to work
Restore it
Try again
Checkpoint
Clean up
Make newComponentLoader
Ignore tests

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/EidosComponents.scala
+++ b/src/main/scala/org/clulab/wm/eidos/EidosComponents.scala
@@ -2,8 +2,6 @@ package org.clulab.wm.eidos
 
 import ai.lum.common.ConfigUtils._
 import com.typesafe.config.Config
-import org.clulab.odin.ExtractorEngine
-import org.clulab.processors.clu.tokenizer.Tokenizer
 import org.clulab.wm.eidos.EidosProcessor.EidosProcessor
 import org.clulab.wm.eidos.actions.CorefHandler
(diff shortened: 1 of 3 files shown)


kwalcock:
Although we are mostly concerned with throughput, the startup time for Eidos is becoming a problem.  The times below are for serial mode startup.  There is a parallel mode as well, and this PR improves on it.  Previous parallel time was 314 seconds and this version finishes in about 274 seconds.  I think the code is much clearer and more maintainable.  These times are without the cached vectors.  At one point we were down to something like 40 seconds.  The big change here is getting processors primed by sending a short sentence through it.

Task | Time (ms)
-- | --
Load AdjectiveGrounder | 7
Load AttachmentHandler | 2
Load ConceptExpander | 1
Load CorefHandler | 8
Load Expander | 5
Load Finders | 7445
Load HedgingHandler | 2
Load MostCompleteEventsKeeper | 21
Load NegationHandler | 5
Load NestedArgumentExpander | 1
Load OntologyHandler | 244968
Load Processors | 40
Load ProcessorsPrimer | 216473
Load SentenceClassifier | 97
Load StopwordManager | 112
  |  
Load components | 470463

image

MihaiSurdeanu:
I hope to reduce the processors load time somewhat in the future. Meanwhile, we can disable the NER and the parser in the processors instance used by Eidos. Eidos only needs SRL, which does not depend on these models.

MihaiSurdeanu:
To disable models, simply write your own version of annotate...(), which calls only the methods that are actually used.

PR #925, opened 2020-11-22
Make components optional

kwalcock:
Also turn everything including processor and annotator into revisors.  Time them all.


kwalcock:
FYI @BeckySharp, you might want to check the end result, definitely not each commit.  In particular, see the EidosSystem class and the components and refiners packages.

PR #929, opened 2020-11-28
Making Subprojects for grounding and common

BeckySharp:
- I am able to publishLocal and use the eidoscommon and ontologies subprojs externally.
- I moved some of the `build.sbt` stuff to `commonSettings` so all subprojs would stay in sync.
- lots of files changed, as would be expected I guess... :(
- lots of tests pass, but I get this when I try to run `sbt test`, did I break dynet?
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000190e9f3cb, pid=94678, tid=0x0000000000015107
#
# JRE version: OpenJDK Runtime Environment (8.0_232-b09) (build 1.8.0_232-b09)
# Java VM: OpenJDK 64-Bit Server VM (25.232-b09 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# C  [dynet2660670813285462059.dll+0x163cb]  _ZN5dynet16ComputationGraph17add_function_nodeEPNS_4NodeE+0x1b
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/bsharp/github/eidos/hs_err_pid94678.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```


BeckySharp:
@kwalcock please let me know what the next steps are.  if you want me to try to track down the core dump thing, please let me know (since I broke it...)
thanks!!

kwalcock:
ComputationGraph::add_function_node is the most common place for it to
crash.  I still don't know why.  The code there is fine.  Does it
happen every single time and on short order?  If so, I would very much
like to see it because getting a reproducible crash has not been
possible.  If it isn't consistent, does adding memory help?  Are you
using processors 8.2.3?

kwalcock:
Ha!  Now it's your problem!!  Thank you very much!!!

kwalcock:
Yes on the processors version.

kwalcock:
Could name := "eidos-eidoscommon" work with not so much eidos?

kwalcock:
The tests passed so you are off the hook :-(

BeckySharp:
re: eidoscommon -- I could rename it to eidos-common

and sorry the tests pass, yes I am using 8.2.3 :)

kwalcock:
+1 for eidos-common and maybe the directory called just common.  There are a few tests that can move, but they can be touched up later.  Thanks for figuring out the common settings.  Other projects have them but somehow we never got around to it.  I am thinking about separating out language components into separate modules somehow.

MihaiSurdeanu:
Hey @kwalcock: if you double (or triple) the memory allocated to dynet in the processors crash experiment, do you observe that it runs two (or three) times as long before it crashes? This may indicate that it is indeed an OOM error.

MihaiSurdeanu:
Also, if we're out of ideas, maybe I can email the clab team personally? Maybe they have an idea on how to debug it?

kwalcock:
That's a good idea emailing.  I can start by filing an issue.  Regarding the timing, I would need to see how regular it will crash in the first place.  There are times when it has crashed almost immediately and other times when it seemingly won't crash at all so that the cause might be a confluence of events, depending on timing or data.

MihaiSurdeanu:
Ok. Then maybe filling an issue is the first step?

kwalcock:
I'm working on the merge conflicts.  Thank you for moving so many of those strings from EidosSystem to EidosParameters.  That's great!  I wonder about STOP_NER.  It is sort of the odd one out.  It looks like the Canonicalizer is already getting a stopwordManaging, so STOP_NER could be added there.  The other use in the RuleBasedEntityFinder isn't covered, but an argument to Finder.fromConfig could take care of that.  Is the StopwordManager a good place for them or should they be in a completely different component like EidosNER?  I'd almost like to read the list from a file there.

BeckySharp:
That all sounds very reasonable to me!

PR #937, opened 2021-01-07
Fix (some) issues with compositional grounder- support entity branch and ignore named entities

zupon:
This PR addresses some issues with the SRL compositional grounder. Some of which had been fixed/discussed earlier, but there was no PR since other issues are still in progress.

- Adds support for the `ENTITY` branch
- Stops compositional grounder from grounding named entities
- Adds DebugGroundingExporter to aid debugging the grounder
- Adds **(IN PROGRESS)** code in `SRLCompositionalGrounder` for trying to deduplicate tuples (e.g. when the same grounding is found for both THEME and PROCESS) -- This can easily be removed for now if it's too messy!


zupon:
@BeckySharp and @kwalcock can you take a look at these changes? A lot of them are things Becky and I added to help with my debugging the compositional grounder back in the day, and they didn't end up getting merged into `master`.

I need them to help me continue working on the collocations with the compositional grounder.

kwalcock:
FYI receipt acknowledged.

src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
```diff
@@ -99,10 +100,19 @@ class SRLCompositionalGrounder(name: String, domainOntology: DomainOntology, w2v
...
+  // Everything after TIME is new to this grounder
+  val STOP_NER: Set[String] = Set("DATE", "DURATION", "LOCATION", "MISC", "MONEY", "NUMBER", "ORDINAL",
+    "ORGANIZATION", "PERSON", "PLACE", "SET", "TIME", "NATIONALITY", "COUNTRY", "TITLE", "STATE_OR_PROVINCE")
+
   override def groundEidosMention(mention: EidosMention, topN: Option[Int] = None, threshold: Option[Float] = None): Seq[OntologyGrounding] = {
     // Do nothing to non-groundable mentions
     if (!EidosOntologyGrounder.groundableType(mention))
       Seq(newOntologyGrounding())
+    // Do nothing to named entities
+    if (mention.odinMention.entities.get.toSet.intersect(STOP_NER).nonEmpty) {
```
BeckySharp, line 113: This looks suspect to me... is it saying that we are not grounding anything that has _any_ ners?

src/main/scala/org/clulab/wm/eidos/exporters/DebugGroundingExporter.scala
```diff
@@ -1,41 +1,175 @@
...
           }
-          pw.println()
-        }
 
       }
     }
+
+
+
+
```
BeckySharp, line 105: I think we should prune some of this commented out stuff unless you need it

src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
```diff
@@ -163,8 +173,111 @@ class SRLCompositionalGrounder(name: String, domainOntology: DomainOntology, w2v
...
 
-    Seq(newOntologyGrounding(srlGrounding))
+    // TODO: make this work - IN PROGRESS
+    //  This attempts to filter out the Theme Process if it is a duplicate of the theme.
+    val tempReturn = Seq(newOntologyGrounding(srlGrounding))
+//    val tuple = srlGrounding.head.predicateTuple
+//    val toReturn = new ArrayBuffer[Seq[OntologyGrounding]]()
+
+    // OntologyGrounding = Map {theme -> List[IndivGr], ...}
+    val bestGrounding: OntologyGrounding = tempReturn.head
```
BeckySharp, line 183: I'm not particularly reviewing this portion, right? bc it's not actually being used (i.e., `tempReturn` is what is returned)... thoughts on that @kwalcock

BeckySharp (commented):
questions/comments added

src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
zupon: It's definitely not *all* ners, but it is a non-short list. See [here](https://github.com/clulab/eidos/pull/937/files#diff-2924da707a0df24b68254e2f9234330df0757e0db36a0103229b520a3860bb43R105) for the specific line with them all listed.

Maybe my understanding was wrong, but I thought one thing we wanted to do was avoid grounding them because there are a bunch of things like woredas and provinces that we didn't want.

src/main/scala/org/clulab/wm/eidos/exporters/DebugGroundingExporter.scala
zupon: Sounds good to me.

src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
BeckySharp: we don't want to ground the locations themselves, and we don't want to include the names of the locations during grounding, but what is the mention here? If we had "Paris poverty" would we skip it completely bc there was an intersection (bc of Paris)?

eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala
```diff
@@ -109,6 +113,32 @@ class EidosPortugueseProcessor(val language: String, cutoff: Int) extends Portug
   def getTagSet: TagSet = tagSet
 }
 
+class EidosCluProcessor(val language: String, cutoff: Int) extends FastNLPProcessorWithSemanticRoles
+  with EidosProcessor with SentencesExtractor with LanguageSpecific {
+  lazy val eidosTokenizer: EidosTokenizer = new EidosTokenizer(localTokenizer, cutoff)
+  override lazy val tokenizer = new EidosTokenizer(localTokenizer, cutoff)
```
kwalcock, line 119: Is there a merge problem here?  Can tokenizer just be eidosTokenizer from the line above?

eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala
```diff
@@ -109,6 +113,32 @@ class EidosPortugueseProcessor(val language: String, cutoff: Int) extends Portug
...
+  val tagSet = new EnglishTagSet()
+  val cluproc = new CluProcessor()
```
kwalcock, line 121: I don't see that cluproc is ever used, so I don't understand why the code is there.  Is it stale?

src/main/scala/org/clulab/wm/eidos/exporters/DebugGroundingExporter.scala
```diff
@@ -1,41 +1,175 @@
...
-              grounding.grounding.foreach { gr =>
-                pw.println(s"  --> grounding: ${gr.name}")
-                pw.println(s"      score: ${gr.score}")
+              pw.println()
+
+              pw.println(s"Compositional Grounding:")
+              groundingComp match {
+                case None => pw.println("no grounding...")
+                case Some(grounding) =>
+                  var tupNum: Int = 0
```
kwalcock, line 67: It's not important, but there is a utils.Counter in eidoscommon that might be used.

eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala
BeckySharp: It's used when we want to have access to semantic roles

BeckySharp: but enabled through the config, so it may not be greppable easily in the code

src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
zupon: The mentions I saw were typically things like `Addis Ababa`, `South Sudan`, and `Minister Hailemariam`. For the first two, it makes sense to just skip the whole thing, but what about people with titles (or locations such as `Somali region`)? Should we keep the more generic word and only skip the ner piece? Or will that cause problems?

src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
kwalcock: If it is code that would actually run in production, then I'm glad it isn't being executed.  However, if it is standing by waiting to aid in debugging, it will be difficult to maintain.  Sometimes I throw things into a def and just never call it.  The overhead seems acceptably low and one can be sure the code continues to track other changes, like renaming.  (IMO)

BeckySharp:
i think so -- and we should think carefully about what's in the entity
branch of the ont, and what we think a given extraction/component of an
extraction *should* ground to (or if it even should ground)

src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
```diff
@@ -273,7 +386,8 @@ class SRLCompositionalGrounder(name: String, domainOntology: DomainOntology, w2v
       GroundedSpan(trimmedChunk, propertyOpt.get, isProperty = true)
     } else {
       // Otherwise, ground as either a process or concept
-      GroundedSpan(trimmedChunk, groundToBranches(Seq(CONCEPT, PROCESS), trimmedChunk, s.sentence, topN, threshold), isProperty = false)
+      GroundedSpan(trimmedChunk, groundToBranches(Seq(CONCEPT, ENTITY, PROCESS), trimmedChunk, s.sentence, topN,
```
kwalcock, line 389: This is nitpicky, but I see several of these Seqs around.  Could they be turned into variables with explanatory names and added to the companion object?  They are being recreated repeatedly.

BeckySharp: +1 for this

eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala
kwalcock: IntelliJ's "Find Usages" is apparently blind to it.  Perhaps the use is outside its field of vision.  The `FastNLPProcessorWithSemanticRoles` has a `CluProcessor` itself, so I think one can do a
```scala
  val cluproc = this.cluProcessor
```
or just use `cluProcessor` instead of `cluproc`.  The things are fairly large and take time to load.

BeckySharp:
Good point -- so FastNLPWithSemanticRoles I think does semantic roles diff
than cluprocessor, so we should extend Cluprocessor instead, and not have a
`cluproc`?

kwalcock:
IDK which processor is better to use, but the EidosCluProcessor already has access to a CluProcessor via the FastNLPProcessorWithSemanticRoles from which it inherits.  If the inheritance changes it may have to organize its own CluProcessor, unless it happens to be a CluProcessor itself.  It's a very minor detail, but one in master, so I have to be a little careful, especially given memory issues and slow startup time.

MihaiSurdeanu:
FastNLPWithSemanticRoles uses the same SRL system as CluProcessor.

kwalcock:
Although, if this EidosCluProcessor is only used in special cases and not the default configuration, then I should not worry so much.

BeckySharp:
> FastNLPWithSemanticRoles uses the same SRL system as CluProcessor.

I thought you only added the expanded to CluProcessor?

and @kwalcock I don't have strong feelings, i like not having extra processors for sure!  We need access to the semantic roles used by the SRL grounder, no more no less, and we need it to be compatible with the EidosProcessor "framework", otherwise, however we can do it with minimal memory usage is ideal!

MihaiSurdeanu:
They should use the same code.

eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala
zupon: You are correct! Will fix in next commit.

eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala
zupon: I think Keith is correct here. I tried running the debug grounder without `cluproc` (and without the line initializing dynet), and it still seems to work, giving me semantic roles and everything.

src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
zupon: Good idea! I added some variables in the object and replaced the various Seqs above.

kwalcock (commented):
Thank you!

zupon:
I removed the in-progress and unused code I was working on in `SRLCompositionalGrounder`, since it's not my focus right now and keeping it all there makes it pretty ugly.

@kwalcock is there anything else I should change before this can be merged?

kwalcock:
LGTM.  Thanks.  Shall I go ahead and press the merge button?

BeckySharp:
good with me!

zupon:
Huzzah! Thank you both!

PR #940, opened 2021-01-21
remove proc model deps

BeckySharp:
As of 8.2.3 we don't need these anymore I think, assuming tests pass (!)


kwalcock:
Thanks.  I don't really know why those dependencies were there.

PR #942, opened 2021-01-28
Make a SimpleEidos object to configure Eidos

kwalcock:
SimpleEidos.newEidosSystem(useGeoNorm: Boolean, useTimeNorm: Boolean): EidosSystem should get an appropriate "simplified" eidos.

### DIFF
--- a/src/main/resources/simpleEidos.conf
+++ b/src/main/resources/simpleEidos.conf
@@ -1,0 +1,11 @@
+include required("eidos.conf")
+
+EidosSystem {
+  // geonorm, timenorm, and seasons should be included so that they are available if/when needed.
+  // If more finders are added to eidos.conf, then they should probably be added here as well.
+  finders = ["gazetteer", "rulebased", "geonorm", "timenorm", "context", "causal", "seasons"]
+}
+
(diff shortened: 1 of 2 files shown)


kwalcock:
I'm not sure who this is for and it doesn't necessitate any changes to Eidos proper, so I'll PR it to master and tag Mihai.  It's only about 25 lines of code without the sample main, and could be packaged in lots of other ways.  This is the gist of it though.

MihaiSurdeanu:
- I would recommend adding an `apply` method to SimpleEidos to make the construction trivial.
- Also, would it be possible to add a simple output format as well? That is, a format that focuses on the triple cause/relation/object, with some provenance as well (sentence and offsets)?

MihaiSurdeanu (approved):

kwalcock:
The first change was straightforward.  It is a little bit deceptive because one might expect a SimpleEidos to be returned, but one gets a full-fledged EidosSystem instead.  Example output is forthcoming.  That seems like a something that will change often based on refined needs.

kwalcock:
I'll add output in a new PR.

PR #943, opened 2021-01-28
Provide simple output mechanism

kwalcock:


kwalcock:
In this (draft) version, output looks like this:


Index | Type | Cause | Effect | Sentence | Start | End | Text
-- | -- | -- | -- | -- | -- | -- | --
0 | EidosEventMention | 1 | 3 | 1 | 17 | 93 | Water trucking in Ethiopia has decreased over August due to the cost of   fuel
1 | EidosTextBoundMention |   |   | 1 | 81 | 93 | cost of fuel
2 | EidosTextBoundMention |   |   | 1 | 70 | 73 | due
3 | EidosTextBoundMention |   |   | 1 | 17 | 43 | Water trucking in Ethiopia

MihaiSurdeanu:
Thanks! I think this is nearly there. But I would propose the following columns:
- Index
- Cause text
- Cause token start
- Cause token end
- Cause polarity (increase/decrease)
- Effect text
- Effect token start
- Effect token end
- Tokenized sentence

Also, you only need to output causal relations not entities.

kwalcock:
The output looks like the table below.  We don't have a definitive representation of polarity as far as I know.  In the random example sentence copied from elsewhere in the code, the decrease happens to be on the effect.  For now I just output all the counts.  It can be easily changed.

Index | Cause text | Cause token start | Cause token end | Cause Inc/Dec | Effect text | Effect token start | Effect token end | Effect Inc/Dec | Tokenized sentence
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
0 | cost of fuel | 11 | 14 | 0/0 | Water trucking in Ethiopia | 0 | 4 | 0/1 | Water trucking in Ethiopia has decreased over August due to the cost of   fuel

PR #948, opened 2021-02-02
Add simplified test outline for posneg groundings

kwalcock:


kwalcock:
@JerryZeyu, here is a simplified example of calling the grounding functions from within the unit test framework.  It may be easier to understand than TestGrounding or TestEvalGroundings.

@BeckySharp, @MihaiSurdeanu, fyi, the name change of Word2Vec in processors that was mentioned this morning is handled by a mapping in eidos.  The code should be up to date.

```scala
import org.clulab.embeddings.{WordEmbeddingMap => Word2Vec}
```

PR #949, opened 2021-02-03
Suppress evicted notices

kwalcock:
This is an experiment

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -140,6 +140,30 @@ libraryDependencies ++= {
   )
 }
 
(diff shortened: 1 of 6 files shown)


kwalcock:
I'd like to see the medium-term effects of suppressing these warnings that we are otherwise ignoring anyway.  They may, for instance, be hiding messages that we are interested in.  One customer was concerned about them today as well.

kwalcock:
This seems to hang Jenkins in some strange way.  I'm deleting the PR and will try again later.

PR #951, opened 2021-02-03
Update .gitignore to account for VS Code use

kwalcock:


kwalcock:
I'm just cleaning up orphaned local branches...

PR #953, opened 2021-02-04
Update Bagger for potential bag of words use

kwalcock:


kwalcock:
@zupon, this may or may not be useful for your bag of bigrams task.

PR #954, opened 2021-02-04
Check for Java 8

kwalcock:


kwalcock:
It seems that we are expecting people to use Java 8.  Unit tests will fail with other versions, 11 for example.  Certain garbage collectors have been discontinued in later versions.  An error message is now displayed if anything other than Java 8 is found.  Perhaps we will discover some unexpected behavior by users.  I have a paragraph prepared for the readme file that recommends version 8 to accompany the change.

BeckySharp:
+1000

kwalcock:
I *think* the tests passed in part by luck.  We enabled testing of the webapp recently.  Two sets of tests are usually running in parallel.  I wonder whether there is a problem if one webapp test tries to run while another is already.  Won't there be contention for port 9000 or won't the two interact?  Those tests take quite a bit of time and perhaps the testing should not be done with the rest of the project.  The compilation was similarly added and that has been helpful.  At least the webapp always builds even when something else was refactored.

PR #958, opened 2021-02-09
Add Odometer

kwalcock:
This helps generate all combinations for brute force searching

Add RangeOdometer

### DIFF
--- a/eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/utils/Odometer.scala
+++ b/eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/utils/Odometer.scala
@@ -1,0 +1,71 @@
+package org.clulab.wm.eidoscommon.utils
+
+import scala.collection.mutable
+
+// This AnyRef thing is not absolutely complete.
+// See https://stackoverflow.com/questions/2440134/is-this-the-proper-way-to-initialize-null-references-in-scala.
+class SeqOdometer[T <: AnyRef](val sequences: Array[Seq[T]]) extends Iterator[mutable.ArraySeq[T]] {
+  require(sequences.nonEmpty)
+  require(sequences.forall(_.nonEmpty))
+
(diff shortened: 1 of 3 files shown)


BeckySharp:
@kwalcock mind if I ask what this does?  I saw the PR go in and was curious

kwalcock:
No questions.  It's a secret.  It's a backport whose use is a long ways away: https://github.com/lum-ai/eidos/blob/39dd25d789136e2b198208a92d921fee41704a4c/src/main/scala/org/clulab/wm/eidos/mentions/MentionUtils.scala#L101.  It combines several sequencey things and gives all possibilities in a single iterator so that it can be used in map, foreach, etc.

BeckySharp:
cool -- fwiw, esp if you're in the prep stages, I really think this functionality (https://github.com/lum-ai/eidos/blob/39dd25d789136e2b198208a92d921fee41704a4c/src/main/scala/org/clulab/wm/eidos/mentions/MentionUtils.scala#L100) should exist in Processors.  That is, a sentence should be able to give the minimum graph distance in its X parse (e.g., SRL, syntax, etc).

kwalcock:
When reading the code I was concerned about whether there actually was a distance between any two tokens.  What is guaranteed about the connectedness of the graph?  In reach I recently had to catch cases where the graph seemed incomplete:
https://github.com/clulab/reach/blob/30cd0daa1c1921767fb6f6935a21460ca94d1438/main/src/main/scala/org/clulab/polarity/LinguisticPolarityEngine.scala#L85-L93.  It does seem like whatever makes the graph, processors, should also provide general support for them.

BeckySharp:
Agreed on all points.  I guess it could return a `Seq[Int]` of distances, so empty if no connection, and shortest first if there are >1.  IDK if this formulation is inherently less efficient.  Another option could be `Option[Int]` if it's more efficient...?

PR #964, opened 2021-02-17
Account for negative examples in ontologies

kwalcock:


kwalcock:
@JerryZeyu, this is your branch after master was merged into it in order to clear up some conflicts that happen when a branch has been checked out for a long time.  (So you will want to pull before continuing.)

Just so you know, I will be moving some things around, in particular wm_posneg_metadata.yml, which I put in a src/main for convenience, but which needs to be in a src/test so that it isn't deployed.  It's not a simply cut and paste, so I'll do it.

In the meantime can you work on these changes?

* Convert the hyperparameters right away from Double to Float so that the conversion happens only once.
* Remove the commented out lines of code and the todo lines related to your changes, anything that isn't useful documentation.
* Can you prevent values from being less than -1 without performing the subtraction twice?

There may be some suggestions related to the test, but probably not until the files have been moved around.

Thank you.

src/test/resources/englishTest.conf
```diff
@@ -2,15 +2,15 @@ include required("eidos.conf") // eidos.conf defaults to English, so start there
...
 }
 
 ontologies {
   // These are values that typically must be changed during testing.
-  useGrounding = false
+  useGrounding = true
   //wordToVecPath = ${EidosSystem.path}/w2v/glove.840B.300d.txt // Local resource
   wordToVecPath = /org/clulab/glove/glove.840B.300d.txt // Remote resource
-  ontologies = ["wm_compositional", "un", "wdi", "fao", "mitre12", "who"] // , "interventions", "icasa", "mesh"]
+  ontologies = ["wm_posneg"] // , "interventions", "icasa", "mesh"]
```
MihaiSurdeanu, line 13: We should continue to point to wm_compositional by default here.

MihaiSurdeanu (approved):
LGTM, with the exception of that comment in the config file.

kwalcock:
BTW this is still a draft that will go through a couple of more revisions.  Please don't merge yet.

JerryZeyu:
Hi Keith,

I have completed these three things and push the changes to github. Thank
you very much.

Best,
Zeyu

kwalcock:
Thank you.  I'll move some stuff around and restore the original grounding, which is probably the reason that the other tests are failing.

kwalcock:
I'm about to check this in, but I don't think that master should go to Ben unless it comes with a lot of instructions.  Master is meant to load the ontology from github automatically.  It won't use a local copy with negative examples without changes and those changes should be added to a branch to get him started.

src/test/resources/englishTest.conf
kwalcock: FWIW this file is englishTest.conf used only in testing, not eidos.conf used in production.  Usually testing of grounding is turned off because it is too expensive.

bgyori:
I'm listening, in case there is something important to keep in mind when using this. The idea here is to use the polarity labels from the ontology to account for inherently positive/negative concepts when determining the event-level polarity, right?

MihaiSurdeanu:
Correct. We will soon merge this in master. Then we should discuss how we should test this.

PR #977, opened 2021-03-05
Handling RecursionError in pdf_to_text

EgoLaparra:
Process a pdf page disabling box_flow  if it raises a RecursionError.


kwalcock:
Thanks!  It's great to know the page number as well.

PR #992, opened 2021-03-23
addresses bug where proc prop shows up without a process

zupon:
This addresses a bug @bgyori brought up where the SRLCompositionalGrounder will produce a grounding that includes e.g. a `theme`, and a `process property` but no `process`. We should not have properties without the basic slot. This fix looks at the resulting tuple, checks if there is a `process property` but no `process`, and if so---and if there is not already a `theme process`---moves the `process property` to the `theme property` slot.

What should we do if there is a `process property` and no `process`, but there is already a `theme property`? Should we just delete the `process property`?

Let me know what needs changed!

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
+++ b/src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
@@ -162,11 +162,29 @@ class SRLCompositionalGrounder(name: String, domainOntology: DomainOntology, w2v
         val sortedSliced = groundings.sortBy(-_.score)
         sortedSliced.take(topN.getOrElse(sortedSliced.length))
     }
-
+    val returnedSRLGroundings = new ArrayBuffer[PredicateGrounding]
+    for (predGrounding <- srlGrounding) {
+      val tupleTheme = predGrounding.predicateTuple.theme
+      val tupleThemeProp = predGrounding.predicateTuple.themeProperties
+      val tupleProc = predGrounding.predicateTuple.themeProcess
+      val tupleProcProp = predGrounding.predicateTuple.themeProcessProperties
+      // move process property to theme property if process is empty AND no existing theme property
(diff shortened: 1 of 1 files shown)


src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
```diff
@@ -162,11 +162,29 @@ class SRLCompositionalGrounder(name: String, domainOntology: DomainOntology, w2v
...
+      val tupleTheme = predGrounding.predicateTuple.theme
+      val tupleThemeProp = predGrounding.predicateTuple.themeProperties
+      val tupleProc = predGrounding.predicateTuple.themeProcess
+      val tupleProcProp = predGrounding.predicateTuple.themeProcessProperties
+      // move process property to theme property if process is empty AND no existing theme property
+      // THIS IS A HACK
+      if (tupleThemeProp.grounding.isEmpty && tupleProc.grounding.isEmpty && tupleProcProp.grounding.nonEmpty) {
+        val newTupleThemeProp = tupleProcProp
+        val newTupleProcProp = tupleProc
+        val newPredGrounding = PredicateGrounding(PredicateTuple(tupleTheme, newTupleThemeProp, tupleProc,
```
BeckySharp, line 176: `tupleTheme` here could be empty, right? is that desired (i.e., not checking)

BeckySharp (commented):
This checks for one edge case, but I'd rather deal with it where it originates.  Is it the case that this happens when a process is promoted to theme, but the property isn't?  Some other situation?

src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
zupon: Good point...this would also give us the situation where we somehow have a `theme property` but no `theme`.

Issue #978, opened 2021-03-06
TestDoc3, Paragraph 9, edges 2


kwalcock:
Sometimes the tests fails with
```
[info] TestDoc3 Paragraph 9
[info] - should have correct edges 1
[info] - should have correct edges 2 *** FAILED ***
[info]   List("
[info]   Errors:
[info]          Could not find EdgeSpec [persistently well above-average rainfall over the western Ethiopia highlands|+INC(above-average, Quant: persistently, Quant: well)+QUANT(above-average, Quant: persistently, Quant: well)+GEO(Ethiopia)+TIME(the coming weeks)]->(Causal)->[flooding|+TIME(the coming weeks)]
[info]   Mentions:
[info]          #0: TextBoundMention@839043712[likely]
[info]          #1: TextBoundMention@1396362116[persistently well above-average rainfall over the western Ethiopia highlands]
[info]          #2: TextBoundMention@1992834692[reduce]
[info]          #3: TextBoundMention@931963655[good performance of seasonal rainfall]
[info]          #4: TextBoundMention@152069899[above-average]
[info]          #5: TextBoundMention@1792444175[Cropping conditions]
[info]          #6: TextBoundMention@861145747[continued rains]
[info]          #7: TextBoundMention@1951454996[forecast]
[info]          #8: TextBoundMention@308425365[northern areas]
[info]          #9: TextBoundMention@1611219991[Ethiopia]
[info]          #10: TextBoundMention@633241497[well]
[info]          #11: EventMention@348363294[continued rains could also reduce the impact of Fall Armyworm](effect = (TextBoundMention@2094461533[impact of Fall Armyworm]), cause = (TextBoundMention@295704041[continued rains]))
[info]          #12: TextBoundMention@1635076512[persistently]
[info]          #13: TextBoundMention@1668370723[start]
[info]          #14: TextBoundMention@892928291[potential]
[info]          #15: TextBoundMention@161345700[six regions]
[info]          #16: TextBoundMention@50903213[flooding]
[info]          #17: TextBoundMention@1729349166[moderate to heavy]
[info]          #18: TextBoundMention@1897015215[central areas]
[info]          #19: TextBoundMention@1598873520[due]
[info]          #20: EventMention@1338039986[persistently well above-average rainfall over the western Ethiopia highlands could result in flooding](effect = (TextBoundMention@1205583422[flooding]), cause = (TextBoundMention@1396362116[persistently well above-average rainfall over the western Ethiopia highlands]))
[...]

PR #993, opened 2021-03-25
filter extractions with likely misparsed triggers

BeckySharp:
I noticed that in certain situations, some adjs were misparsed, and so they ended up triggering a causal event when they shouldn't.
Example: "Concurrently , female holders have limited access to extension and advisory services." (the original sentence was longer).

I made a filter and added a test, which passes... but I am unsure about the consequences.
I'd like to (a) see if the tests pass, and (b) perhaps have someone (me? other?) check the differences on a small-med set of docs, hand check a sample, and see if it's a net good change.


BeckySharp:
thanks jenkins!
ok, working on:
```Failed tests:
[error] 	org.clulab.wm.eidos.text.english.eval6.TestDoc5
[error] 	org.clulab.wm.eidos.text.english.eval6.TestDoc3
```

kwalcock:
Regarding TestDoc3, see #978.  It may not be your fault.  It was failing intermittently as recently as last night.  Of course if you can figure out why it's unstable, that would be great, too.  I may get to it as part of the testing framework revision.

BeckySharp:
> Regarding TestDoc3, see #978. It may not be your fault. It was failing intermittently as recently as last night. Of course if you can figure out why it's unstable, that would be great, too. I may get to it as part of the testing framework revision.

to bump this -- any news?

kwalcock:
I *think* this was solved yesterday with https://github.com/clulab/eidos/pull/1006.

BeckySharp:
> I _think_ this was solved yesterday with #1006.

can we trigger a retry on the tests?

kwalcock:
The change in master would need to be merged into jj_triggers, I think.

kwalcock:
Otherwise the test result might not change.

BeckySharp:
@kwalcock what happened to jenkins?

kwalcock:
It is overloaded with 6 jobs in the queue.  I may also have duplicated what you just did.  I only cherry picked the two changes, but it looks like you merged in master.  Sorry.  It will keep Jenkins busy for longer than normal.

BeckySharp:
looks like it passes, are we good to merge @MihaiSurdeanu ?

MihaiSurdeanu:
Yep!

BeckySharp:
ping @kwalcock :) :)

kwalcock:
Thanks for the reminder.

PR #1000, opened 2021-04-20
Update processors

kwalcock:


BeckySharp:
wow! pr 1000! cause for celebration? happy birthday, Eidos!

kwalcock:
This passed the tests on River, but now the tests don't pass on my machines even with 16 or even 32GB configured.  They crash out saying that no more memory is available.

MihaiSurdeanu:
This is most likely caused by the storing of glove embeddings twice, including in the less efficient dynet representation...
I am working on yet another idea that might fix this. ETA <= 1 week, depending on how much actual time to work on it I get...

kwalcock:
I have to take some of that back, especially for the 32GB  The messages I saw were related to the test which specifically exhausts the memory and makes sure eidos recovers.  It was not recovering with 16GB, but that might not be because of the 16GB.

PR #1004, opened 2021-05-06
added case class and pruning by threshold

BeckySharp:
FYI @JerryZeyu @ZhengTang1120 @reynoldsm88

### DIFF
--- a/src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
+++ b/src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
@@ -2,7 +2,7 @@ package org.clulab.wm.discovery
 
 import org.clulab.dynet.Utils
 import org.clulab.wm.eidos.extraction.RuleBasedEntityFinder
-import com.typesafe.config.{Config, ConfigFactory}
+import com.typesafe.config.ConfigFactory
 import org.clulab.processors.clucore.CluCoreProcessor
 import org.clulab.wm.eidos.EidosSystem
 import org.clulab.wm.eidos.document.AnnotatedDocument
@@ -11,8 +11,6 @@ import org.clulab.wm.eidoscommon.EnglishTagSet
 import org.clulab.wm.eidos.utils.StopwordManager
 import org.clulab.processors.{Document, Processor}
 import scala.collection.mutable
-
-import scala.collection.mutable.ArrayBuffer
 
 case class CdrDocument(docid: String, sentences: Seq[ScoredSentence])
 case class ScoredSentence(text: String, start: Int, end: Int, score: Double)
@@ -35,7 +33,7 @@ class ConceptDiscovery {
     val entityFinder = RuleBasedEntityFinder.fromConfig(config, tagSet, stopwordManager)
     val processor = new CluCoreProcessor()
     val conceptLocations = mutable.Map.empty[String, Set[String]].withDefaultValue(Set.empty)
-    for ((cdr, idx) <- cdrs.zipWithIndex) {
+    for (cdr <- cdrs) {
       // make a Processors Document, pruning sentences with a threshold if applicable
       val document = mkDocument(processor, cdr.sentences, sentenceThreshold)
       val mentions = entityFinder.find(document)
--- a/src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
+++ b/src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
@@ -14,8 +14,9 @@ import scala.collection.mutable
 
 import scala.collection.mutable.ArrayBuffer
 
-case class CdrDocument(text: String, docid: String, sentenceRankings: Seq[SentenceScore])
-case class SentenceScore(start: Int, end: Int, score: Double)
+case class CdrDocument(docid: String, sentences: Seq[ScoredSentence])
+case class ScoredSentence(text: String, start: Int, end: Int, score: Double)
+
 
 case class Concept(phrase: String, documentLocations: Set[String]) {
   def frequency: Int = documentLocations.size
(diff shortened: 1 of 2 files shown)


BeckySharp:
@reynoldsm88 -- the case class here (`CdrDocument`, which in turn uses the case class `SentenceScore`) is what you'll be targeting for the mapping you'll do on your end.

I tried to make the `SentenceScore` definition match closely with what you already have in the CDR but IDK what you create when you instantiate the CDR object, so if there are things that I can adjust in the definition to make the mapping more trivial, please let me know.

BeckySharp:
@johnhungerford

BeckySharp:
I **think** I fixed the merge conflict right, but sorry if not

kwalcock:
Thank you, @BeckySharp!

src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
```diff
@@ -46,6 +51,31 @@ class ConceptDiscovery {
...
+      val validSentences = rankedSentences.slice(0, lastValid)
+      val sentencesByOffsets = document.sentences.map(s => ((s.startOffsets.head, s.endOffsets.last), s)).toMap
+      val keptSentences = validSentences
+        // keep ones that are in the validSentences
+        .map(s => sentencesByOffsets((s.start, s.end)))
+        // put them back in order
+        .sortBy(_.startOffsets.head)
+        // get their text
+        .map(_.getSentenceText)
+        .toArray
```
bethard, line 73: If `keptSentences` is just an `Iterable[String]`, why not just:

    val keptSentences = rankedSentences.filter(_.score <= sentenceThreshold).sortBy(_.start).map {
        case SentenceScore(start, end, _) => text.slice(start, end)
    }

src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
```diff
@@ -22,18 +25,20 @@ case class RankedConcept(concept: Concept, saliency: Double)
...
     val config = ConfigFactory.load("reference")
     val entityFinder = RuleBasedEntityFinder.fromConfig(config, tagSet, stopwordManager)
     val processor = new CluCoreProcessor()
+    val textListAll = new ArrayBuffer[String]()
+    val sentenceListAll = new ArrayBuffer[String]()
     val conceptLocations = mutable.Map.empty[String, Set[String]].withDefaultValue(Set.empty)
-    for (path <- documents) {
-      val text = FileUtils.getTextFromFile(path.toString)
-      val document = processor.annotate(text)
+    for ((cdr, idx) <- cdrs.zipWithIndex) {
```
bethard, line 39: Is there a reason there's an `idx` and `zipWithIndex` here? Seems unnecessary.

src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
```diff
@@ -22,18 +25,20 @@ case class RankedConcept(concept: Concept, saliency: Double)
...
+  def discoverConcepts(cdrs: Seq[CdrDocument], sentenceThreshold: Option[Double]): Set[Concept] = {
     Utils.initializeDyNet()
     val tagSet = new EnglishTagSet()
     val Config = EidosSystem.defaultConfig
     val stopwordManager = StopwordManager.fromConfig(Config, tagSet)
     val config = ConfigFactory.load("reference")
     val entityFinder = RuleBasedEntityFinder.fromConfig(config, tagSet, stopwordManager)
     val processor = new CluCoreProcessor()
+    val textListAll = new ArrayBuffer[String]()
+    val sentenceListAll = new ArrayBuffer[String]()
```
bethard, line 37: These don't seem to be used at all.

BeckySharp:
ok, modified the case classes bc @johnhungerford indicated they already slice up the doc into sentences, so we can use that and avoid offset alignment sadness.

re: unused stuff, yes, happened in the merge conflict, fixed
re: idx etc and other formats, how much of this do you want me to address vs a student?

src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
BeckySharp: idk, it was there when i started and idk what it was/is to be used for

BeckySharp:
note @MihaiSurdeanu that i commented out the tests bc the input format changed a lot, so they will need to change too

src/test/scala/org/clulab/wm/discovery/TestConceptDiscovery.scala
```diff
@@ -48,20 +48,20 @@ class TestConceptDiscovery extends AnyFlatSpec with Matchers {
...
+//    }
+//
+//    it should "have reasonable frequency estimates" in {
+//      concepts.foreach{
+//        case c @ Concept("food security", _) => c.frequency should be > 4 // actual: 7
+//        case c @ Concept("access", _) => c.frequency should be > 2 // actual: 4
+//        case c @ Concept("availability", _) => c.frequency should be > 2 // actual: 4
+//        case _ =>
+//      }
+//    }
```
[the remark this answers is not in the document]
BeckySharp: i feel like this is a good task for a student...?

bethard:
@BeckySharp The `idx` thing was something that was not in the previous code, so presumably you're the best one to decide if you need it or not. I'm not sure what "and other formats" refers to. Was it the more efficient rewrite of your filtering by threshold code?

src/test/scala/org/clulab/wm/discovery/TestConceptDiscovery.scala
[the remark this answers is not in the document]
bethard: Sure. Feel free to assign one.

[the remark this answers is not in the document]
bethard: But in general, I would not commit commented out tests. Better to let the compile error stay so that we all know the work is incomplete.

BeckySharp:
> s something that was not in the previous code, so presumably you're the best one to decide if you need it or not. I'm not sure what "and other formats" refers to. Was it the mo

it was there when i grabbed the branch, but maybe someone had already removed it when you looked (which would also explain the merge conflicts even though I started with the `wg_1` branch). I really have no idea why it was there, but again, I don't need it so anyone can remove it

BeckySharp:
> and other formats

IDK what of your review still applies now that I re-write things with the diff input.  But, if you still want things done differently/more efficiently, please let me/ @JerryZeyu / @ZhengTang1120  know.

src/test/scala/org/clulab/wm/discovery/TestConceptDiscovery.scala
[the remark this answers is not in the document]
BeckySharp: ok, then @JerryZeyu please uncomment out this test and change the input to match what is now expected of the API, thanks!

src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
```diff
@@ -31,18 +35,41 @@ class ConceptDiscovery {
...
+
+      // only accept sentences with a saliency above some threshold
+      val lastValid = rankedSentences.indexWhere(_.score < sentenceThreshold.get)
+      val validSentences = rankedSentences
+        .slice(0, lastValid)
+        // put them back in order
+        .sortBy(_.start)
+
+      // annotate and return as a Document
+      processor.annotateFromSentences(validSentences.map(_.text))
```
bethard, line 72: @BeckySharp I think this whole method is equivalent to the following code:

    val validSentences = sentenceThreshold match {
      case None => sentences
      case Some(threshold) => sentences.filter(_.score >= threshold).sortBy(_.start)
    }
    // annotate and return as a Document
    processor.annotateFromSentences(validSentences.map(_.text))

I can make this change if that's true, but since there are no tests for this behavior, I need some confirmation that this is indeed the intent.

bethard: Or maybe even more succinctly:

    val selectedSentencesOption = sentenceThreshold.map(t => sentences.filter(_.score >= t))
    processor.annotateFromSentences(selectedSentencesOption.getOrElse(sentences).map(_.text))

bethard:
@JerryZeyu I've restored the tests, so you don't need to do that. But it would be good if you could try to add an additional test of @BeckySharp's feature where some of the input sentences are filtered if they don't have a high enough score. That would probably mean changing `texts` from a `Seq[Seq[String]]` to a `Seq[Seq[(String, Double)]]` and then making up scores for each sentence. (It wouldn't matter what scores you make up, as long as you test that the resulting concepts are appropriately filtered.)

bethard:
I think this can be merged. The extra test Zeyu should write can go into the [wg1 branch](/clulab/eidos/tree/wg1) after the merge.

BeckySharp:
> I think this can be merged. The extra test Zeyu should write can go into the [wg1 branch](/clulab/eidos/tree/wg1) after the merge.

thanks @bethard !

PR #1005, opened 2021-05-06
Account for matches is separate sentences

kwalcock:


kwalcock:
is -> in

zupon:
Thanks for looking into this!

PR #1009, opened 2021-05-10
Simplifies concept ranking code

bethard:
@ZhengTang1120 Please review.

I think this simplified code is equivalent to your code, but since there are no tests of the actual concept ranking (only tests of the filtering that ranking does), I can't be sure.

If you agree it's equivalent, please merge.

### DIFF
--- a/src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
+++ b/src/main/scala/org/clulab/wm/discovery/ConceptDiscovery.scala
@@ -59,55 +59,31 @@ class ConceptDiscovery {
     }.toSet
   }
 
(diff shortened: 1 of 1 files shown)


kwalcock:
It sounds like this discovery code will move to processors.  If not, we need to figure out why it has broken the eidos tests.  I see changes in avoidLocal.yml and RuleBasedEntityFinder.

In the meantime, just getting this into wg1 would be welcome.  @ZhengTang1120, are you able to review?

kwalcock:
OK.  I have code for the CDR conversion that will eventually move with it.

BeckySharp:
moved to processors, with the suggested edits, thanks @bethard !

PR #1011, opened 2021-05-11
Merge in changes from hackathon

kwalcock:
We're hoping to add these changes to processors instead of eidos.  This PR has been made in part to facilitate discussion.


MihaiSurdeanu:
Good points. All valid.

kwalcock:
I just added a PR on top of wg1-cleanup, so the lineup is

wg1 <- wg1-cleanup <- kwalcock/processCdrs.

I'll be sure not to erase the local copy until things get sorted out.  There should be no dependencies on eidos in that part.

BeckySharp:
moved code to `conceptdiscovery` so closing for now, we can def reopen as needed

PR #1012, opened 2021-05-11
Read files from actual CDRs and output json in an app

kwalcock:


kwalcock:
The lineup is

wg1 <- wg1-cleanup <- kwalcock/processCdrs.

BeckySharp:
@kwalcock I did things a little diff -- maybe we should coordinate before you do too much more?

kwalcock:
I wasn't going to do much more, but I'd certainly help integrate if it has messed things up.

BeckySharp:
> I wasn't going to do much more, but I'd certainly help integrate if it has messed things up.

I think it’s perfect and didn’t mess anything up, I hadn’t read things well enough by email. that said i reserve the right to ask for help if i need it :)

BeckySharp:
ok @kwalcock and @MihaiSurdeanu -- I moved the ConceptSink (and think I accounted for the no sentences thing) to processors.  However, do you think the other two (ConceptSource and the App) belong there?  They seem so use-case specific.  Which makes me wonder if I should actually move **all** of the concept discovery stuff to a separate repo.  Then they'll need to publishLocal two repos, but still.
FWIW:
I can convince myself that discovering and ranking concepts is general purpose.
I feel like the ConceptSink is generally useful bc it's exporting the ScoredConcepts.

thoughts?

BeckySharp:
for reference: https://github.com/clulab/processors/pull/486

kwalcock:
IM(not exceedingly strong)O all three and the rest of the discovery classes should be in a separate repo in the beginning and then only as they stabilize be moved "up" to processors.  Libraries are meant to change slowly and only leaf nodes of the dependency graph should be published locally.  Eidos was OK for that, but if it's too big, then better a new one than processors.  If the app shouldn't be an app, that's no problem.  It seemed like someone somewhere would need one, though.  It's a tiny amount of code, so please don't worry too much.

MihaiSurdeanu:
i kinda agree with Keith. I also don't have a very strong opinion. The openie subproject in processors is a natural fit for this, so I'd be Ok too if you moved them there directly.

BeckySharp:
closing bc code in conceptdiscovery repo, can reopen if desired

PR #1018, opened 2021-05-15
Update Dockerfiles for wmexchanger

kwalcock:


kwalcock:
These changes originate with Two Six and don't generally concern the average Eidos user.  The significant exception is the changes made to the build system, which has been upped to sbt 1.4.5.  sbt interaction with IntelliJ has been problematic recently and although this is working for Windows and on our test servers, that's a long ways from every student's computer.  Still, we have to start somewhere and perhaps better here than in processors or reach.  Then based on results this experiment we can update other projects over time.

kwalcock:
I'm sure it's possible, but I could not find a way to squash all the changes in advance so that the multiple changes to a file would be presented together, especially with some coming from a fork.  It will hopefully be apparent after I squash and merge them into master.

PR #1019, opened 2021-05-19
wip verbose grounding exporter

BeckySharp:


BeckySharp:
@zupon do you know the status of this?

zupon:
I'm not sure if we still need this? We didn't end up doing anything with it for the ontology augmentation exercise. Alternatively, it might still be good to have another exporter as an option for debugging in the future. If we *do* want it, I can make sure my other edits/updates are pushed before merging this in. What do you think?

BeckySharp:
yes, go ahead and push your edits and we should merge for future usage

zupon:
Actually, it looks like the most recent update was your `print topk` commit, so it should be good to merge! All I did after that was run it on some docs to test it out.

BeckySharp:
@kwalcock I guess my next steps will involve seeing why this can't be built

kwalcock:
You may have the wrong language.  I think it is trim() in Scala.

```
[info] compiling 152 Scala sources to /var/jenkins_home/workspace/eidos_grounding_exporter/target/scala-2.11/classes ...
[error] /var/jenkins_home/workspace/eidos_grounding_exporter/src/main/scala/org/clulab/wm/eidos/exporters/GroundingInsightExporter.scala:56:71: value strip is not a member of String
[error]         pw.println(s"Sentence $i:\t${doc.sentences(i).getSentenceText.strip()}\n\n")
[error]                                                                       ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

BeckySharp:
hahahaha NICE! yay me!

kwalcock:
Adding an exporter to the collection now...

PR #1021, opened 2021-05-26
Show compositional ontology version

kwalcock:


kwalcock:
The ontology files from https://github.com/WorldModelers/Ontologies get packaged into a jar and stored like other library dependencies of eidos.  They can usually be found somewhere like ~/.ivy2/cache/com.gitub.WorldModelers/Ontologies/jars (or more recently a coursier directory).  Inside the jar are org/clulab/wm/eidos/english/ontologies/CompositionalOntology_v2.1_metadata.yml and .../CompositionalOntology_v2.1_metadata.properties.  The latter should contain the hash code from the most recent commit that changed the ontology file.  It may not be the last commit in the repo, because other files could have been changed later.  The command `sbt runMain org.clulab.wm.eidos.apps.ShowOntologyVersion" should print it out.  This may all need to revised if the ontology starts to change daily.

FYI @yanzv, @reynoldsm88.

PR #1023, opened 2021-05-28
[DART-1043] -  Set Kafka enable auto commit to true by default and a fix to a compose file variables

yanzv:


yanzv:
@kwalcock I've pushed the change to set Kafka auto commit to true because I think wmexchanger is not committing offset manually.  If you are going to merge the PR please squash merge.  Not sure why the rest of the commits are part of this PR. 

Thanks!

kwalcock:
Thanks for the tip!  Too many tabs in too many tabs and I was mostly looking in GitKraken.  Next time will be a charm.

PR #1024, opened 2021-05-28
[DART-1043] - Set Kafka enable auto commit to true by default and a fix to a compose file variables

kwalcock:


kwalcock:
I'm not doing this right, but it is on purpose.  If git or GitHub can't show me what will change (or I can't figure it out), then it can expect users to take a detour.  Maybe it's a side-effect of the PR from a fork plus a squash commit.  This set of changes is the result of a manual directory diff between yanzv/eidos and clulab/master and replaces #1023.

kwalcock:
@yanzv, apparently I can't make you a reviewer.  There may eventually be a better procedure for seeing exactly what is going to change, but not this time.  Thanks for keeping the code up to date.  This should have the same effect as what you were doing.

yanzv:
@kwalcock I think this make sense.  I was thinking that if you look at files changed tab in the 1023 commit then you can see from 17 commits only 3 files were changed and you can see the changes and if you do squash commit it would just work.  I am going to pull from master to my fork and I see if next time I want to open another PR it will still bring these 17 commits.  It would be annoying if it will.

PR #1027, opened 2021-06-01
Add some paperwork regarding SRL regrounding

kwalcock:


kwalcock:
@BeckySharp, hope you don't mind.  I thought the code needed to explain itself a little more and nag us to address it, or I will lose track of it.

PR #1033, opened 2021-06-05
fix infinite loop in SRL grounder

### LINKED ISSUE #1032: This one seems to loop forever in the SRLCompositionalGrounder.  The sentence is
This one seems to loop forever in the SRLCompositionalGrounder.  The sentence is 

```
BOOST INCOME TO SUPPORT FOOD SECURITY AND NUTRITION .
```
from the attached file.  It does seem to loop in ExtractFromText.

[1e533493057c91173ac82bf821f7f82c.json.txt](https://github.com/clulab/eidos/files/6581565/1e533493057c91173ac82bf821f7f82c.json.txt)

_Originally posted by @kwalcock in https://github.com/clulab/eidos/issues/1022#issuecomment-852662669_

BeckySharp:
Happened when the predicate and the theme were both properties.
Now this is checked for.

closes #1032


kwalcock:
There was already an obscure (bad) system for dealing with the grounding tests.  It probably needs to be revisited, but maybe not just now.  I have a small update to this PR so that it works the old way.

kwalcock:
On second thought, I think the grounding tests may be revived sooner rather than later if it works to use the updated processors which has the fast loading vectors.  This fix is good if/when eidos is ready for it.

PR #1034, opened 2021-06-08
filter table sentences

### LINKED ISSUE #1022: File seems to loop endlessly in ArgumentExpander
The attached file, when run by ExtractCdrMetaFormDirectory, seems to loop endlessly in the ArgumentExpander until there is an out of memory error.  I'm trying to run it in a simpler program, but haven't gotten it to work.  The problematic sentence seems to be
```
15.0 20.0 25.0 30.0 35.0 40.0 45.0 Ja n -1 7 A p r1 7 Ju l1 7 O ct -1 7 Ja n -1 8 A p r1 8 Ju l1 8 O ct -1 8 Ja n -1 9 A p r1 9 Ju l1 9 O ct -1 9 Ja n -2 0 A p r2 0 Ju l2 0 Price trend of teffin Addis Ababa market Teff' white Teff' mixed Maize prices reduced or remained stable since June through August in most markets following harvests in unimodal areas but also increased imports from Uganda and Tanzania after the Government issued permits for private companies to import about 4 million bags of maize.
```
The filename needs to have .txt removed to be used with the program.

[630a3467d24e80c8e6097f2971794b82.json.txt](https://github.com/clulab/eidos/files/6551202/630a3467d24e80c8e6097f2971794b82.json.txt)

BeckySharp:
This filters sentences that contain >= 10 numbers or have >5 single-character words.
Using travis to check the tests...

closes #1022

### DIFF
--- a/eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala
+++ b/eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala
@@ -325,8 +325,8 @@ class EidosTokenizer(tokenizer: Tokenizer, cutoff: Int) extends Tokenizer(
     val shortSentences = sentences
       // The second change is to filter by sentence length.
       .filter { sentence => sentence.words.length < cutoff }
-      // This is to filter out tables mis-parsed as text.
-      .filter { sentence => notTable(sentence) }
+      // This is to filter out tables/graphs/etc mis-parsed as text.
+      .filter { sentence => notFubar(sentence) }
     val skipLength = sentences.length - shortSentences.length
 
     if (skipLength > 0)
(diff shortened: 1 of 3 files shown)


kwalcock:
It sounds like a good idea in general, but isn't it also skipping a sentence because the code that would run on it will crash?  Will there just be some other sentence that breaks it?  It will be great to be able to finish this document, though.  I think that others have processed sentences in different threads and given them time limits in case they run into problems (and/or added exception handling!).  This text might have come from the vertically printed labels on a graph rather than a table.  The PDF to text converter had difficulty figuring out whether the glyphs connected.  Maybe it should be isCryptic or isFubar or isNaS, not a sentence?

BeckySharp:
@kwalcock agreed on all points, but i care more about the latter.
If we run into issues with this in the future with a real sentence, I'd rather debug with that one.

I can rename the method

PR #1035, opened 2021-06-09
Be nitpicky about it

kwalcock:
Document my magic number
Do a count once instead of filter.length
Try lazy val within method

### DIFF
--- a/eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala
+++ b/eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala
@@ -322,16 +322,16 @@ class EidosTokenizer(tokenizer: Tokenizer, cutoff: Int) extends Tokenizer(
     // split() looks only at the word, not the token, so it is safe to have
     // rewritten the tokens at this point.
     val sentences = sentenceSplitter.split(tokens, sentenceSplit)
-    val shortSentences = sentences
+    val reasonableSentences = sentences
       // The second change is to filter by sentence length.
       .filter { sentence => sentence.words.length < cutoff }
       // This is to filter out tables/graphs/etc mis-parsed as text.
-      .filter { sentence => notFubar(sentence) }
-    val skipLength = sentences.length - shortSentences.length
+      .filterNot { sentence => isFubar(sentence) }
+    val skipLength = sentences.length - reasonableSentences.length
 
     if (skipLength > 0)
       EidosProcessor.logger.info(s"skipping $skipLength sentences")
-    shortSentences
+    reasonableSentences
   }
 
   // This is a bit misnamed, but we are overriding a processors method here.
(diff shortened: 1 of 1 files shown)


kwalcock:
I wonder if 5 or 6 is too low a value for the single char, especially given that most punctuation is a single char and has already been counted.  The period at the end of the sentence counts for one and pairs of quotes and parentheses count as two.

BeckySharp:
nice thought, this is one of the reasons I love working with you.  I didn't
even think about it.
Maybe instead we should check for only _.isLetter and .length == 1 (if you
get my drift)

kwalcock:
I would be (a little) more comfortable with that, I guess.  The previous sentence already has 6.

kwalcock:
Quite a few tests are failing.  The first one I checked had filtered out this sentence:
```
Crisis (IPC Phase 3) is widespread and
Emergency (IPC Phase 4) outcomes exist in parts of Unity,
Western Bahr el Ghazal,
```
This was even after the change to the single letter condition.  The numbers probably need to be higher or the kinds of symbols different.

BeckySharp:
OK, i'll look tomorrow, thanks for point this all out!

kwalcock:
I could output the counts for each measurement and then look at their distribution for at least the tests so that they can be made to pass.  It might be a good idea to check a real corpus eventually.

kwalcock:
Here it is for numSingleAlpha:

count | instances
-- | --
0 | 10384
1 | 666
2 | 92
3 | 8
4 | 1
5 | 0
6 | 0
7 | 0
8 | 0
9 | 0
10 | 0
11 | 0
12 | 0
13 | 0
14 | 0
15 | 2
16 | 0

kwalcock:
The 33 and 15 are for the one sentence we're trying to avoid.

kwalcock:
We know all the one-letter English words, don't we?  I and a.  Perhaps they should not be counted.

kwalcock:
Maybe not count valid numbers and punctuation.  For purposes of ruling out the single sentence, numSingleAlpha < 10 is sufficient.  We're getting too many false fubars otherwise IMO.

BeckySharp:
sounds fine to me

kwalcock:
This code isn't as elegant as that of @BeckySharp, but it does pass the tests.

kwalcock:
Do you think this is OK, @BeckySharp?  The numSingleAlpha from the table above looks like a more reliable indicator of a problem that causes the program to hang.  Because there are some valid 1-letter words in English that I wanted to take into account, the code changed location a little.

BeckySharp (approved):
Looks great, and yes I think that's a good litmus test.  If this were just an Object it wouldn't need to be passed around, but may not be as easily extensible.  then again, we may never need to extend it.  Either way, I totally approve and am simply adding my belated 1 cent. :)
thanks!

kwalcock:
Good, valid point.  I flipped a coin.  Luckily it's only about three lines of code.  I imagined a table detector, graph detector, PDF problem detector, ways to combine them, plugins, and language specializations... that would eventually be easier if it was a class.  For today, it's close enough for me.

PR #1038, opened 2021-06-17
Ground by default and test that way

kwalcock:


kwalcock:
This seems to add some 25 minutes to the tests.  That's too much.  It is probably quicker to restart eidos with grounding on for three tests than to ground everything.  I'll try that.

kwalcock:
Even if this one passes it will be replaced by a cleaner version.

PR #1039, opened 2021-06-21
Prepare for testing

kwalcock:
Change tests to should contain

Shorten grounding test code

Update TestGrounding to pass

Make a different category of test

Refactor test grouping code


kwalcock:
This is mostly a moving of tests that involve grounding to a separate package that can be used in test.sbt to consolidate them.  Having grounding turned on for all tests is rather expensive, so any in a package containing ".text.englishGrounding." are run as a group with grounding active.  Tests are still overall slow, but probably not because of this.  They will be revisited eventually.  Previously, grounding was off on all tests, but this changes the policy.

kwalcock:
I'm going to go ahead and merge.  I don't think any of @BeckySharp 's code was changed enough to be of concern.

PR #1047, opened 2021-07-13
Merge with master

kwalcock:


kwalcock:
This merge with master seemed to take care of the problems with TestGrounding and TestGrounderStability.  There was still infinite recursion in TestSRLGrounder, though.

BeckySharp:
@kwalcock though now after the merge I'm having compile isssues all the places where we are constructing a dependency graph, because now the constructor wants the roots.  e.g., when deserializing the sentenceObj from the jsonld.
Did you already make a fix for that and I messed it up? or should I work on it?

kwalcock:
Uh oh.  It's supposed to not want the roots now.  I'll check.

BeckySharp:
i may be using the wrong procVersion or constructor

kwalcock:
I just did a checkout srl_ner and built and ran tests with IntelliJ and it didn't have a problem.  Should I look harder?  Processors has been updated so that it calculates the roots automatically starting with 8.4.3.  Local changes that weren't yet at github could cause problems.

kwalcock:
FWIW the dependency on processors is in a subproject, eidoscommon, and IntelliJ will not automatically detect a change to the build.sbt file in that directory.  IntelliJ needs to be manually instructed to reload the main project to notice the change.

PR #1049, opened 2021-07-23
Update for July embed and publishing

kwalcock:


kwalcock:
I should have merged in master first.

PR #1054, opened 2021-08-03
Reorganize apps

kwalcock:


kwalcock:
* Apps are now configured in their own apps.conf separately from eidos.conf.  There is an EidosApp.scala to automate use of the conf file.
* The non-eidos conf files are moved out of the top level directory to avoid collisions.
* The many apps are organized into small groups, each with a README.md and very short descriptions.
* SimpleEidos was documented some more.
* Some util classes were moved around.
* No non-App depends on an App except for some tests, which will change eventually.
* The apps will probably be moved to a subproject in the next reorganization.  The root project will be reserved for the eidos library with a few exceptions like for the shell and an example.

BeckySharp:
Keith I love you, let me count the ways…

PR #1058, opened 2021-08-12
make easier to disable

BeckySharp:
Unless there's already a way...

### DIFF
--- a/src/main/resources/eidos.conf
+++ b/src/main/resources/eidos.conf
@@ -81,6 +81,7 @@ context {
 }
 
 sentenceClassifier {
+  enable = false
   classifierPath = ${EidosSystem.path}/sentenceClassifier
   tokenIDFWeights = ${sentenceClassifier.classifierPath}/sentence_classifier_idf_weight_1000_20000.txt
   evaluationFilePath = ${sentenceClassifier.classifierPath}/SentenceClassifierEvaluation.tsv
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -127,6 +127,7 @@ coref {
 }
 
 sentenceClassifier {
+  enable = true
   classifierPath = ${EidosSystem.path}/sentenceClassifier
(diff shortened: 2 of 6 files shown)


kwalcock:
There can be multiple ways to enable and disable things, but one runtime possibility had been accounted for.  Code would look approximately like this:

```scala
  val eidosSystem = {
    val config = ConfigFactory.load(EidosSystem.defaultConfig)
    val componentOpts = new ComponentOpts(eidosSentenceClassifier = false)
    val components = new ComponentsBuilder(config, EidosSystem.PREFIX, componentOpts = componentOpts).build()

    new EidosSystem(components)
  }
```

It was so that each component didn't need to have its own enabled/disabled setting in the conf file, a setting that probably wouldn't work for other tests or configurations.  Then we also don't need to edit the config in memory and don't need to reload the components each time.  They can also be filtered from existing components, like

```scala
    val componentOpts = new ComponentOpts(eidosSentenceClassifier = false)
    val newEidosSystem = new EidosSystem(oldEidosSystem.components.filter(componentOpts))
```

Maybe that would help, maybe not.

BeckySharp:
yeah, but the disadvantage of that is that it's in code, vs being able to set things up outside of code.  I think just about everything else is disableable in the config right?

BeckySharp:
also @kwalcock and @MihaiSurdeanu if you choose not to merge this, ping me so I change what I wrote in the wiki. (Though i vote for including it....) :)

kwalcock:
I saw this but can't look at it yet :-(

BeckySharp:
> I saw this but can't look at it yet :-(

no worries! you work too much as it is!

kwalcock:
@BeckySharp, you have convinced me, which is no easy feat :-)  However, I'm going to move the functionality to SentenceClassifier because it should know how to enable/disable itself and not need to pass that responsibility to the tests.  The PR will be ready shortly.

kwalcock (approved):

PR #1063, opened 2021-08-17
Test Python (without Scala)

kwalcock:


kwalcock:
This is an attempt to arrange for python unit testing for eidos.  This testing is only performed on the top level project right now.  Python files are expected to be in src/main/python and src/test/python.  The testPython task runs the tests.  All the files are first copied to target/python from which the tests are run with pytest.  testPython is not integrated into the normal test procedure, which continues to test only Scala/Java code.  Travis will test the python code and if you need to, just run `sbt testPython`.  See python.sbt for details.

MihaiSurdeanu (approved):
very nice!

kwalcock:
I meant that Jenkins will test the Python code since that is being used for eidos.  Travis may be doing the ontologies.

zupon:
Looks good! I'll try it out with the other scripts. Is this okay to merge?

PR #1066, opened 2021-08-23
Take care of RegroundExporter, useLeicons, Scala version

kwalcock:


kwalcock:
That is useLexicons.  More and more of my keyboard is going bad.

PR #1073, opened 2021-09-08
Override constructor for PredicateTuple

kwalcock:
...in a way that works for both 2.11 and 2.12.


kwalcock:
I'm merging this anyway because the tests weren't passing before.  They're at least all running now.

PR #1077, opened 2021-09-17
Output groundings in shell, webapp

kwalcock:
This isn't quite ready.  It is working in the EidosShell and in a little App GroundFromText.  The webapp is forthcoming.


MihaiSurdeanu:
Thanks @kwalcock ! This is awesome.

PR #1081, opened 2021-09-28
Make (filtered) examples available to grounders.

kwalcock:


kwalcock:
@zupon, this might contain code useful to you.  Let me know how it might be improved.  Some things will change with the ontology reformat.  I don't think it should go into master yet.  Can you merge this branch into yours?

zupon:
I think this should be useful! I'll merge it into my branch and try it out. This should give me access to all the examples in the ontology and the nodes they are associated with, correct? That should allow me to at least get a match to nodes with exact patterns as a starting point. After that, I can refine it to add the edit distance formula.

kwalcock:
This made it into the branch and will get to master via the branch.

PR #1084, opened 2021-09-30
Test, and demonstrate, EditDistance

kwalcock:


kwalcock:
@zupon, there are already at least three MED implementations available to eidos.  Any would probably work for experimentation.  If one is eventually used in production, then the license differences can be sorted out or an alternative found.  I didn't see an implementation directly in processors.

zupon:
Nice! I'll try these out.

PR #1090, opened 2021-10-18
Keep with old ontology format for now

kwalcock:


kwalcock:
@zupon and anyone else, you will need this really small fix to continue with the old ontology format.  There will be an update to Eidos to use the new format fairly soon.

PR #1091, opened 2021-10-20
updates NegationHandler to fix 'not only' bug

zupon:
Stops from making a negation attachment based on dependencies or left/right context if
- the outgoing dependency `neg` is coming from the token "only"
- the subsequent token to the word "not" is the token "only"

### DIFF
--- a/src/main/scala/org/clulab/wm/eidos/attachments/NegationHandler.scala
+++ b/src/main/scala/org/clulab/wm/eidos/attachments/NegationHandler.scala
@@ -150,7 +150,7 @@ class NegationHandler(val language: String) {
       if !argumentIntervals.exists(_.contains(tok))
       out <- outgoing.lift(tok)
       (ix, label) <- out
-      if label == "neg"
+      if label == "neg" && !(event.sentenceObj.words(tok) == "only")
     } negations.append(
       new TextBoundMention(
         Seq("Negation_trigger"),
@@ -178,7 +178,7 @@ class NegationHandler(val language: String) {
     for {
       (ix, lemma) <- leftContext ++ rightContext
       if !argumentIntervals.exists(_.contains(ix))
-      if (Seq("fail", "not") contains lemma) && !(previouslyFound contains ix)
+      if (Seq("fail", "not") contains lemma) && !(previouslyFound contains ix) && !(event.sentenceObj.words(ix+1) == "only")
     } yield new TextBoundMention(
       Seq("Negation_trigger"),
       Interval(ix),
--- a/src/main/scala/org/clulab/wm/eidos/attachments/NegationHandler.scala
+++ b/src/main/scala/org/clulab/wm/eidos/attachments/NegationHandler.scala
@@ -5,6 +5,7 @@ import org.clulab.wm.eidos.EidosSystem
 
 import scala.collection.mutable.ArrayBuffer
 import org.clulab.struct.Interval
+import org.clulab.wm.eidos.attachments.NegationHandler.failNot
 import org.clulab.wm.eidoscommon.EidosParameters
 import org.clulab.wm.eidoscommon.Language
 
@@ -150,7 +151,7 @@ class NegationHandler(val language: String) {
       if !argumentIntervals.exists(_.contains(tok))
       out <- outgoing.lift(tok)
       (ix, label) <- out
-      if label == "neg" && !(event.sentenceObj.words(tok) == "only")
+      if label == "neg" && event.sentenceObj.words(tok) != "only"
     } negations.append(
       new TextBoundMention(
         Seq("Negation_trigger"),
@@ -178,7 +179,7 @@ class NegationHandler(val language: String) {
     for {
       (ix, lemma) <- leftContext ++ rightContext
       if !argumentIntervals.exists(_.contains(ix))
-      if (Seq("fail", "not") contains lemma) && !(previouslyFound contains ix) && !(event.sentenceObj.words(ix+1) == "only")
+      if (failNot contains lemma) && !(previouslyFound contains ix) && event.sentenceObj.words(ix+1) != "only"
     } yield new TextBoundMention(
       Seq("Negation_trigger"),
       Interval(ix),
@@ -233,5 +234,7 @@ class NegationHandler(val language: String) {
 object NegationHandler{
 
   def apply(language: String): NegationHandler = new NegationHandler(language)
+
+  val failNot = Seq("fail", "not")
 
 }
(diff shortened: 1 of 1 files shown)


src/main/scala/org/clulab/wm/eidos/attachments/NegationHandler.scala
```diff
@@ -178,7 +178,7 @@ class NegationHandler(val language: String) {
     for {
       (ix, lemma) <- leftContext ++ rightContext
       if !argumentIntervals.exists(_.contains(ix))
-      if (Seq("fail", "not") contains lemma) && !(previouslyFound contains ix)
+      if (Seq("fail", "not") contains lemma) && !(previouslyFound contains ix) && !(event.sentenceObj.words(ix+1) == "only")
```
kwalcock, line 181: Can it be the case that `words(ix + 1)` is off the end?  It's language and weird things happen.  While this line is being touched, can that Seq be constructed once outside the loop in case it is being repeatedly generated?  It's old code, but better now than never.  It could be as far away as the `object NegationHandler` and be done just once ever.  For that matter, it may as well be an Array.
```scala
val failNot = Seq("fail", 'not")
...
if (failNot contains lemma)...
```
So both "not only" and "fail only" will be caught here.  Is that expected?

kwalcock:
This seems like something that could get broken, perhaps because something else stops even calling this code.  It would be very reassuring to have a unit test showing it in action and remaining in action.  Would it be possible to get one at some level?  I'm not sure it would have to affect the finding of Mentions or not.

kwalcock:
@zupon, thanks for tracking that down so deeply buried in the code!

src/main/scala/org/clulab/wm/eidos/attachments/NegationHandler.scala
zupon: I think this originally was intended to catch cases like "failed to X" and "did not X", but that did run into problems with "not only X", which we want to not put a negation attachment on. 

It is sort of hacky in the sense that I don't expect "fail not" to show up anywhere, but this does address the bug. I suppose we could get "fail only" in a context where we *do* want a negation attachment, but I can't think of an obvious case. Again, language *is* weird, so it might be good to separate those cases, which might mean separating `failNot` into two separate conditions. One that checks `if ("fail" == lemma)` and another one that checks `if ("not" == lemma && event.sentenceObj.words(ix+1) != "only")`.

Which would be better here, do you think?

kwalcock: I'm inclined to assert linguist's prerogative on your behalf and let you decide.  The tighter check comes at a computational cost and if the consequence of getting it wrong is minimal, I can see skipping it, although perhaps it would be good to record the decision in a comment.

I would still be worried about the ix + 1 because that might crash the program.

The `import` would generally be skipped and instead the line written
```scala
if (NegationHandler.failNot contains lemma)
```
so that imports name nouns (classes and objects) rather than verbs (methods).  Picky, I know.

zupon: Picky is good! I am but a humble scala amateur, so it's good to learn some best practices.

I think we can leave them combined for now, then if we get another complaint about a bug where "fail only" shows up we can adjust it then. I'll leave a comment about it though.

Agree about `ix + 1`. Without it being too convoluted, is there a good way to make sure that the `ix` is not the end of the sequence?

Something like the following to make sure the `ix + 1` is not beyond the last index of the sentence.
```
// if rightContext is empty OR if ix != last entry of right context (to avoid ix+1 error at end of sent)
if rightContext.isEmpty || (rightContext.nonEmpty && rightContext.last._1 < sentenceWords.lastIndexOf(sentenceWords.last))
if (NegationHandler.failNot contains lemma) && !(previouslyFound contains ix) && event.sentenceObj.words(ix+1) != "only"
```
If I understand my own code correctly, this will allow us to do `ix + 1` when the rightContext is empty (meaning we're only looking to the left of the trigger, so no worries about being off the end) or when the index at the end of the rightContext is less than the index of the end of the sentence (meaning we will still have at least one token to `ix + 1` into).

Can you see any glaring oversights with this?

src/main/scala/org/clulab/wm/eidos/attachments/NegationHandler.scala
```diff
@@ -175,11 +174,14 @@ class NegationHandler(val language: String) {
...
     val argumentIntervals = event.arguments.values.flatten.map(_.tokenInterval)
+    val sentenceWords = event.sentenceObj.words
     // Check for single-token negative verbs
     for {
       (ix, lemma) <- leftContext ++ rightContext
       if !argumentIntervals.exists(_.contains(ix))
-      if (failNot contains lemma) && !(previouslyFound contains ix) && event.sentenceObj.words(ix+1) != "only"
+      // make sure ix+1 is not the end of the sentence (to avoid ix+1 error at end of sent)
+      if rightContext.isEmpty || (rightContext.nonEmpty && rightContext.last._1 < sentenceWords.lastIndexOf(sentenceWords.last))
+      if (NegationHandler.failNot contains lemma) && !(previouslyFound contains ix) && event.sentenceObj.words(ix+1) != "only"
```
kwalcock, line 184: For someone like me who is not as familiar with the code, it's quite difficult to follow the relationship between rightContext and ix.  I'm hoping for an option that will minimize the probability that I (or others) will break this later, and that's a nearly impossible task :-(  Your patience is appreciated!

This version passes for me, but if it is wrong or it doesn't make sense to you, please "improve" it as needed:

```scala
    for {
      (ix, lemma) <- leftContext ++ rightContext
      // These are ordered roughly from fastest and most effective to slowest and least.
      if NegationHandler.failNot contains lemma     // lemma is either "fail" or "not", and
      if !(previouslyFound contains ix)             // ix is not a duplicate, and
      if !argumentIntervals.exists(_.contains(ix))  // ix is not part of any argument interval, and
      if !sentenceWords.indices.contains(ix + 1) || // either there is no next word at all or
          sentenceWords(ix + 1) != "only"           // there is one and the next word is not "only"
    } yield new TextBoundMention(
```
That is definitely a complicated set of circumstances.

MihaiSurdeanu:
Thanks @zupon !

I think this should be reorganized as we discussed today. That is, have a component that recognizes fake negation patterns once, and then check that the negation words are not covered by those patterns.
Thanks!

kwalcock:
@MihaiSurdeanu, at the meeting on Thursday we were wondering whether you would want to leave it like this or have it implemented as Odin rules (https://docs.google.com/document/d/1a1dmCW-haqQh4JJ47Quwvv6U0C49c8Pi6P2hjBGNgIc/edit).

MihaiSurdeanu:
Given that this is a simply surface pattern, it is fine as is!
Ok to merge. Thank you @zupon !

PR #1096, opened 2021-11-18
Streamline grounding code

kwalcock:


zupon:
This looks gorgeous! Thanks for all the streamlining!

src/main/scala/org/clulab/wm/eidos/groundings/grounders/EidosOntologyGrounder.scala
```diff
@@ -101,27 +102,52 @@ abstract class EidosOntologyGrounder(val name: String, val domainOntology: Domai
...
-    val overlapWithNodeName = embeddings.filter(embedding => StringUtils.afterLast(embedding.namer.name.toLowerCase.replaceAll("_", " "), '/', true).contains(lowerText))
-    // get the maximal overlap and return it
-    val overlaps = overlapWithText++overlapWithNodeName
-    val maxOverlap = if (overlaps.nonEmpty) overlaps.maxBy(_.namer.name.length) else null
-    val returned = if (maxOverlap != null) Seq(maxOverlap).map(exactMatch => SingleOntologyNodeGrounding(exactMatch.namer, 1.0f)) else Seq.empty
-    Some(returned)
+    // This tuple is designed so that Seq.min gets the intended result, the one with the min negLength
+    // (or max length) and in case of ties, the min position in the sentence, so the leftmost match.
+    // The embedding.namer should not be required to break ties.  It goes along for the ride.
+    // For expediency, the word count is used rather than the letter count.
```
zupon, line 112: Does this mean that the bigger match is the one that uses up more words rather than more letters? So "food aid transportation" would have more overlap to a "food_aid" node (2 words but only 7-8 chars) than to a "transportation" node (1 word but 14 characters)? If so, I think I agree with that approach!

kwalcock: Yes and I will update the comment.  I'm not sure the answer will be better, but the wrong answer is calculated faster.

src/main/scala/org/clulab/wm/eidos/groundings/grounders/SRLCompositionalGrounder.scala
```diff
@@ -187,87 +185,105 @@ class SRLCompositionalGrounder(name: String, domainOntology: DomainOntology, w2v
...
+            PredicateTuple(emptyOntologyGrounding, emptyOntologyGrounding, ontologyGrounding, emptyOntologyGrounding, tokenInterval.toSet)
         }
-        // If there is no content left, stop here.
-        // Just return the exact match
-        if (remainingContent.isEmpty) {
-          Seq(PredicateGrounding(exactMatchPredicateTuple))
+      }
+    }
+
+    def groundWithoutPredicates(sentenceHelper: SentenceHelper): Seq[PredicateGrounding] = {
```
zupon, line 204: I really like separating the without/with predicates conditions like this!

kwalcock: Thanks for being flexible.  It was mostly to cut down the length.  The method/paragraph was too long for me.

kwalcock:
If these tests still pass (and after I fix a typo), then I'm probably done with it.  No functionality is supposed to have changed.  The grounding with predicates was divided again into functions for exact and inexact matching, just for length.  That might mess up some documentation :-(  Some names have been changed.  I hope it didn't get out of hand.

kwalcock:
That will teach me not to fix typos.  I'm looking into this.

```
[info] - should process "The brewing conflict had already had a serious impact in disrupting farming which had led to higher prices ." effect themeProperty correctly *** FAILED ***
[info]   "[]" was not equal to "[wm/property/price_or_cost]" (TestGrounding.scala:40)
[info]   Analysis:
[info]   "[]" -> "[wm/property/price_or_cost]"
```

kwalcock:
There seems to be a problem.  I had changed the order of

```scala
      val predicateGroundings = inexactPredicateGroundings :+ exactPredicateGrounding
```
to
```scala
      val predicateGroundings = exactPredicateGrounding +: inexactPredicateGroundings
```
before the line
```scala
      val sortedPredicateGroundings = predicateGroundings.sortBy(-_.score)
```

figuring that the exact match will more often have a higher score, so it is better to put it at the front of the list and not have to sort it all the way there.

In this case they are all tied at zero, though.  The exact one is empty, with a predicateTuple of (None, None None, None).  That's an obvious score of zero.  The other two candidates are (Some, None, None, None), but they also have scores of zero.  I'm looking into it.

kwalcock:
I should have written (None, Some, None, None).  It is the themeProperties that are filled in.  If that is a valid grounding, at it is something that can be produced as a result, then I think that the themeProperties and themeProcessProperties need to be taken into account in PredicateTuple.score.  Right now only the theme and themeProcess are used.  Should the properties by discounted by half (or something), or are they just as significant?

kwalcock:
The change to PredicateTuple.score was made the tests pass locally, so I will push it.  However, it should be looked at critically.  Perhaps some of the tests can be rerun to see if it has made any difference elsewhere.

zupon:
Thanks for looking into this in depth! This sounds like a legacy of when we _had_ to have a theme (so it was required in calculating the PredicateTuple.score), but now that a theme is optional, I think it's okay to change PredicateTuple.score to account for the various properties.

zupon:
Now that everything is passing again, is this okay to merge into the `comp_grounding_unit_tests` branch?

kwalcock:
Sounds good, especially if I don't merge my own code.

PR #1100, opened 2021-12-08
Reformat ontology

kwalcock:


kwalcock:
Here's a version that I think can be merged.  However, it does bring up a grounding issue that should be considered, @zupon and others.

The canonicalizing of ontology nodes via the Namer interface used to work by (possibly) expensively putting together the path of the ontology node with parts separated by / into a string, then looking after the last / for words.  If a node isn't a leaf node, the convention is that it end with a /.  When looking after the last /, one gets nothing and this essentially takes non-leaf nodes out of the running for exact matches.  This is an around the bush way of making that happen and I'm not sure it is intentional or good.  Unfortunately, if I go ahead and use the non-leaf nodes, which are possibly terms that are overly generic, one of the tests fails.  It's the `behavior of "aug13_501"` in TestGrounding.  Instead of economy/economy it gets crisis_or_disaster/conflict/.  Is that better or worse?  Have we already tried both ways?  How would checking for exact matches on non-leaf nodes affect other non-passing tests that aren't being run?  To do it, one would just need to remove the two lines in Namer like `if (!isLeaf) ""`.  I don't see anything about this in the wiki documentation.  If we do want to continue to leave out the non-leaf nodes, then I would move that decision to the grounding algorithm with an explicit line of code and not have the Namer make the exception.

PR #1101, opened 2021-12-10
Use Scala Map, Set, Bag for Identity and Equality

kwalcock:


kwalcock:
It's probably not a good policy to merge ones on PRs, but I'll do it shortly.  I've decided to run a regression test first.

kwalcock:
The extra regression and round-trip tests passed.

PR #1110, opened 2022-01-04
[DART-1248] - Adding a way for wmexchanger to connect to DART without specifying authentication credentials

yanzv:
For DART in the box eidos is going to be running within the same docker network as the rest of the applications and I would like to be able to configure for wmexchanger to connect to Kafka and REST endpoints directly without authentication. @kwalcock please let me know if you have any concerns with these config changes.  Also, squash merge to if you are going to accept the pr.

Thanks!


kwalcock:
Cool.  I'll look at it soon.

kwalcock:
It looks OK to me.  Thanks for dealing with the code duplication.  To run on separate networks again, one would need to revert the changes to application.conf, I guess.  Do you all there prefer an empty String to an Option[String] of None?

yanzv:
Great. To run on a separate network you would just configure authentication via ENV variables.  
 ```
username: ${?REST_PRODUCER_USERNAME}
 password: ${?REST_PRODUCER_PASSWORD}
```

I think Option[String] would be preferable but I didn't want to change the method signature.

**Comment by kwalcock** [comment]
Were these values unnecessary or wrong then?  (I'm making notes in case it does need to go back one day.)

```
    sasl.jaas.config: "org.apache.kafka.common.security.plain.PlainLoginModule required username=\""${KAFKA_USERNAME}"\" password=\""${KAFKA_PASSWORD}"\";"
    security.protocol: SASL_SSL
```

yanzv:
This is when you are connecting to SASL_SSL authenticated kafka endpoint otherwise the connection will fail if the protocol is PLAINTEXT

kwalcock:
You might be interested in webapp/docker.sbt.  Recently we've been having sbt build the docker images.  It works in two stages (as your system does) and probably gets to skip some of the time-consuming intermediate steps.  It has seems fast to me.

PR #1116, opened 2022-03-01
Compose compositional groundings

kwalcock:


kwalcock:
I should go over this again and clean it up, particularly if the code continues to be used.  It's the version doing the reading right now and it is otherwise only tagged.  Quite a bit of dead code has accumulated and the unit tests should also be revised to stay relevant, but not on this PR.

MihaiSurdeanu:
Agreed!
Thanks Keith!

kwalcock:
This is going in followed right away by the wmexchanger branch which has all the additional compositional changes.

PR #1125, opened 2022-04-09
Clean groundings tests

kwalcock:
To test them all, change 

          case FAIL =>
            failingTest should title taggedAs Somebody in {

to

          case FAIL =>
            passingTest should title taggedAs Somebody in {

### DIFF
--- a/src/test/scala/org/clulab/wm/eidos/text/englishGrounding/TestGrounding.scala
+++ b/src/test/scala/org/clulab/wm/eidos/text/englishGrounding/TestGrounding.scala
@@ -298,501 +298,325 @@ class TestGrounding extends EnglishGroundingTest {
     }
   }
 
-  // val tests = Array()
-  // do foreach.test()
-
-  {
-    val test = Test(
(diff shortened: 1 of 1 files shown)


kwalcock:
@EgoLaparra, this may affect the work you are doing, so it would be great if you could take a look and let me know if I should change something or delay merging.  It shouldn't be necessary anymore for you to replace failingTest with passingTest in a hundred places.  The one change above should be sufficient.

I notice there are about 53 texts being tested.  If there are causes and effects and four slots, there are 8 tests per text for just things being in the right slot.  That accounts for 424 of the 789 tests that are passing.  At some point the code could produce values in a bad slot, but the code was later changed to not be able to do that.  Those tests are guaranteed to pass and I'm not sure that makes for a fair comparison.

kwalcock:
@EgoLaparra, can you take a look?  I'll send email in case the notifications aren't working.

EgoLaparra:
@kwalcock this doesn't seem to have any effect on the evaluation I've been working on. The tests run noticeable faster though.

